Issue #98, opened 2015-01-01
Twig process not working on modular pages


namaless:

rhukster:
Can you provide some information and perhaps an example of how to recreate?

namaless:
Simple, you put this in your module page:
---
title: blabla
---
Content **example** {{ page.title }}

inthe config enable twig process ofc.

namaless:
I can confirm the process not working because if i use a plugin function (in the page working fine) in the page module not working.

rhukster:
Twig is not enabled by default in any markdown, so did you enable Twig with:

```
---
process:
  twig: true
---
```

???

rhukster:
This was fixed in Grav 0.9.12

PR #99, opened 2015-01-02
Update Twig.php

namaless:
Related this issue: https://github.com/getgrav/grav/issues/98

### DIFF
--- a/system/src/Grav/Common/Twig.php
+++ b/system/src/Grav/Common/Twig.php
@@ -183,6 +183,10 @@ class Twig
                 $twig_vars['content'] = $content;
                 $template = $item->template() . TEMPLATE_EXT;
                 $output = $local_twig->render($template, $twig_vars);
+                
+                // Render page content
+                $this->setTemplate($template, $output);
+                $output = $local_twig->render($template, $twig_vars);
             } else {
                 $name = '@Page:' . $item->path();
                 $this->setTemplate($name, $content);


system/src/Grav/Common/Twig.php
```diff
@@ -183,6 +183,10 @@ public function processPage(Page $item, $content = null)
                 $twig_vars['content'] = $content;
                 $template = $item->template() . TEMPLATE_EXT;
                 $output = $local_twig->render($template, $twig_vars);
```
mahagr: Did you forget to remove this line?

namaless: I see only one content, you sure to use the latest grav version and cache cleared?

mahagr: I was just commenting the code; take a look the lines 185 and 189.Basically you're rendering output, setting template and rendering output again (discarding the old one).

namaless: This method need when you use a modular template with more one module. If you compile only one time, you compile only the first module, then other modules is empty.

mahagr: Oh, now I see it: you're using the same $output variable for two different meanings. Can you please change one of the variables to $modular_output (or just to be different) and add a comment why we need to do this? Otherwise someone comes later and tries to change the code without understanding it -- just breaking the feature again.

rhukster:
Implemented similar but slightly different :)

PR #105, opened 2015-01-05
Update composer.json to not use GitHub API for rockettheme

jeffdm:
I kept running into the following error when installing composer packages (as a part of  `bin/grav install`). It came down to my computer hitting the GitHub API rate limit for rockettheme. This changes rockettheme to not user the API and just download the repo directly.

`
Loading composer repositories with package information
Failed to clone the git@github.com:rockettheme/toolbox.git repository, try running in interactive mode so that you can enter your GitHub credentials


                                                                                                                                                     
  [RuntimeException]                                                                                                                                 
  Failed to execute git clone --mirror 'git@github.com:rockettheme/toolbox.git' '/root/.composer/cache/vcs/git-github.com-rockettheme-toolbox.git/'  
                                                                                                                                                     
                                                                                                                                                     
                                                                                                                                                     


install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-plugins] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [--ignore-platform-reqs] [packages1] ... [packagesN]


Failed to clone the git@github.com:rockettheme/toolbox.git repository, try running in interactive mode so that you can enter your GitHub credentials
`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
     "repositories": [
         {
             "type": "vcs",
+            "no-api": true,
             "url": "https://github.com/rockettheme/toolbox"
         }
     ],


rhukster:
We will look at this asap. Thanks

rhukster:
I'm having trouble finding any documentation on this `no-api` method. Where did you find out about it?

jeffdm:
I read about it here: https://getcomposer.org/doc/05-repositories.md :smiley: 

```If you set the no-api key to true on a github repository it will clone the repository as it would with any other git repository instead of using the GitHub API. But unlike using the git driver directly, composer will still attempt to use github's zip files.```

jeffdm:
If it helps, I'm trying to create a Dockerfile for development and deployment of Grav.

One of the Dockerfile steps is to run `./bin/grav install`, this runs in non-interactive mode so there's no way for the user to insert their GitHub credentials. They probably shouldn't have to anyways :smiley:

PR #116, opened 2015-01-18
Change constants to streams.

namaless:
Added @todo for phpstorm can help me to see where already tried to use
locator.


system/src/Grav/Common/File/CompiledFile.php
```diff
@@ -14,6 +15,8 @@
  */
 trait CompiledFile
```
mahagr: CompiledFile is not meant to be using locator; instead I would make static function to initialize the cache path.

namaless:

PR #117, opened 2015-01-18
Change find method to dispatch

namaless:
- Change find method to dispatch because find is deprecated.


rhukster:
$page->find() is not deprecated, it's used quite a bit

namaless:
In the method header @deprecated it's called, i see from this, so sorry.

PR #119, opened 2015-01-18
Fix page media url.

namaless:
- Remove unused variables.
- Change from base_url to pages base. this fix the issue with
multisite/multilang/themes_preview plugin.


rhukster:
I've reworked this a little.  Please test to see if this works ok for your multisite stuff.

namaless:
Confirmed, working with multisite.

PR #120, opened 2015-01-18
Replace constants with streams.

namaless:
- Replace constants with streams.


system/src/Grav/Common/Assets.php
```diff
@@ -105,6 +106,8 @@ public function __construct(array $options = array())
      */
     public function config(array $config)
```
mahagr: I'm not sure if assests should be using locator either as they are compiled files. We should probably have static function to initialize its public directory.

namaless:

PR #121, opened 2015-01-18
Bugfix/replace contants with streams

namaless:


namaless:

PR #122, opened 2015-01-18
Move down deprecated constants.

namaless:
- Move down deprecated constants.
- Fix trailing slash for GRAV_ROOT because ROOT_DIR now deprecated.

### DIFF
--- a/system/defines.php
+++ b/system/defines.php
@@ -7,25 +7,25 @@ define('DS', '/');
 
 // Directories and Paths
 if (!defined('GRAV_ROOT')) {
-    define('GRAV_ROOT', getcwd());
+    define('GRAV_ROOT', getcwd() . '/');
 }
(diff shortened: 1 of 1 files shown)


system/defines.php
```diff
@@ -7,25 +7,25 @@
 
 // Directories and Paths
 if (!defined('GRAV_ROOT')) {
-    define('GRAV_ROOT', getcwd());
+    define('GRAV_ROOT', getcwd() . '/');
```
mahagr: GRAV_ROOT was not meant to have trailing slash; trailing slash has been very problematic for me and caused a lot of extra work to remove the flash in many locations. For example locator needs normalized path without trailing slash to work properly.

There's a good rule to remember: always try make minimum possible changes. I would keep the variables as they were defined and just make them deprecated. Otherwise you will easily end up breaking the new or old code instead of just deprecating it.

namaless:

PR #123, opened 2015-01-19
Fixed twig url function to return empty if not found

namaless:
- If not found the stream with locator return empty instead the root url.


rhukster:
I looked into this, but I can find no way for the locator to work with URLs.  Internally the locator is looking for file paths, and a URL is not going to be a valid filepath, nor is locator going to return a valid URL.  Passing back to @mahagr to investigate further.

namaless:
Fixed with your modification in the latests commits. Nice thanks!

PR #124, opened 2015-01-20
Replace constant with streams. (need for multisite).

namaless:


rhukster:
Going to need this one again, against Develop...

mahagr:
Yes, it needs to be rebased against develop.

PR #131, opened 2015-01-27
Bug fixes phpdoc/missed vars/locator.

namaless:
- Removed unused variables.
- Add missed phpdoc params.
- Update constants with locator.


system/src/Grav/Common/Page/Medium.php
```diff
@@ -407,11 +417,15 @@ public function __call($method, $args)
      */
     public function image($variable = 'thumb')
     {
+        $locator = self::$grav['locator'];
+
+        $images_dir = $locator->findResource('image://');
```
mahagr: For relative path I suggest of doing 

    $foo = $locator->findResource('image://', false);

All resources get cached, so it's fast to receive the same path.

PR #132, opened 2015-01-29
Added 'enabled' option for summary and fixed query support in assets

Sommerregen:
Dear Andy,

according to the recent forum post about adjusting summary lengths in Grav, I realized that there is no easy switch to easily enable or disable truncationation of summary contents. I think it is better to let the user decide, whether he or she wants a short version of his/her articles shown on frontpage or not.

Further I noticed, that using
```
$this->grav['assets']->add('https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML');
```
does not add the javascript as expected. That is a bug due to pathinfo which doesn't strip out the query of the asset.


rhukster:
Thanks for your contribution!

PR #133, opened 2015-01-29
Added merge config and disable function in "Plugin.php"

Sommerregen:
While developing plugins I missed some really important functions again and again:

- There should be definetely a `disable` function doing right the opposite as 'enable'.
- I really like the idea of having global plugin configurations and page specific configurations. But since merging them together when needed is always a mess, I propose to add a `mergeConfig` function to the core functions. It will automatically load the plugins configurations merge them with the page header, returning back a new config data object. Usage in the plugin would be e.g.
```php
$config = $this->mergeConfig($page);
$option = $config->get('plugin_name.option');
```

### DIFF
--- a/system/src/Grav/Common/Plugin.php
+++ b/system/src/Grav/Common/Plugin.php
@@ -1,6 +1,9 @@
 <?php
 namespace Grav\Common;
 
+use Grav\Common\Inflector;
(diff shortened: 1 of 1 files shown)


mahagr:
I know this is already merged, but there are some "issues" I found...

PR #144, opened 2015-02-15
Fix to merge nested page options with default plugin configurations

Sommerregen:
Dear Andy,

today I encoutered a minor bug in the **Plugin** class, that nested page overwrite default plugin configurations although some are not set. For example consider we have the follwing configuration in our YAML file (see e.g. my DropCaps plugin):

```
titling:
  enabled: false              # Toggle to (de-)activate titling of text
  first_line: true            # Highlight first line of paragraph
  breakpoints: ".,:;!?'\"-"   # Specify which characters will end titling
```

We then want to enable `titling` in a page via

```
titling:
  enabled: true
```

When using `$config = $this->mergeConfig($page)` we finally would expect to get

```
titling:
  enabled: true     # <-- Changed via page header
  first_line: true
  breakpoints: ".,:;!?'\"-" 
```

but at the moment we get

```
titling:
  enabled: true
```

My PR will fix this to get the desired result from above. It would be nice to have this behavior in core.

### DIFF
--- a/system/src/Grav/Common/Plugin.php
+++ b/system/src/Grav/Common/Plugin.php
@@ -141,7 +141,7 @@ class Plugin implements EventSubscriberInterface
 
         // Get default plugin configurations and retrieve page header configuration
         if (isset($page->header()->$class_name)) {
-            $header =  array_merge($defaults, $page->header()->$class_name);
+            $header =  array_replace_recursive($defaults, $page->header()->$class_name);
         } else {
             $header = $defaults;
         }


rhukster:
The only problem i can for-see with this is that if your configuration is an array such as:

myplugin.yaml
```
enabled: true
title: My Plugin
sizes_available:
  - xSmall
  - Small
  - Medium
  - Large
  - xLarge
```

Then in your page you override:

```
myplugin:
  sizes_available:
    - Small
    - Large
```

You will get the deep merged result:

```
sizes_available:
  - xSmall
  - Small
  - Medium
  - Large
  - xLarge
```

When you intended to only get what you set:

```
  sizes_available:
    - Small
    - Large
```

This is the fundamental problem with merging.  How do we determine if the intended behavior is to replace or merge?

mahagr:
This is where blueprints are supposed to help. But even then we need some hints inside the configuration on what's the supposed behavior.

namaless:
It's simple with array_replace_recursive, if you want to reset before setup your array you simple:

sizes: []
sizes:
  - xSmall
  - Large

Sommerregen:
Good points, @all!

Next time I open up such things as an issue, not as pull request... (can someone move it?) However, I don't know what's the best way. The easiest way is to change the structure in the YAML file, so that a deep merge is no longer necessary. Then the best would be to have something like a guidline of how to write plugins.

But I really like the approach to define the deep copy in the blueprint. Maybe something like

```
config:
  fields:
    titling:
      merge: 'deep'
```

could do the job. Another approach I can think of is to add an additional parameter to `mergeConfig` function like

```
protected function mergeConfig(Page $page, $deep = false) {
  ...
}
```

where `$deep` can be boolean in the simplest case or an array (e.g. `array('titling')`). Would love to hear your thoughts and other ideas.

namaless:
I used internal version for my plugins because sometimes include the config+header+twig function in the same config array, so for me it's useless this, for other i post the best way without any modification to core :)
Btw your idea is nice no worries ;)

rhukster:
I thing the $deep option might not be a bad choice for now.  It doesn't solve all the problems, but at least it gives some more control to the developers.

rhukster:
Deep option has been added in commit 6811fbe

PR #145, opened 2015-02-20
Updated dependencies to stable versions

hason:


mahagr:
I think we had to use dev-master on some of the projects because of the bug fixes in there. But I do agree that we should really stay on stable releases as using dev-master will likely cause site to fail to load after composer update; especially if there are backwards incompatible fixes in there.

rhukster:
I have every intention  of moving to more stable releases as we get closer to **1.0**, but there are quite a few projects here that are actively being developed and bug fixes are needed.  I'll keep this one around, to remind me to get to stable versions in the coming months.

rhukster:
There's too many important updates in the dev-master of debugbar, but the rest i've set to specific versions.  Thanks.

PR #150, opened 2015-03-08
Changed Twig Link Regex which fixes #149 [Markdown Doesn't Handle [words](#)

### LINKED ISSUE #149: Markdown Doesn't Handle [words](#)
You can see the problem on the Grav one page demo around this Markdown:

```
We believe the best recipe to make something great, is to bake with [quality ingredients](#) and add dollops of [cool stuff](#).
```
Everything after `[quality ingredients](#)` is cut off and the "quality ingredients" link doesn't work.

If (#) is replaced with (full url), then it renders and behaves correctly.

But I can't tell if it's Parsedown's problem or Grav's.

Sommerregen:
Hi @rhukster,

here is the PR of bugfix for issue https://github.com/getgrav/grav/issues/149 .

### DIFF
--- a/system/src/Grav/Common/Markdown/ParsedownGravTrait.php
+++ b/system/src/Grav/Common/Markdown/ParsedownGravTrait.php
@@ -19,7 +19,7 @@ trait ParsedownGravTrait
     protected $pages_dir;
     protected $special_chars;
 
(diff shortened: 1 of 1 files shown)


Sommerregen:
Edge case "fixed" :grinning:

bobrocke:
That looks like a fix - works for me.

rhukster:
Looks good to me! Thanks!

PR #188, opened 2015-05-05
Added resumable download support

Sommerregen:
Hi @rhukster,

this PR adds a resumable download support into Grav. It supports a download limit as well as a qualitity parameter called `download.balance` to tune the download rate (num flushes versus memory size used for buffering). Further it polishes the `getMimeType` method (having speed in mind) and adds support for more mime types.

### DIFF
--- a/system/config/system.yaml
+++ b/system/config/system.yaml
@@ -42,7 +42,6 @@ cache:
   lifetime: 604800                     # Lifetime of cached data in seconds (0 = infinite)
   gzip: false                          # GZip compress the page output
 
-
 twig:
   cache: true                          # Set to true to enable twig caching
   debug: false                         # Enable Twig debug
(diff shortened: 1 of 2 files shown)


rhukster:
This is interesting benny.  One thing I would like to see before merging, is that this be off by default and something you can enabled. Resumable downloads are great for large files, but the download() method is also used to download/send other files including images.  it needs to work optimally and fast for small files too.  There's a lot of work going on here to support the resumable that is only going to add overhead in the majority of cases.

Sommerregen:
Alright I will add another option in the config and/or param to the method and change the behaviour to off by default. Concerning small and large files, I considered both. Small files are almost sent to the browser in one flush (this depends on the `chunk_size` when no download rate is given). You can try it. 

But I totally understand your point. Most of the code is needed to handle all kinds of byte ranges. The rest can be done in a few lines. Thus, I will refactor the code and break it in at least two methods to reduce the overhead (one fast one for downloading only and the other one for supporting resumable downloads).

rhukster:
Sounds great! Thanks.

mahagr:
Wow, nice work on this one! You even added parameter to change the filename what user will see. Will be great especially if you want to support protected downloads (with hashed filenames).

Sommerregen:
Dear @rhukster, as already mentioned in the chat, the best option concerning performance of the core and extensibility would be to release my changes as an own plugin. I think the basics are already present in core. For doing this I will use the `onBeforeDownload` event and add some stuff there.

Thanks, for the good discussion here! :-D

PR #189, opened 2015-05-05
Added params option for mergeConfig method

Sommerregen:
Ok and here is my second PR changing a little bit the behaviour in the `mergeConfig` method in the`Plugin.php` file. Now it supports addtional params i.e. the function can be used for Twig filters. Further it interprets in the header

```
pluginname: true
```

now as

```
pluginname:
  enabled: true
```

for the lazy man, who only wants to enable/disable a plugin in one-line. It would be very nice, if this feature will be a part in Grav, since I have a plugin pending (the `Toc` plugin), which already uses this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,76 @@ php: 5.6
 branches:
   only:
     - master
+    - build_test
 notifications:
   email:
     on_success: never
     on_failure: always
+  hipchat:
+  # hipchat_api@grav
+    rooms:
+      - secure: "bqO0wM1B7bJnQw2fuhquSXEqI9gw6WmFytIh9sEWXzbYTzTUP5t0PcKOd3FT2BNMRaDxPJLVl+vG/oqmqDUBkEmOGcG504IQjeNzZqnMz0tXQMIcCc22Las9tFfc4Jf6RVi/qGomFtHGE9Wgii+TAN4zqZaufbNjwd8SyjO0+W8="
+    template:
+      - '%{repository}#%{build_number} (%{branch}): Grav Core and Skeletons packages have been uploaded. (<a href="%{build_url}">Details</a>)'
+    format: html
+env:
+  global:
+    # Colors!
+    - TEXTRESET=$(tput sgr0) # reset the foreground colour
+    - RED=$(tput setaf 1)
+    - GREEN=$(tput setaf 2)
+    - YELLOW=$(tput setaf 3)
+    - BLUE=$(tput setaf 4)
+    - BOLD=$(tput bold)
+    # User
+    - GH_USER="getgrav"
+    # Paths
+    - RT_DEVTOOLS=$HOME/devtools
+    - GOPATH="$HOME/go"
+    - PATH="$GOPATH/bin:$PATH"
+    # GH_TOKEN [API Key]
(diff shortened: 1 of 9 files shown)


Gertt:
Tested this, works fine. Few questions/consideration though, these are just my own thoughts on the matter.

- I *think* **most plugins** only check the enabled flag once at the start, by the time they call mergeConfig, they only check other config values. Will it not add extra complexity instead of making it easier when users start writing `pluginname: false` and nothing happens.

- Is the added value of an extra parameter in mergeConfig worth the extra complexity? Plugin writers can very easily do this themselves with one extra line of code in their plugin.

Gertt:
Refactored: https://github.com/getgrav/grav/commit/45c11188e6653b71c19534f721d3ca916bb93523

rhukster:
Comment on @Gertt's first point, enabled option is checked by plugins class and the plugin is not even loaded if this is set to `false`.

Gertt:
@rhukster Not sure if you are agreeing or disagreeing :) But imho your comment further validates my concerns as mergeConfig (which you call from inside your plugin code) happens only after Grav has checked the enabled option and thus the plugin has already been loaded by the time you would override the enabled option from your page.

Sommerregen:
@Gertt Overwriting the enabled option is intended to switch on/off the behaviour of the plugin just for a specific page. I.e. consider a plugin has `enabled: true` option, but a user want's to disable this plugin for one or few pages then the user can do it by setting `pluginname: false` or `pluginname: enabled: false`.

Ok, there is something which I indeed haven't considered yet: the case, where the plugin is globally disabled and one wants to enable it for one page. Then, this is not working though.

Concerning your second point, as far as the plugin is simple, it doesn't need it. However, @namaless has used a similiar feature in his plugins and I really need it, too. Especially for Twig filters. I can write a special method, which I will henceforth use in my plugins. But I thought instead of copy and pasting the same code (which contradicts the DRY principle) it is better to push it to the core... However I'm already using another method throughout my plugins (`compileOnce`) to ensure that the code just runs once for every page (and not two or more times). Whereas this method is something special, the params merging option is more fundamental and thus in my eyes belongs to the core.

Maybe I'm wrong, but it finally depends on yours and @rhukster decision.

rhukster:
There is actually two variables that can be used:

1) config `enabled` is intended to be used at a low level to enable/disable the plugin from even loading
2) `$this->active` is a local plugin-level setting that is intended to be used to dynamically enable/disable the plugin from within the plugin.

The case you bring up is valid and the thought is that the plugin is enabled, but can be 'inactive by default', then switched to 'active' if enabled/activated by the page.

This would have the enabled config setting set to true, but some 'active_by_default` set to false, until activated by the page event.  If this makes any sense!  :)

Sommerregen:
@rhukster Yeah, it makes any sense. Maybe I'm reflecting once again on the implementation and provide some additonial commits to this PR. The possibility to use `$this->active` is really attractive and could be used, when `pluginname: false` is set in the page header.

Sommerregen:
@Gertt You did a great job in simplifying my code! I have nothing to add. Thus, Gertts code https://github.com/getgrav/grav/commit/45c11188e6653b71c19534f721d3ca916bb93523 should be merged and this PR can be closed.

Sommerregen:
Any update on this PR? For the convenience I incorporated Gertt's changes into this PR and did a rebase.

rhukster:
Damn I had forgotten all about this one.  Will take a look at it this weekend, promise!

PR #191, opened 2015-05-14
Use global composer install when available

eschmar:
To avoid the manual placement of composer into ``bin/``, this pull request proposes a centralised check for the availability of a global composer installation (and fallback to ``bin/composer.phar``).


rhukster:
Nice, I like it! merging!

eschmar:
Thanks for the quick merge, @rhukster :blush: Please have a look at #192, though. Forgot ``trim`` in the copy process.

PR #192, opened 2015-05-14
Fix composer location method

eschmar:
My previous pull request #191 unfortunately had a small bug - the ``shell_exec`` call returned a string with an ending line break, thus breaking the command. Sorry for the trouble.

### DIFF
--- a/system/src/Grav/Common/Composer.php
+++ b/system/src/Grav/Common/Composer.php
@@ -17,8 +17,12 @@ class Composer
      */
     public static function getComposerLocation()
     {
+        if (!function_exists('shell_exec')) {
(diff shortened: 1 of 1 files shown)


w00fz:
Could you also add a check with `function_exists` to make sure `shell_exec` is available and to prevent possible errors?

w00fz:
Also perhaps `command -v` instead of `which`, more universal and faster!

rhukster:
If you can please implement w00fz ideas (good ones), i'll get this merged too.   Thanks!

eschmar:
Sure, good ideas! Just pushed the changes.

PR #206, opened 2015-06-20
Add `onBuildPagesInitialized` event for memory & time consuming plugins

Sommerregen:
Hi @rhukster,

regarding our yesterday's chat, I prefer to add a new event `onBuildPagesInitialized` into core, to allow memory and time consuming plugins to be initialized only when it is really needed (e.g. a page has changed and therefore `onPageContentRaw` or `onPageContentProcessed` is triggered). Thus, when there are no changes the plugin isn't even initalized! This makes Grav (in my opinion) very powerful, since otherwise all plugins (even the time consuming ones) have to be initialized in the `onPluginsInitialized` event.

Ok, caching could be speed up the process. But not everything could be chached, i.e. when one can register additional classes and imho it would be the best, if even the caching (load and process data) should only happen, when there is actually some uncached pages. This saves memory and Grav delivers sites faster.


rhukster:
merging but will refactor a smidgen after.

PR #207, opened 2015-06-20
Modified `$page->summary()` to allow custom summary assignments

Sommerregen:
This is a PR to allow custom summary texts for pages. The PR is fully-backward compatible, but now accepts integers (for limiting the length of a summary) and strings (to set the summary), e.g.

```php
// Sets the summary
$page->summary('My superawesome summary');

// Gets the summary with custom size (here: with length 15)
$page->summary(15);

// Or return summary with default settings
$page->summary();
```

Further, even if one sets a custom summary, it will respect the settings made in the [`system/config/site.yaml`](https://github.com/getgrav/grav/blob/develop/system/config/site.yaml#L15) file.

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -361,7 +361,7 @@ class Page
             $var = 300;
         }
 
-        return Utils::truncateHTML($content, $size);
+        return Utils::truncateHTML($content, $var);
     }
 
     /**
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -307,11 +307,10 @@ class Page
     /**
      * Get the summary.
      *
-     * @param  int|string  $var  Summary or max summary size.
-     *
+     * @param  int    $size Max summary size.
      * @return string
      */
-    public function summary($var = null)
+    public function summary($size = null)
     {
         /** @var Config $config */
         $config = self::getGrav()['config']->get('site.summary');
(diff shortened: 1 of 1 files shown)


rhukster:
So, i've looked over this and there's one little thing that bothers me. 

Basically, if you call `summary(<string>)` it sets the summary, but if you call `summary(<int>)` it retrieves the summary but only with X characters.

I know this is mainly my fault by creating the Pages object with a single method to set and get values rather than separate setters and getters, but I did that intentionally to keep it clean and simple on the Twig side of things. Overall this has not proved a major issues, but in this case, where optional params to a getter are used, it makes a setter all but impossible.

So what you have works fine, does what you need, but is a little confusing.  I think the best option may be to do a nonstandard `setSummary(<string>)` method, this simply set `$this->summary` and would need to be called before `summary()` to get the value.  We already have a `setRawContent()` so it's not unprecendented, just not 100% inline with the rest of the methods.  

I think in a Grav 2.0 I would just go with regular getter/setters for all these methods...

Anyway, could you pull out that optional string var detection stuff, and just have a separate `setSummary()` method and i'll merge that.

Cheers!

Sommerregen:
Good point! I always try to keep things as untouched as possible, but you are right. I will change it tomorrow and implement besides a `setSummary(<string>)` a `getSummary()`, too. To make it consistent with you aforementioned `setRawContent()` method.

Sommerregen:
Ok, yesterday I was tired. No need for additional `getSummary()` method. Now, the code is refactored and the `setSummary(<string>)` method added.

However you could not have been foreseen the necessity of setting a custom summary, since by default Markdown is really limited and does not provide such a feature. Maybe there are some other changes needed in the future, which will get an unique interface in **Grav 2.0** then. Grav is evolving and it gets really powerful, also for me as a plugin developer. Thanks for all and for considering my PRs!

rhukster:
Thanks, this looks good now.

PR #211, opened 2015-06-26
Spelling corrections.

barryanders:


rhukster:
Wow, just goes to show you what happens when you turn off your spellchecker in your IDE! Thanks.

barryanders:
I don't use a spell checker normally, but I love what Grav is becoming so much that I wanted to contribute in some way, so that's what I came up with. Not much, but it's something.

PR #232, opened 2015-07-30
Refresh session timeout

flaviocopes:
Call setcookie() after the session is started to refresh the session timeout. This prevents the session to end when the user is still active.


flaviocopes:
PR updated :+1:

PR #249, opened 2015-08-06
Fix #248 (Plugin language interference) and fixed summary

### LINKED ISSUE #248: Plugin language interference
Dear Grav Team,

while implementing mult-language features in my plugins, I stumbled over a possible bug concerning language interference of two plugins. Consider for example you have `pluginA` with languages.yaml defined by

```yaml
en:
  PLUGINS:
    PLUGINA:
      TEXT: "Hello plugin A"
```

and another plugin called `pluginB` with languages.yaml defined by

```yaml
en:
  PLUGINS:
    PLUGINB:
      TEXT: "Hello plugin B"
```

Usually I then expect to get

```yaml
en:
  PLUGINS:
    PLUGINA:
      TEXT: "Hello plugin A"
    PLUGINB:
      TEXT: "Hello plugin B"
```

but instead, depending which plugin loads last, one gets one of the two first configurations (inspected by dumping `$grav['config']` to print out the available language entries). It seems like the join in https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Config/Config.php#L375  is the problem, which does effectively an `array_merge`. Instead an `array_merge_recursive` is necessary here.

Sommerregen:
This PR references https://github.com/getgrav/grav/issues/248. I further added a bugfix when calling `$page->summary` with system option `summary.enabled: false`.

### DIFF
--- a/system/src/Grav/Common/Config/Config.php
+++ b/system/src/Grav/Common/Config/Config.php
@@ -371,9 +371,7 @@ class Config extends Data
                 foreach ((array) $languageFiles['user/plugins'] as $plugin => $item) {
                     $lang_file = CompiledYamlFile::instance($item['file']);
                     $content = $lang_file->content();
-                    foreach ((array) $content as $lang => $value) {
-                        $this->languages->join($lang, $value, '/');
-                    }
+                    $this->languages->mergeRecursive($content);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


rhukster:
Yah saw that :)

Issue #274, opened 2015-08-21
Add the ability to use any date for a collection dateRange filter


akoebbe:
I'd like to be able to use an arbitrary date attribute (a custom one) on a page to use as the collection filter for list pages. Right now only the "date" field can be used.

I have an "event" page type that has an event.time (datetime) field that I want to only show upcoming events in a listing page.

akoebbe:
From my PR:

This feature allows the user to specify a custom field to use in the dateRange filter of a collection page. While the same thing is most likely possible to using the Publish and Unpublish Date fields on each child page, this removes the chance of errors and extra fields to fill out.

Use case: Display a list of "event" page types that only shows upcoming events.

Yaml usage:
```
content:
    items: @self.children
    dateRange:
        start: today
        field: header.event.time               // can use anything $page->value() will take
```

rhukster:
Thanks for your PR. merged!

vitopepito:
I experienced that when using a custom field for dateRange, it is not properly ordering the items when the european date format is used. For example an items YAML like: header.event.date: 30-03-2017 will somehow be treated as 30/03/2017. Ordering then follows Year -> Month -> Day even if Month = 30 does not exist. Issue does not occur when using default header.date for range.

akoebbe:
@vitopepito could you make a new issue for this bug. I'll see if I can dig in to it tomorrow.

PR #275, opened 2015-08-21
Feature/Allow the use of custom datetime fields for collection filtering

akoebbe:
Issue: #274 

This feature allows the user to specify a custom field to use in the dateRange filter of a collection page. While the same thing is most likely possible to using the Publish and Unpublish Date fields on each child page, this removes the chance of errors and extra fields to fill out.

Use case: Display a list of "event" page types that only shows upcoming events.

Yaml usage:
```yaml
content:
    items: @self.children
    dateRange:
        start: today
        field: header.event.time               // can use anything $page->value() will take
```

### DIFF
--- a/system/src/Grav/Common/Page/Collection.php
+++ b/system/src/Grav/Common/Page/Collection.php
@@ -216,14 +216,14 @@ class Collection extends Iterator
      * Dates can be passed in as text that strtotime() can process
      * http://php.net/manual/en/function.strtotime.php
      *
-     * @param      $field
      * @param      $startDate
      * @param bool $endDate
+     * @param      $field
      *
      * @return $this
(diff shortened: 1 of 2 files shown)


rhukster:
Can you move the `$field = false` from the start of the method signature to the end?  I just have a problem with changing the signatures in a non-backwards compatible way.  Even if this is only used the once, it's just bad precedence that can cause issues if it's used by another plugin or something.

Thanks!

akoebbe:
Will do. I had a bit of an uneasy feeling about that, as well. Thanks for the confirmation.

akoebbe:
Done.

rhukster:
Thanks! Looks good.

PR #276, opened 2015-08-21
Fix error from grav-plugin-admin issue # 80

Sommerregen:
This change fixes an issue connected with https://github.com/getgrav/grav-plugin-admin/issues/80 that occurred when installing plugins removing the duplicate slash in the beginning due to https://github.com/getgrav/grav/blob/develop/system/defines.php#L17 .


rhukster:
nice find. thanks!

PR #284, opened 2015-08-26
Fix issue with date format for PHP 5.1.0+

mufac:
As of PHP 5.1.0 the character 'o' represents the ISO-8601 year number in a date format. Therefore the 'o' needs to be escaped to avoid being converted to a year in the admin > configuration > system > long/short date format pulldown menu.

I should note that this fixes the issue for English, but you might want to consider moving the date format options into a language string in order to provide better multi-language support.


rhukster:
Good catch!

rhukster:
Need to merge first! :)

PR #288, opened 2015-08-29
Solve an issue with apostrophes in the taxonomy name

flaviocopes:
Fixes https://github.com/getgrav/grav-plugin-admin/issues/123

When a tag contains an apostrophe it's encoded and it cannot not match the page taxonomy items unless decoded.


rhukster:
Looks ok to me!

PR #292, opened 2015-08-30
.htaccess hardening

fantasticme:
Fork updated


rhukster:
Can you add/sync these changes to `.htaccess` also?  Need to keep these in sync (i think i forgot to sync backup before)

rhukster:
Hmm not sure this worked properly :) seeing no changes in the last ea5c44f commit. And a bunch of merges before that.

PR #293, opened 2015-08-30
Fork updated

fantasticme:
Should work now


rhukster:
So strange, i'm expecting to see an `.htaccess` and `htaccess.txt` commit.. Instead i'm seeing all the release merges and just `.htaccess` now.  Not sure what's going on :(

fantasticme:
ah wait, i try again

PR #297, opened 2015-09-03
Correctly generate 2x image too and make it available to srcset, fixes #133

flaviocopes:


flaviocopes:
Fixes https://github.com/getgrav/grav-plugin-admin/issues/133

PR #310, opened 2015-09-08
Fix bug in newuser command

mufac:
grav newuser was setting password to null irrespective of what was entered at the console


rhukster:
Doh!

mufac:
Did you mean to merge this before closing?

rhukster:
man. i need to get more sleep!

Issue #319, opened 2015-09-13
GPM: eval not found on Windows system


namaless:
https://github.com/getgrav/grav/blob/develop/system/src/Grav/Console/Gpm/InstallCommand.php#L105

We need to validate before if is a windows system or not before or give a notice (the process continue without problem).

nunull:
Sorry for asking: What is that line supposed to do anyway? Isn't the result the same as doing `$local_config_file = '~/.grav/config';`. If so, this seems entirely unreasonable to me.

namaless:
Thanks for the other line found.

http://www.unix.com/man-page/posix/1posix/eval/

rhukster:
That line is used when installing via symbolic links (via `-s`) and is used by the Grav development team. We're all on macs (with one linux user) so this is not a big deal.

rhukster:
BTW, if you have GitBash installed, there is a windows compatible version of `eval` provided by that, so it works fine. That's why if your running windows, I suggest running under GitBash to make it a more 'nix' friendly environment.

namaless:
Isn't a big deal but really bad to see the notice :)

Btw repeat: the process continue without problem :)

rhukster:
If you want to to add more 'windows' friendly process if windows platform is detected, i'm fine with that.  It's just not something I'm able to do as I'm a mac user.

namaless:
I found this: https://gist.github.com/asika32764/90e49a82c124858c9e1a

In my system, Windows 10, I have "WINNT" for PHP_OS constant.

nunull:
I don't understand why `$x = exec('eval echo x');` is used instead of `$x = 'x';`. Isn't the first really just constructing a string in a pretty unreadable and error-prone way?

rhukster:
Thanks guys!

PR #321, opened 2015-09-13
Patch 1

namaless:
Reference to https://github.com/getgrav/grav/issues/319


rhukster:
Thanks! Looks good to me!

PR #326, opened 2015-09-16
Fix chaining for sizes and derivatives

attiks:
Quick fix for the chaining of sizes and derivatives, so the order does not matter.


rhukster:
will be in next release

PR #335, opened 2015-09-27
Add CSS media attribute

Sommerregen:
This PR adds the ability to set custom media attributes to CSS stylesheets via e.g. 

```twig
{{ do assets.addCss('theme://css/print.css', 100, null, 'print') }}
```

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -227,21 +227,23 @@ class Assets
      * It checks for duplicates.
      * You may add more than one asset passing an array as argument.
      *
-     * @param  mixed $asset
-     * @param  int   $priority the priority, bigger comes first
-     * @param  bool  $pipeline false if this should not be pipelined
+     * @param  mixed  $asset
+     * @param  int    $priority the priority, bigger comes first
+     * @param  bool   $pipeline false if this should not be pipelined
+     * @param  string $media    specifies what media/device the target
+     *                          resource is optimized for
      *
      * @return $this
      */
-    public function addCss($asset, $priority = null, $pipeline = null)
+    public function addCss($asset, $priority = null, $pipeline = null, $media = null)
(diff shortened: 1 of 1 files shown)


rhukster:
Could you rewrite this to use the array attribute style rather than override the method signature. I recently updated these methods to support arrays, and this is the better way to add new features and functionality.

rhukster:
Oh actually you did, i think we just not include it in the method signature.

Sommerregen:
Can you clarify this point? You mean https://github.com/getgrav/grav/pull/335/files#diff-036bc801ec8840ea4247592b9ded7a66R267 ?

Ahh, again an undocumented feature. You mean using it via, 

```php
$grav['assets']->addCss('mytheme.css', ['media' => 'screen']);
```

supersedes my PR.

rhukster:
Yes, I documented it last week actually: http://learn.getgrav.org/themes/asset-manager#options

Sommerregen:
Ah, too bad... Will close it.

rhukster:
Oh no, it doesn't supercede your PR, it does not add random attributes.  Your PR bit that takes the value from the array and adds it in the HTML is still valid. Just I don't want to add these to the method signature as we find more things to add.

Sommerregen:
Ok, and what should I gonna do? Leave it open for moment? Change the code?

rhukster:
Yah change the code and remove the method signature changes, and just check to see if its in the array, and if it is, add it to the HTML as you have it.

Sommerregen:
Ok, done!

rhukster:
Just minor, when you get $media from the array, if it's empty you are setting $media to a space, won't this add an extra space that is not needed?

Sommerregen:
Yah, you are right. Can you correct this after merging?

rhukster:
yah no problem

Issue #154, opened 2015-03-14
Theme Inheritance Breaks Down If The Parent Needs A Custom Plugin


bobrocke:
In the case of the Materialize theme, the Materializer plugin is needed for functionality. I accomplished that with code in `materialize.php`:
```
<?php
namespace Grav\Theme;

use Grav\Common\Theme;

class Materialize extends Theme
{
	// Materializer plugin will look for this class var to know it should load
	public $load_materializer_plugin = true;

	public static function getSubscribedEvents()
	{
		return [
			'onThemeInitialized' => ['onThemeInitialized', 0]
		];
	}

	public function onThemeInitialized()
	{
		if ($this->isAdmin()) {
			$this->active = false;
			return;
		}

		$this->enable([
			'onPageInitialized' => ['onPageInitialized', 0]
		]);
	}

	public function onPageInitialized()
	{
		$this->config->set('plugins.pagination.built_in_css', false);
		$this->config->set('plugins.relatedpages.show_score', false);
   }

}
```
Child themes don't inherit this code, so don't handle plugins properly.

A workaround is to replicate the parent theme's `.php` file functions in the child theme's `.php` file.

Seebz:
I have not tested but maybe the better way can be to extend the parent class in your child theme.

Something like this :
```
<?php
namespace Grav\Theme;

class Mytheme extends Materialize
{

}
```

(sorry for my bad english)

rhukster:
Should be sorted with @Sommerregen 's PR

PR #353, opened 2015-10-12
Feature/theme inheritance

Sommerregen:
This PR fixes an unreferenced bug, when dealing with language files in child themes; if you don’t provide a language file, it now falls back properly to the parent theme language file or replaces/adds some keys from the parent theme language file if present.

Further this PR finally addresses #154 and finalize somehow the way themes are inherited. With this PR it is now possible to derive the child theme by extending the parent theme class e.g.

```php
namespace Grav\Theme;

class Mytheme extends Antimatter
{
  // Some new methods, properties etc.
}
```

This is accomplished by prociding a (simple and fast) autoloader mechanism for theme instantiation, which only gets active if a theme under the namespace `Grav\Theme` is constructed.

### DIFF
--- a/system/src/Grav/Common/Themes.php
+++ b/system/src/Grav/Common/Themes.php
@@ -226,7 +226,11 @@ class Themes extends Iterator
         $config->joinDefaults("themes.{$name}", $themeConfig);
 
         if ($this->config->get('system.languages.translations', true)) {
-            $languages = CompiledYamlFile::instance("themes://{$name}/languages". YAML_EXT)->content();
+            $languages = [];
+            $schemes = array_reverse($config->get("themes.{$name}.streams.schemes.theme.prefixes.", []));
+            foreach ($schemes as $scheme) {
+                $languages = array_replace_recursive($languages, CompiledYamlFile::instance("$scheme/languages". YAML_EXT)->content());
+            }
             if ($languages) {
                 $config->getLanguages()->mergeRecursive($languages);
             }
--- a/system/src/Grav/Common/Themes.php
+++ b/system/src/Grav/Common/Themes.php
@@ -27,6 +27,9 @@ class Themes extends Iterator
     {
         $this->grav = $grav;
         $this->config = $grav['config'];
+
+        // Register instance as autoloader for theme inheritance
+        spl_autoload_register([$this, 'autoloadTheme']);
     }
 
     public function init()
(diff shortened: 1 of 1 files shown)


mahagr:
Why don't you just use composer autoloader, it's available for everyone in CI.

    $loader = $grav['loader'];
    $locator = $grav['locator'];
    $loader->addPsr4("Grav\\Theme\\', $locator->findResources("theme://"), 'src');

Above would locate class `Grav\Theme\MyClass` from `user/themes/xxx/src/MyClass.php`

mahagr:
Errr, OK, I read the text again and looks like you meant extending theme class itself and not adding new classes. Composer loader is still available, though.

Sommerregen:
@mahagr I was thinking about using Composer, too. But I really don't know, whether it can load classes (here themes), which were added by the user later. Otherwise the built-in autoloader is totally ok. For theme inheritance we just have to make sure that the parent theme class is loaded correctly. It doesn't matter how. Thus, I will test your idea and see, whether this works here (better than adding another autoloader).

mahagr:
What I do in Gantry 5 is something a bit different -- we have the same need in there.

Instead of relying a class to load the theme, I've added a new yaml file called `gantry/theme.yaml` which gets loaded before anything else. It contains all the customizable information including base theme, used engine, base and overrides etc. Those informations are used to initialize the streams which makes theme inheritance really easy to implement. The hardest thing is to build those streams and allowing multiple levels of inheritance, but it has all been implemented in Gantry...

rhukster:
We can look at improving theme inheritance in a later version.  What we have today is used by lots of people, we could add an alternative solution but would still need to maintain the current one.

Sommerregen:
I totally agree with @rhukster . In my point of view Gantry shouldn't be a part of Grav, since it will blow up the slim Grav core in an unecessary way (and is may be not used by everyone). I see Gantry more like a plugin but for themes. We may take some improvements, when it is time to do so. 

For the moment I will check out your idea @mahagr and push some changes if necessary.

mahagr:
I didn't say that we should force Grav to use Gantry, just that they both are derived from the same code making it easy to borrow some code from one and put it into another. :)

Sommerregen:
@mahagr Don't worry. I didn't mean it like this. You are totally right. Borrowed something, which is already implemented (and maybe well tested) helps a lot maintaining both projects. I just wanted to say, that we definitely need here an alternative approach for the moment. And later we can build something up on top of that (like integrating Gantry's way of theme inheritance) :-) It shouldn't be any problem though, since I'm sure both approaches don't conflict with each other.

PR #354, opened 2015-10-13
Allow override of directory constants

hwmaier:
This change allows to override and change some of the directory names from a customized index.php entry script:

Example:
In index.php before registering the autoloader one now can re-define the name and location of the cache or image directory like this:

    define('ROOT_DIR', __DIR__ . '/');
    define('CACHE_DIR', ROOT_DIR . 'site-cache/');
    define('IMAGES_DIR', ROOT_DIR . 'img/');

### DIFF
--- a/system/defines.php
+++ b/system/defines.php
@@ -6,19 +6,17 @@ define('GRAV_VERSION', '0.9.45');
 define('DS', '/');
 
 // Directories and Paths
-if (!defined('GRAV_ROOT')) {
-    define('GRAV_ROOT', str_replace(DIRECTORY_SEPARATOR, DS, getcwd()));
-}
-define('ROOT_DIR', GRAV_ROOT . '/');
-define('USER_PATH', 'user/');
-define('USER_DIR', ROOT_DIR . USER_PATH);
-define('SYSTEM_DIR', ROOT_DIR .'system/');
-define('ASSETS_DIR', ROOT_DIR . 'assets/');
-define('CACHE_DIR', ROOT_DIR . 'cache/');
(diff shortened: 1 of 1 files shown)


mahagr:
Grav uses streams and its fully possible to relocate most of the directories by using `setup.php` file. Those constants should really go away in long term as they do not give the same extensibility as streams do.

hwmaier:
I agree that the streams are the way to go but until it is implemented fully, the only way is to change these constants. Currently this requires editing a system file `defines.php` and the change is lost after an update.

ROOT_DIR, USER_DIR, CACHE_DIR, LOG_DIR, SYSTEM_DIR and ASSETS_DIR  cannot be changed through streams at present.

Apparently the PAGES_DIR and IMAGES_DIR constants are not used in the source any more and maybe should be moved into the DEPRECATED section.

rhukster:
I appreciate this PR, but I think the better solution will be to finish implementing streams throughout Grav rather than more deeply entrenching ourselves with the older constant approach. 

Providing 100% streams support has been on our list of things to do for some time, we just need to get it done! :)

hwmaier:
Ok, let's drop this PR then and focus on a more future-proof solution

niahoo:
Hey guys, 

I would like to know if with the current stable version, is it ok to change `USER_DIR` to another directory in the filesystem (serving files from Dropbox). I know I would have to change it after upgrades.

Thank you !

mahagr:
Yes, but you also need to change `user://` stream as well.

niahoo:
Ok thank you !

PR #363, opened 2015-10-16
Update Uri.php

cord:
fix for override of alias on localhost

### DIFF
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -69,7 +69,12 @@ class Uri
         $address = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : '::1';
 
         // check for localhost variations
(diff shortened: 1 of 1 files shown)


rhukster:
I'm just not comfortable making this change as although you are using it with multiple local aliases and running into issues because all are bound to 127.0.0.1, changing the behavior would impact other people that are relying on that functionality.

However, there is a simple solution.  Simply bind your various aliases to other local ip addresses.  On a mac this is pretty easy:

1. Duplicate the primary interface via the gear in the Network panel of System Preferences
2. Use a private IP address, for example: `192.168.2.2`
3. Add a host entry for your internal hostname and this ip: `grav.de    192.168.2.2`
4. bind this IP to the vhost entry and restart Apache:

```
  <VirtualHost 192.168.2.2:80>
    DocumentRoot "/Users/cord/www/grav-site"
    ServerName grav.de
  </VirtualHost>
```

Now when you go to `http://grav.de` in your browser, you are not going to be aliased to 127.0.01, and your environment will be `grav.de` to be used for all configurations etc.

PR #365, opened 2015-10-16
add specific links to 'adding functionality' intro

aranaur:
add links directly to the tabs for plugins and themes on getgrav.org. The first time I loaded the main downloads page I didn't see the plugins/themes tabs because I immediately started scrolling down. It might be worth adding to the bottom of that main page a quick line to the tune of "Looking for plugins or themes? Click here or here."


flaviocopes:
Thanks!

PR #370, opened 2015-10-19
Use media.yaml to determine downloaded files mime type

flaviocopes:


rhukster:
This is probably a good idea for a number of reasons, but there are a couple of issues.  

Basically anything that is an 'image' should be able to have the image manipulations applied to it. So BMP, Tiff, etc are not valid images in in this sense.  Same goes for Videos, they must be able to be supported with the <video> tag. -> http://www.w3schools.com/html/html5_video.asp, so basically that's only: MP4, WebM, and Ogg

Everything else should be under 'files' which are accessible via media['files'] but not much can be done with them outside of getting their URL or a simple HTML link

flaviocopes:
Ok added the correct mime types, should be good to go

PR #372, opened 2015-10-21
Allow setting of redirect http response code through config.system.pa…

hwmaier:
If `redirect_default_route` is set to true, the redirects are using 303 HTTP response codes.

For SEO purposes it may be beneficial to issue 301 repsonse codes for route alias redirects.

The `redirect_default_route` can now be also set to a numeric value reflecting the desired response code. example:
```
pages:
  redirect_default_route: 301
```

This makes routes defined using the `route: alias` header use 301 as http response code.

### DIFF
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -131,8 +131,14 @@ class Grav extends Container
                     }
                 }
                 // Default route test and redirect
-                if ($c['config']->get('system.pages.redirect_default_route') && $page->route() != $path) {
-                    $c->redirectLangSafe($page->route());
+                $redirect_default_route = $c['config']->get('system.pages.redirect_default_route');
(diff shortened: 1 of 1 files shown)


hwmaier:
I just realised there are other re-direct scenarios which also should be configurable between 303 and 301. What about `site.redirects`  and simple page `header.redirect`?

Maybe there is a better way to configure between 303 and 301 for `redirect_default_route`, `site.redirects`  and simple page `header.redirect`.

rhukster:
2015-10-22-13-bo748

I'm going to handle this a slightly different way.  Will let you configure the code at a system level independent of that route setting.. This will impact all redirects which I think is more useful.

rhukster:
Also I've added the ability to append an HTTP code at the any of any URL eg:

```
redirect: /some/path[303]
```

or 

```
redirect: http://somesite.com[307]
```

PR #379, opened 2015-10-25
html method adds automatically alt and title attributes if defined in…

hwmaier:
… media file's meta.yaml file

### DIFF
--- a/system/src/Grav/Common/Page/Medium/Medium.php
+++ b/system/src/Grav/Common/Page/Medium/Medium.php
@@ -216,6 +216,10 @@ class Medium extends Data implements RenderableInterface
         !empty($title) && empty($attributes['title']) && $attributes['title'] = $title;
         !empty($alt) && empty($attributes['alt']) && $attributes['alt'] = $alt;
         !empty($class) && empty($attributes['class']) && $attributes['class'] = $class;
+        if (empty($attributes['title']) && !empty($this->items['title']))
+            $attributes['title'] = $this->items['title'];
+        if (empty($attributes['alt']) && !empty($this->items['alt_text']))
+            $attributes['alt'] = $this->items['alt_text'];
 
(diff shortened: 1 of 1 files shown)


rhukster:
Is this intended to add them if they are defined? or saying they are already added and you don't want them to be? Your title is not clear :)

hwmaier:
This is a proposal for a new feature.

Currently alt and title tags have to be specified explicitly as parameter to the html call. But we also have the great meta.yaml feature were they can be specified in a companion file for each media file.

If you want to use the alt and title attributes specified in the meta.yaml file, then one has to assemble the img tag manually.

This PR would embed these attributes automatically should they be specified in the meta.yaml file and not specified as part of the html function call.

rhukster:
BTW for simplicity I propose:

```
        if (empty($attributes['title'])) {
            if (!empty($title)) {
                $attributes['title'] = $title;
            } elseif (!empty($this->items['title'])) {
                $attributes['title'] = $this->items['title'];
            }
        }

        if (empty($attributes['alt'])) {
            if (!empty($alt)) {
                $attributes['alt'] = $alt;
            } elseif (!empty($this->items['alt'])) {
                $attributes['alt'] = $this->items['alt'];
            }
        }

        if (empty($attributes['class'])) {
            if (!empty($class)) {
                $attributes['class'] = $class;
            } elseif (!empty($this->items['class'])) {
                $attributes['class'] = $this->items['class'];
            }
        }
```

hwmaier:
Your re-factored version is much nicer and better to understand.

rhukster:
Fixed in develop with adjusted version. Thanks!

hwmaier:
Would it make sense to even go a step further and add width and height as well?

This would be however much trickier as width/height is only known of the unmodified image. Once resize actions have been added, those attributes are invalid for the modified incarnation of the image.

PR #386, opened 2015-10-27
onTwigPageVariables event handler gets Page object passed as Event parameter

hwmaier:
This patch passes the Page object down to the onTwigPageVariables event handler of a plugin so it can determine and manipulate the page headers of the corresponding page object. It is otherwise not possible to access page headers of modular pages because `$this->grav['page']` will point always to the main page not the modular subpage.

Refer to discussion here: http://getgrav.org/forum#!/getgrav/plugin-development:ontwigpagevariables-event-h::unread


rhukster:
Thanks! should be useful!

PR #387, opened 2015-10-28
Asset management uses now streams rather the constant ASSETS_DIR

hwmaier:
Asset management uses now streams rather the constant ASSETS_DIR to configure location of pipelined asset files. In addition it supports sub-directories for assets location, for example `assets/runtime`.

The assets directory can now be re-configured changing the stream in `streams.yaml` Example:

```
schemes:
  asset:
    type: ReadOnlyStream
    paths:
      - assets/runtime
```

### DIFF
--- a/system/defines.php
+++ b/system/defines.php
@@ -13,14 +13,14 @@ define('ROOT_DIR', GRAV_ROOT . '/');
 define('USER_PATH', 'user/');
 define('USER_DIR', ROOT_DIR . USER_PATH);
 define('SYSTEM_DIR', ROOT_DIR .'system/');
-define('ASSETS_DIR', ROOT_DIR . 'assets/');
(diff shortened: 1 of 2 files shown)


flaviocopes:
Good to merge for me.

@rhukster can we now also drop

```php
define('ASSETS_DIR', ROOT_DIR . 'assets/');
define('IMAGES_DIR', ROOT_DIR . 'images/');
define('ACCOUNTS_DIR', USER_DIR .'accounts/');
define('PAGES_DIR', USER_DIR .'pages/'); 
```

? 

They're not referenced anywhere in the core and all the code that was using them is using streams now. Or BC is an issue?

hwmaier:
I left the defines in the code but moved them to the deprecated section. Its up to you guys as maintainers to decide to clean them out. Maybe some plugins could still referencing them? A quick grep through the plugin repository should clarify this.

PR #396, opened 2015-10-29
Fix #395

### LINKED ISSUE #395: Problem with default language and slugs that is starting with language name
I have multilingual site that have many languages (en, fi, et, de ...). Let's play that english is a default language of site so English version of site is running in url http://domain.com and other languages get language slug after it (http://domain.com/fi etc.). 

Something strange happens if I create a page whose slug is for example "english" (http://domain.com/english) or something else that starts with string "en". Site gives me a 404 error and it seems that Grav is try to get pages named "glish". I debug this little and Grav debugger is telling me like this:

Grav\Common\Uri {#50
  +url: "http://localhostglish"
  #basename: "glish"
  #base: "http://localhost"
  #root: "http://localhost"
  #bits: null
  #extension: null
  #host: "localhost"
  #content_path: "glish"
  #path: "glish"
  #paths: array:1 [
    0 => "glish"
  ]
  #query: []
  #params: []
}

Is this a bug or am I missing something?

Sommerregen:
This PR fixes #395 (Problem with default language and slugs that is starting with language name), which anchors the regex such that the language from the uri is only stripped, when a slash is following with the path to the page or the uri ends there.

### DIFF
--- a/system/src/Grav/Common/Language/Language.php
+++ b/system/src/Grav/Common/Language/Language.php
@@ -162,7 +162,7 @@ class Language
      */
     public function setActiveFromUri($uri)
     {
-        $regex = '/(^\/(' . $this->getAvailable() . ')).*/';
+        $regex = '/(^\/(' . $this->getAvailable() . '))(?:\/.*|$)/i';
 
         // if languages set
         if ($this->enabled()) {


flaviocopes:
Thanks!

PR #398, opened 2015-10-29
Remove image location

Angristan:
location /images/ is empty !


rhukster:
I've refactored the nginx.conf

Issue #404, opened 2015-10-31
Default routes with trailing slash end up in infinite redirect loop


hwmaier:
This occurs when `redirect_default_route` is set to `true`. The setting of `redirect_trailing_slash` does not matter for this to occur.

When a page's route is defined with a trailing slash and its URL is requested, Grav ends up in a redirect loop.

```
---
routes:
  default: /support/
---
```

In class Grav in the page function line 117, the trailing slash is removed.

Later in same function line 134 the path is tested against the default route and if no match, it redirects.
Because the default route has a slash but the path not, we end up in an infinite redirect loop.

rhukster:
I can see this.  routes should not have  a trailing slash and we probably should trim them anyway.

Also wouldn't hurt to have logic to take trim that trailing slash in the redirect also.

hwmaier:
I personally do not like routes with a trailing slash, but sometimes you want to be able migrate existing sites to Grav and keep existing and long established URLs and bring all the page SEO juice across as well.

So unless Grav can support trailing slash URLs I would not be able to use it for some sites.

hwmaier:
I was proactive and submitted a PR. See https://github.com/getgrav/grav/pull/405

rhukster:
This has been merged.  Thanks!

PR #405, opened 2015-10-31
Added support for default routes with a trailing slash

hwmaier:
This PR enables Grav to specify URLs with a trailing slash for a page, by means of a default route with a trailing slash.  Such a feature shall help migrating legacy sites to Grav while maintaining existing and well established URLs which may have trailing slashes.

Refer to https://github.com/getgrav/grav/issues/404

Example 
```
---
routes:
  default: /support/
---
```
A request to this page will NOT redirect to `/support` and resolve as `/support/`
This feature works with `redirect_default_route` either on or off and if  `redirect_default_route` is on a request to `/support` redirects to `/support/` otherwise the page is served as is.

`redirect_trailing_slash` must be set to false for this use case for obvious reasons as this would conflict.

### DIFF
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -114,7 +114,7 @@ class Grav extends Container
             /** @var Uri $uri */
             $uri = $c['uri'];
 
-            $path = rtrim($uri->path(), '/');
+            $path = $uri->path(); // Don't trim to support trailing slash default routes
             $path = $path ?: '/';
 
             $page = $pages->dispatch($path);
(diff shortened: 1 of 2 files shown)


rhukster:
I want to do some testing of my own with this one as it's quite a big change on how trailing slashes are handled.

hwmaier:
Of course a change like this needs proper review and thorough testing, I understand that. 
Is there an automated test harness once could employ?

rhukster:
I'm going to merge and then test and fix if required.. but looks good in my online review :)

PR #408, opened 2015-11-01
Use streams instead of hardcoded paths for clearCache()

hwmaier:
If the asset directory was relocated using streams, clearCache would delete the wrong directory.

### DIFF
--- a/system/src/Grav/Common/Cache.php
+++ b/system/src/Grav/Common/Cache.php
@@ -248,7 +248,7 @@ class Cache extends Getters
         foreach ($remove_paths as $stream) {
 
             // Convert stream to a real path
-            $path = $locator->findResource($stream);
+            $path = $locator->findResource($stream, true, true);
             // Make sure path exists before proceeding, otherwise we would wipe ROOT_DIR
             if (!$path)
                 throw new \RuntimeException("Stream '{$stream}' not found", 500);
--- a/system/src/Grav/Common/Cache.php
+++ b/system/src/Grav/Common/Cache.php
@@ -21,6 +21,8 @@ use Grav\Common\Filesystem\Folder;
  */
 class Cache extends Getters
 {
+    use GravTrait;
+
     /**
      * @var string Cache key.
      */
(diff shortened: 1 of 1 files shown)


rhukster:
Nice work!

rhukster:
BTW @hwmaier , just tested this and it does not work with `bin/gpm clear` in the console.  

```
  [InvalidArgumentException]
  Invalid resource asset://
```

If you don't get chance to look at it, I should have a bit of time tomorrow to take a peek

hwmaier:
gpm clear ? Is that a new command or do you refer to grav clean ?

hwmaier:
Issue confirmed here with `grav clear-cache`. I'll have a look.

hwmaier:
I tested it using the cachebuster plugin before submitting the PR.

With grav CLI, the `streams.yaml` file is not loaded, and there is no default stream set in Config.php for assets.

We have to add an entry for assets to Config.php I reckon.

rhukster:
I will talk to matias tomorrow about the reasoning behind having some streams defined in streams.yaml and some in Config.php as he did that implementation.  I've run into issues like this a few times, seems that all the streams should be loaded together, and if they are not, what this reasoning behind it.

hwmaier:
Good. Even if we now defined a default in Config.php, then issues arise if the override in streams.yaml does not match the same value. A stream must either be only defined in Config.php or setup.php OR in streams.yaml but not in both locations.

hwmaier:
Reasoning for separation is bootstrapping. You need config streams to load stream.yaml.

rhukster:
Ok should be working.. just  moved the stream to config and tweaked the cache command as it was not working before.

hwmaier:
Looks good. Maybe as a convention:

- Streams used in CLI and during bootstrapping must be defined in Config.php and overrides must be added to setup.php
- Other streams can be defined and overriden in streams.yaml.
- A stream definition must either live in group `Config.php / setup.php` or in group`system/config/streams.yaml / user/config/streams.yaml` but not in both groups..

PR #426, opened 2015-11-06
Add nonce functionality

flaviocopes:
# Grav Nonces

A nonce is a "number used once" to help protect URLs and forms from certain types of misuse, malicious or otherwise. Grav nonces aren’t numbers, but are a hash made up of numbers and letters. Nor are they used only once, but have a limited "lifetime" after which they expire. During that time period the same nonce will be generated for a given user in a given context. The nonce for that action will remain the same for that user until that nonce life cycle has completed.

Grav’s security tokens are called "nonces" despite the above noted differences from true nonces, because they serve much the same purpose as nonces do. They help protect against several types of attacks including CSRF, but do not protect against replay attacks because they aren’t checked for one-time use.

## Creating a nonce

You can create a nonce and add it to the query string in a URL, you can add it in a hidden field in a form, or you can use it some other way.
Note that the nonces are unique to the current user’s session, so if a user logs in or out asynchronously any nonces on the page will no longer be valid. 

## Adding a nonce to a URL

To add a nonce to a URL, call `Uri::addNonce()` specifying the bare URL and a string representing the action. For example:

	$url_with_nonce = $this->grav['uri']->addNonce($url, $action);

For maximum protection, ensure that the string representing the action is as specific as possible.

By default, `Uri::addNonce()` adds a field named `nonce`. You can specify a different name in the function call. For example:

	$complete_url = $this->grav['uri']->addNonce( $url, 'my-action', 'my_nonce' );

## Adding a nonce to a form

To add a nonce to a form, use the `{{ nonce_field('action') }}` Twig function.

For maximum protection, ensure that the string representing the action is as specific as possible.

You can specify a different name for the nonce field by adding that as a second parameter:

	{{ nonce_field('action', 'my_nonce') }}	

## Creating a nonce for use in some other way

To create a nonce for use in some other way, call `Utils::createNonce()` specifying a string representing the action. For example:

	$nonce = Utils::createNonce('save-page' . $page->title());

This simply returns the nonce itself. For example: `yEFaN4G4qT`

For maximum protection, ensure that the string representing the action is as specific as possible.

# Verifying a nonce

To verify a nonce, call `Utils::verifyNonce()` specifying the nonce and the string representing the action. For example:

	Utils::verifyNonce($nonce, $action);

If the result is false, do not continue processing the request, instead return a `401 Unauthorized` error.

# Nonce lifetime

By default, a nonce has a lifetime of one day. After that, the nonce is no longer valid even if it matches the action string. To change the lifetime, add a nonce_life filter specifying the lifetime in seconds. For example, to change the lifetime to four hours:

Grav uses a system with two ticks (half of the lifetime) and validates nonces from the current tick and the last tick. In default settings (24h lifetime) this means that the time information in the nonce is related to how many 12h periods of time have passed since Unix epoch. This means that a nonce made between midday and midnight will have a lifetime until midday the next day. The actual lifetime is thus variable between 12 and 24 hours.

# Nonce security

Nonces are generated via BCrypt using a salt that should be unique to your site. Customize it in your `user/config/system.yaml` file `security.default_hash` parameter.


flaviocopes:
Do not merge yet

hwmaier:
Currently all Grav installs out there use the same security.default_hash. So far this was not an issue as the hash wasn't really used for anything security related, only to perform a time-consuming calculation to prevent timing attacks. The value of the hash did not really matter for this.

Once this hash is going to be used as a seed for security related features it should be ensured that every Grav install has its unique hash. This could be enforced by the problems plugin which would test that this hash is not the standard hash from the repository.

Until the uniqueness is guaranteed or enforced I would be careful to use security.default_hash  for anything security related.

flaviocopes:
Since Grav does not have an installer we should either make it clear in the docs, or use the Problems plugin, or create a bin/grav command, or execute the new hash generation automatically when performing an operation such as creating the first user. Or add a notice to the admin plugin (throwing in ideas)

hwmaier:
Can I assume that the nonce feature will be used for CSFR protection of form posts?

rhukster:
That is the intention :)

flaviocopes:
Once this is approved and merged there's a PR already on the Form plugin to bring automatic protection on forms based on the Form plugin, and the same methods can be used on any form.

hwmaier:
Btw, this in conjunction with the form plugin is a great feature I was hoping Grav would have at some point, thanks for adding it.

hwmaier:
I am currently testing the new nonce feature and realised it depends on the login plugin. 
This could be avoided if in method `generateNonceString` access to `self::getGrav()['user']` is checked for existence first:
```
        if (isset(self::getGrav()['user'])) {
            $user = self::getGrav()['user'];
            $username = $user->username;
        }
        else
            $username = false;

```

flaviocopes:
Good tip, thanks!

PR #434, opened 2015-11-10
remove duplicate key

yaman-jain:
removing duplicate key from an array and using relative base class name.
removing unused 'use' statements.


rhukster:
Thanks!

PR #438, opened 2015-11-11
Debugger tool and twig dump

muuvmuuv:
I'm working with twig since yesterday and never heart about it before but love it at the same moment.
The time I've worked on a new theme for GRAV so I#m using the debugger tool a lot. Even I dump some loops result or something with many resultings I want to have some preformatted results inside the debugger tool. So for now its pretty awesome just to have something like this in HTML. But I want more instead of just giving me the full array as string back. In the JS console you can click on arrows and dive deeper and deeper in objects. It could be nice if this can be done in the debugger tools to.

Hope you could understand me.
Have a nice day.


flaviocopes:
I think you opened a Pull Request by mistake, right?
You can use {{ dump(variable) }} and that will be printed in the Grav Debugger (http://learn.getgrav.org/advanced/debugging)

PR #449, opened 2015-11-17
Added a translation for "Validation failed:" text

bovisp:
The English term "Validation failed:" was hard coded on line 83. Instead, I created a translation and placed this at /system/languages/en.yaml. It will need to be translated into other languages by others.

### DIFF
--- a/system/src/Grav/Common/Data/Blueprint.php
+++ b/system/src/Grav/Common/Data/Blueprint.php
@@ -72,12 +72,15 @@ class Blueprint
     {
         // Initialize data
         $this->fields();
+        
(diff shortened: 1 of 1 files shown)


system/src/Grav/Common/Data/Blueprint.php
```diff
@@ -72,12 +72,15 @@ public function validate(array $data)
     {
         // Initialize data
         $this->fields();
+        
+        // Get language class
+        $language = self::getGrav()['language'];
```
flaviocopes: Can you move this inside the catch statement? As it's used only there

PR #451, opened 2015-11-17
PhpDoc

yaman-jain:
updated/added PhpDoc for methods


flaviocopes:
Thanks, just one thing, can you edit it so callbacks have this definition, instead of being of type `null`?

`@param callable $callback Either a function or callback in array notation`

(taken from https://github.com/yaman-jain/grav/blob/docfixes/system/src/Grav/Common/GPM/GPM.php#L38)

flaviocopes:
Thanks!

yaman-jain:
:+1:

PR #475, opened 2015-11-24
Proper handling of list fields

nazwa:
Moved field handling to a separate function and added logic to correctly process nested lists. 

Code diff looks way more complex than it actually is. Copied everything that originally was under `else if ($field['type'] !== 'ignore') {` into a new function, so that it can be called from multiple places.

Fixes https://github.com/getgrav/grav-plugin-admin/issues/273

### DIFF
--- a/system/src/Grav/Common/Data/Blueprint.php
+++ b/system/src/Grav/Common/Data/Blueprint.php
@@ -314,91 +314,115 @@ class Blueprint implements \ArrayAccess, ExportInterface
     }
 
     /**
-     * Gets all field definitions from the blueprints.
-     *
-     * @param array $fields
-     * @param array $params
-     * @param string $prefix
-     * @param array $current
-     * @internal
-     */
-    protected function parseFormFields(array &$fields, $params, $prefix, array &$current)
-    {
-        // Go though all the fields in current level.
-        foreach ($fields as $key => &$field) {
-            $current[$key] = &$field;
-            // Set name from the array key.
-            $field['name'] = $prefix . $key;
-            $field += $params;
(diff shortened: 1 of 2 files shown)


system/src/Grav/Common/Data/Blueprint.php
```diff
@@ -314,91 +314,115 @@ protected function extraArray(array $data, array $rules, $prefix)
...
-                // Initialize predefined validation rule.
-                if (isset($field['validate']['rule']) && $field['type'] !== 'ignore') {
-                    $field['validate'] += $this->getRule($field['validate']['rule']);
-                }
-            }
-        }
-    }
+				$this->addProperty($prefix . $key);
+
+				$this->parseFormField($field, $prefix, $key, true);
```
w00fz: Should be

```php
$this->parseFormField($field);
```

system/src/Grav/Common/Data/Blueprint.php
```diff
@@ -314,91 +314,115 @@ protected function extraArray(array $data, array $rules, $prefix)
...
+                // This means we need to loop through the list to get to the actual field
+                // The property and rule setting need to be set on the list field
+                // and NOT on the children, which is why we need to duplicate some of
+                // the code below :(
+
+				$this->rules[$prefix . $key] = &$field;
+				$this->addProperty($prefix . $key);
+
+				foreach($field['fields'] as $subName => &$subField) {
+					$this->parseFormField($subField, $prefix, $key, false);
```
w00fz: Should be

```php
$this->parseFormField($subField);
```

w00fz:
Thanks!

PR #480, opened 2015-11-26
Include set_time_limit in a try/catch

flaviocopes:
Refs http://getgrav.org/forum#!/getgrav/general:set-time-limit-problems


vladdu:
This works for me!

PR #481, opened 2015-11-26
FE: Add ability to set-up custom repositories

Sommerregen:
This PR adds the ability to setup custom repositories in a `user/config/repositories.yaml` file with the structure

```yaml
plugins:
  - http://..../plugins.json

themes:
  - http://..../themes.json
```

For security reasons Grav repositories are loaded always at last, thus no "shadowing" of packages is possible.

Although at the moment there is definitely no use for it, this PR is for the future and hence not needed to be documented yet. The reason is that this PR enables one to provide their own repositories, which is useful to integrate private plugin repositories transparently in the Grav update process.

P.S.: Of course I already have a plugin for that ;-).


rhukster:
As discussed, we're going to table this for now.  We are going to revisit it when we start to build out the marketplace functionality.

PR #489, opened 2015-11-30
Adding language string for MISSING_REQUIRED_FIELD

hugoaf:
Adding language string for MISSING_REQUIRED_FIELD in checkRequired
Adding translation for new string in en.yaml
Adding es.yaml


hugoaf:
hey, just send this PR, but dont you think this language strings should be located in `PLUGIN_FORM` instead of `system\languages\`  `FORM`. 

If you agree, is convenient to move also the other two strings to the plugin language file:
```
FORM:
    VALIDATION_FAIL: <b>Validation failed:</b>
    INVALID_INPUT: Invalid input in
```

PR #508, opened 2015-12-07
Add support to manually initialize theme

mahagr:
Gantry admin needs to initialize selected theme in order to set the proper streams to load/save data.

This pull request adds a new function `Themes::initTheme()` allowing plug-ins to initialize theme manually. It also moves `onThemeInitialized` event into proper location to prevent theme being unset on the event call.


rhukster:
Looks good to me, merge when ready.

PR #511, opened 2015-12-07
Make $container available in setup.php

flaviocopes:
@mahagr made this PR to fix a problem introduced in this commit: https://github.com/getgrav/grav/commit/dd2ddfeb409baafc0bf7fed2a5c1f74b2bf8cae2 

The multisite subfolder example http://learn.getgrav.org/advanced/multisite-setup#quickstart-for-beginners was failing because $container was no longer available. Made a PR so you can review and tell me if this is an OK change or you want to revisit it.

### DIFF
--- a/system/src/Grav/Common/Config/Setup.php
+++ b/system/src/Grav/Common/Config/Setup.php
@@ -113,8 +113,13 @@ class Setup extends Data
         ],
     ];
 
-    public function __construct($environment = 'localhost')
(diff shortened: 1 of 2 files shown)


rhukster:
@mahagr can you confirm this won't effect anything in your changes?

Sommerregen:
:+1:

mahagr:
Should be all good.

PR #515, opened 2015-12-09
Escape graph root in page medium to work with special characters.

lennerd:
The current implementation of the `url` function for page medium does not work when the grav root path contains regular expression characters (e.g. `+`).

This PR fixes this by escaping the grav root with `preg_quote`.

### DIFF
--- a/system/src/Grav/Common/Page/Medium/ImageMedium.php
+++ b/system/src/Grav/Common/Page/Medium/ImageMedium.php
@@ -135,7 +135,7 @@ class ImageMedium extends Medium
      */
     public function url($reset = true)
     {
-        $output = preg_replace('|^' . GRAV_ROOT . '|', '', $this->saveImage());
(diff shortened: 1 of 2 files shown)


flaviocopes:
Can you provide an example of a domain / subpath that didn't work before, and works after the fix?

lennerd:
I think the easiest way to have a comprehensive example is to look at the characters, who gets escaped by `preg_quote`: `. \ + * ? [ ^ ] $ ( ) { } = ! < > | : -`

As we dealing here with the physical path on the harddrive of the local machine or server, depending on the file system used, most of the time every of this characters (except the backslash `\`) can be used:

`/path/to/project.folder/`
`/path/to/project/folder/grav+partner/`
… and more.

More information: https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

PR #527, opened 2015-12-14
Make sure Grav can work with the 2.0.0-alpha1 Whoops update, and with the current 1.1 too

flaviocopes:
Whoops 2 made Run a final class, so we cannot extend it any more. Moved whoops configuration to the `Error` class directly, instead of configuring it in `ErrorServiceProvider` and then reconfiguring it in `Error`. 

With those changes Grav runs fine on the currently included 1.1.10 Whoops, and with the upcoming 2.0.0 version.

Unfortunately Whoops 2.0.0-alpha1, current latest, is supposed to work on PHP 7 but in reality it does not yet run correctly, so we need to wait prior to include it in Grav (sent PR https://github.com/filp/whoops/pull/352 to fix a bug that prevents it to work on PHP 7)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_install:
   - export TZ=Pacific/Honolulu
   - composer self-update
   - go get github.com/aktau/github-release
-  - echo "Travis Date: `date`"
+  - echo "Travis Date - `date`"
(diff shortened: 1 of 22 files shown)


flaviocopes:
Ok the PR on Whoops was merged, just needs a release.

rhukster:
As this relies on PHP 5.5 we're going to hold off on this until next release.

PR #530, opened 2015-12-17
Prevent error if no collections present

mufac:
If there are no collections you can get a "Invalid argument supplied for foreach()" without an empty default.


flaviocopes:
Thanks!

Issue #541, opened 2015-12-20
Tag with special characters currently not supported


acki-m:
When I have a tag called: "c++" it will result in an empty collection list, although I have defined items with this tag.
Please add support for this.

flaviocopes:
Tricky thing here is the usage of '+' in the URL. 
We use urldecode() in processing the URL parameters, which interprets `+` as a space. Going to propose a fix.

PR #543, opened 2015-12-20
Use rawurldecode and rawurlencode instead of urldecode/urlencode

flaviocopes:
Proposed fix for #541.

After reading http://stackoverflow.com/a/6998242 rawurldecode (and
rawurlencode) is better over urldecode/urlencode: "rawurlencode is the
way to go most of the time. It deals with the modern scheme for URI
components, where as urlencode does things the old school way, where +
meant "space."

### DIFF
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -484,7 +484,7 @@ class Grav extends Container
         if ($page) {
             $media = $page->media()->all();
 
-            $parsed_url = parse_url(urldecode($uri->basename()));
+            $parsed_url = parse_url(rawurldecode($uri->basename()));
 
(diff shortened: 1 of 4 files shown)


rhukster:
How does this affect params and URLs with spaces, like searching for "something with spaces" or whatever?  Does it change space to `%20` rather than `+`? If so should be ok.

rhukster:
Also what about **encoding** ? Don't we have to also convert `encode` into `rawurlencode` to ensure that they encoding matches up?

flaviocopes:
Yes search was already working with %20

I replaced all Grav calls of `urldecode` to `rawurldecode` and all calls to `urlencode` to `rawurlencode`. Will do more testing.

Will test a bit more, but from my testing I didn't find any issue.

PR #549, opened 2015-12-21
Fix orientation for images added via twig

### LINKED ISSUE #544: Image orientation ?
Hello,
I have a picture taken with an iPad and directly uploaded to a page in Grav. It seems that Grav does not take into account the picture orientation. I have now renamed and resize the orientation of the picture, the picture is displaying correctly in my article but the thumbnail generated by Grav and used by pinpress theme on the front page is 180° rotated, any idea what is wrong? (I have cleared all the cache from Grav admin)

edit: my [example](http://geobib.fr/brickbrack/) with the last post (Christmas theme ;)) displaying wrong .

flaviocopes:
In-progress PR to fix #544 

WIP, do not merge yet

### DIFF
--- a/system/src/Grav/Common/Markdown/ParsedownGravTrait.php
+++ b/system/src/Grav/Common/Markdown/ParsedownGravTrait.php
@@ -267,7 +267,8 @@ trait ParsedownGravTrait
         // if this is a link
         if (isset($excerpt['element']['attributes']['href'])) {
             $url = parse_url(htmlspecialchars_decode($excerpt['element']['attributes']['href']));
-
+            $actions = [];
+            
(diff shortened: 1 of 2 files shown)


flaviocopes:
Did some work for auto fixing exif-rotated images, but only works for twig, not when images are added through markdown (where it should mainly work) and even in Theme Twig files, only works when `pages.process.twig` is true.

rhukster:
Try just doing a `?cache` on markdown images.  If there are no operations, it doesn't go through gregwar library by default, but just adding the **cache** operation should force it to.  Then the `fixOrientiation` call should take effect.

PR #555, opened 2015-12-22
Add ability to adjust images orientation using EXIF data

flaviocopes:
Proposed fix to auto-adjust images orientation using EXIF data. Requires a gregwar/image library upgrade.


flaviocopes:
Need to wait for https://github.com/Gregwar/Image/pull/109 to be merged

flaviocopes:
Upstream PR merged, we now need to wait for a release unless we want to use the latest develop of that library. Last release was 1 year ago, need to check if a release is planned in the near term or not.

w00fz:
Crazy we haven't merged this one yet.
I propose we just fix the dependency to the latest commit which is all about `fixOrientation` anyway.

When you have a moment, @flaviocopes, could you sync and update this PR to:

```json
"gregwar/image": "master#72568cfbeb77515278f2ccb386fc344e874b7ae8"
```

Then we just go ahead and merge it in, I don't foresee any release anytime soon, at least we can move on with this issue. Whenever a release happens we can just revert to the versioning reference.

// commit ref. [72568cf](https://github.com/Gregwar/Image/commit/72568cfbeb77515278f2ccb386fc344e874b7ae8)

flaviocopes:
Ok slightly modified the approach. Uses a system configuration option which defaults to false, to avoid changing how images currently work.

Once that setting `system.images.auto_fix_orientation` is enabled, any image will be processed and cached even if they don't have any processing options, and orientation will be automatically fixed on them.

Going to merge soon if you don't have any objection.

flaviocopes:
(have corresponding language translation ready for Admin)

PR #570, opened 2015-12-28
Update index.php

MATsxm:
Following the new requirements, changing to PHP 5.5.9 or higher.
Thanks


flaviocopes:
Right, we have other places too with this hardcoded requirement. 
Going to use a common shared constant we already set, closing this for now, good reminder anyway!

PR #577, opened 2015-12-31
Update Caddyfile

mholt:
fastcgi was using php preset defaults, anyway


rhukster:
I'm trusting yo on this :)  Don't have php-fpm installed locally at the mo to test.

PR #582, opened 2016-01-02
Add a onPageContentFromCache event

aptly-io:
This event notifies plugins that content came from the cache.
It allows plugins to inspect this cached content and
decide if specific plugin related assets need to be added.

Now plugins typically use the onTwigSiteVariables callback to add their assets to a HTML page;
irrespectively whether the page makes actually use to that plugin.
That is because there's no other means for a plugin to check on cached pages
(onPageContentRaw nor onPageContentProcessed are broadcast for cached content).

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -520,6 +520,9 @@ class Page
                 if ($cache_enable) {
                     $this->cachePageContent();
                 }
+            } else {
+                // Inform plugins that content was cached
+                self::getGrav()->fireEvent('onPageContentFromCache', new Event(['page' => $this]));                
             }
 
             // Handle summary divider


rhukster:
I'm not sure the real advantage of this new event rather than just using `onPageInitialized()` event?  That event happens after the page has been retrieved from cache.

I understand this event is used **only** when page is retrieved from cache, but i'm just hesitant to add another event that is going to happen every page load that serves a very limited purpose.

Trying to keep Grav as fast as possible you know! :)

rhukster:
Please provide me a solid use case where this new event is the only way to solve the problem.  That way I can really determine if it's required or just another approach.

aptly-io:
Hello Andy, thanks for considering the extra event pull request.
Sure I'm considering other workarounds and suggestions.

##### Here's my use case
Since grav was tagged v1.x, I wanted to check-up the googlemaps plugin as well.
Such plugin might only be used on a web site's specific _contact type_ of page.

The plugin uses the `onTwigSiteVariables` event, like numerous other plugins, to add its assets.
Now, because of this mechanism, _each_ page gets these assets added, even if _not_ using googlemaps.
For my plugin, that means pulling JS from google, some common JS wrapper for the plugin, some per-map dynamic inline JS (to get around the modular page issue) and css.

I sought a way to only send the assets if the page's markdown contains `[GOOGLEMAP:<tagid>]`.
But once the content is cached, my plugin no longer gets in control (`onPageContentProcessed` is no longer sent). 

That's how I came up with the alternative event. Now the plugin looks in the cached html to figure out if assets are necessary.

Here's how my googlemaps plugin would be using that new event for now:
https://github.com/aptly-io/grav-plugin-googlemaps/blob/grav_v1_checkup/googlemaps.php
- [defines the call-back](https://github.com/aptly-io/grav-plugin-googlemaps/blob/grav_v1_checkup/googlemaps.php#L66)
- [call-back implementation](https://github.com/aptly-io/grav-plugin-googlemaps/blob/grav_v1_checkup/googlemaps.php#L95-L104)
- [the real work (a grep for plugin specific html attributes)](https://github.com/aptly-io/grav-plugin-googlemaps/blob/grav_v1_checkup/googlemaps.php#L233-L255)

I'm unsure how I can leverage `onPageInitialized` for this case?

I agree, this event triggers always, for each cached page, for each plugin that is subscribed. It could slow down Grav. For now I don't have other inspiration on how to limit the amount of data assets send.

rhukster:
We do what you are requiring quite frequently, ie adding assets only on pages that are needed.  There are several ways we do this in other plugins today:

1) Have a page header that enables the plugin for a page (and can be reliably used to add JS/CSS). This is usually used in conjunction with other config options that are merged into the main config.

2) Like the assets plugin (i recently made fixes), you can use your onContentRaw or any other even to process the google map tag, and then cache/save some state to cache for this particular page route.  Then in a later event, look for the current page in cache and if found, add the assets.

There are other ways, and #2 is really overkill.  #1 is best way although it does require you to 'manually enable' the plugin in the page header, but has the benefit of allowing you to not even bother processing the page content if not enabled.

aptly-io:
Hello Andy,

Thank you for the reply. 

Indeed option 1 would be perfect. Can you point me to one plugin that actual implements option 1 please? The most used plugins I checked have _only a global site-wide_ enable header.

A plugin can use only these events  to check a _per-page_ enabled flag: `onPageContentRaw` and `onPageContentProcessed`. But these events are only triggered _as long as these are *not* cached yet_. Once a page is cached, a plugin does not receive the event, cannot run its code to check if its enabled for that page, cannot add any per page assets.

I will investigate your option 2) this weekend.

flaviocopes:
@aptly-io for example https://github.com/getgrav/grav-plugin-highlight/blob/develop/highlight.php#L34-L36

rhukster:
Closing due to inactivity

PR #583, opened 2016-01-03
Call finalizeObject on Config from loadCompiledFile() when cache is already built

flaviocopes:
finalizeObject, which assigns the checksum to the Config object, was
never called after the cache was built.

So the checksum on Config was always empty.


flaviocopes:
@mahagr when you're back can you have a look at this? The PR fixes the issue, but I'm not sure if this is the root cause of the problem.

mahagr:
Missed that one. Your fix looks correct to me.

flaviocopes:
:+1:

PR #585, opened 2016-01-04
Allow grav to be installed as a composer project

jeromegamez:
As a developer, instead of ([as described in the docs](http://learn.getgrav.org/basics/installation#option-2-install-from-github))

```shell
$ git clone https://github.com/getgrav/grav.git
$ cd grav
$ composer install --no-dev -o
$ bin/grav install
```

I would like to be able to execute

```shell
$ composer create-project getgrav/grav my_new_grav_project
```
to bootstrap a new Grav project.

This PR includes the following changes:
- Change package type from `library` to `project`
  - This ~~is needed to be able to execute the `composer create-project` command, and also~~[<sup>1</sup>](#issuecomment-168844993) fits better, in my opinion
- Add a `post-create-project` command which executes the `bin/grav install` command after project creation.

This has the following benefits:
- It is possible to bootstrap a new project with one single command instead of four
- Other than with a `git clone`, no `.git` directory is created in the project root
- It is possible to specify a specific grav version (e.g. `^1.0` or `dev-develop`)

If you consider merging this PR, the project also must be submitted to http://packagist.org so that `composer create-project` can find it.

:octocat:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "getgrav/grav",
-    "type": "library",
+    "type": "project",
     "description": "Modern, Crazy Fast, Ridiculously Easy and Amazingly Powerful Flat-File CMS",
     "keywords": ["cms","flat-file cms","flat cms","flatfile cms","php"],
     "homepage": "http://getgrav.org",
@@ -40,5 +40,8 @@
     },
     "archive": {
         "exclude": ["VERSION"]
+    },
+    "scripts": {
+        "post-create-project-cmd": "bin/grav install"
     }
 }


rhukster:
Very interesting! I didn't even know about the `project` type in composer. I'm going to check it out now.

jeromegamez:
Thank you for your quick reply!

In fact, a `composer create-project` isn't really different from `git clone` + `composer install`. As composer is not able to find the project yet, the steps to test the procedure would be to execute:

```shell
$ git clone git@github.com:jeromegamez/grav.git grav_test
$ cd grav_test
$ composer run-script post-create-project-cmd
```

jeromegamez:
As a reference, here are the Composer docs for projects:
- Package Type: https://getcomposer.org/doc/04-schema.md#type
- The `create-project` CLI command: https://getcomposer.org/doc/03-cli.md#create-project

As I am reading and testing it again, the `project` type is actually not needed to execute the `create-project` command - it still fits Grav more :)

rhukster:
Awesome, going to add to packagist now.

rhukster:
Can install from packagist now, but the script won't be enabled until the next release as it needs to be in the release version it seems.

jeromegamez:
You can specify a version when creating a new composer project:

```shell
$ composer create-project getgrav/grav grav dev-develop
```

jeromegamez:
I also proposed an addition to the grav-learn repo, although I have to admit that I markedowned it blindly :).

rhukster:
Great, i'll update that repo after this release. Thanks!

jeromegamez:
This all went so fast, thank **you**! :rocket:

PR #588, opened 2016-01-05
Introduce testing

flaviocopes:
Ongoing PR. 
Introduces testing using Codeception. 

First implementation and setup of a single Unit test.

Base for further work on testing. Needs `composer update` to install the new dependencies.

Run tests with `./vendor/bin/codecept run`

Add a new Unit test with `./vendor/bin/codecept generate:test unit Example`


tests/_support/_generated/UnitTesterActions.php
```diff
@@ -0,0 +1,348 @@
...
+     *
+     * Fails the test with message.
+     *
+     * @param $message
+     * @see \Codeception\Module\Asserts::fail()
+     */
+    public function fail($message) {
+        return $this->getScenario()->runStep(new \Codeception\Step\Action('fail', func_get_args()));
+    }
+}
```
javigomez: Glad to see you using Codeception, great choice! ^_^

I suggest to ignore the _generated folder like we do in weblinks: https://github.com/joomla-extensions/weblinks/blob/master/.gitignore#L53
Is anyway locally generated when codecept do the `build` process, so not needed to be tracked in github.

Just my 2 cents
^_^

flaviocopes: Good suggestion, added :)

PR #602, opened 2016-01-09
Fix parents for home page second level childs when hide home in url is enabled

flaviocopes:
TESTING IN PROGRESS

Return the parent routes in the `route` options.

Added a Pages::parentsRawRoutes() method, uses the same internal method
as Pages::parents().

Refactored Pages::getList() to accept an optional $rawRoutes bool and
if set, return the raw routes


flaviocopes:
Sorry didn't mean to restore

Issue #600, opened 2016-01-09
feature request: page update time for modular pages


simonhaenisch:
The date for the last update of a page that is modular is the update time of the page markdown file, but I think it should be the update time of the module that was last modified, because thats the real update time of that page. The sitemap plugin uses this date for the sitemap and thus it seems a homepage wasn't updated even though the modules changed.

flaviocopes:
Seems a good improvement to me

flaviocopes:
Can you test the linked PR?

rhukster:
I looked at this PR and while this should resolve the issue in most cases, tecnically a modular page can be defined by a variety of options that are set in the page header, not just children.  So to to do this properly it would have to use the modular page's collection, not just the children.

rhukster:
@simonhaenisch Please test the latest updates to the PR so I can close this.  Cheers.

PR #605, opened 2016-01-10
The date for the last update of a page that is modular is the update time of the module that was last modified

flaviocopes:
Refs #600


rhukster:
I think this last update should fix it.

flaviocopes:
Works for me @rhukster, good to merge

PR #609, opened 2016-01-17
Remove ircmaxell/password-compat as it's intended to provide forward compatibility for PHP 5.4

flaviocopes:
However, it will still be installed in `vendor` as required by
`rockettheme/toolbox`


mahagr:
I removed dependency from toolbox; it wasn't needed in there.
https://github.com/rockettheme/toolbox/commit/578fabab2387e907dcd3f7b3c2bf9c37fe8a23bc

flaviocopes:
:+1:

PR #613, opened 2016-01-18
Add a twig filter to cast an object to array

flaviocopes:
Usage: `anObject|cast_to_array`


flaviocopes:
Already available as a Twig function `array(var)`

PR #616, opened 2016-01-19
Update Whoops to latest

Vivalldi:
Should have line highlights.


flaviocopes:
Maybe we can set to `~2.0` so we get up to the next significant release?

Vivalldi:
Thanks! I learn something new each day :smile:

PR #623, opened 2016-01-21
Fix loading js and css from collections

flaviocopes:
Calling add() lose information on

- CSS: $group
- JS: $loading, $group


flaviocopes:
Refs https://getgrav.org/forum#!/getgrav/general:loading-jquery-async

flaviocopes:
WIP - Recreated an issue in calling with an array of values `{% do assets.addJs('jquery', {'loading':'async'}) %}` introduced in https://github.com/getgrav/grav/commit/d5bd99b363ec955a90bf50f9f8856a819bec31bf

Checks for func_num_args() == 2 but func_num_args() is now 3 as argument # 3 defaults to `true`.

Still looking to make it 100% fine

flaviocopes:
Finally got to the root of the problem, supported by a bunch of tests in the testing branch https://github.com/getgrav/grav/commit/27af9432e0fa9a7abb2b6a2e3f3a6f6d155a37e9

rhukster:
So to confirm, adding an array to addCss, will force all items to be added 'as' css, and same for JS? This fixed the issue reported?

flaviocopes:
Actually it's 2 fixes, they both solve the issue.

First we check `if (func_num_args() > 1) {` instead of `if (func_num_args() == 2) {` otherwise the array is never merged and `loading` and `group` are never applied when using the syntax `{% do assets.addJs('jquery', {'priority':101, 'loading':'async'}) %}`

Then we force to use `addCss()` or `addJs()` because otherwise the information on `$loading` and `$group` is lost as `add()` does not accept them as parameters

PR #671, opened 2016-02-12
fix lighttpd config

rkusa:
Consider the following path: `/grav_path/user/pages/01.home/_about/profilepic.jpg`
The pattern `^/grav_path/(\.(.*))|(\.(.*)/)` will match `.home/_about/` and will therefore forbid access to this image.

If I understand it right, this pattern should prevent access to dot files. In this case, it should be fine to simply remove the right *or branch*.

Another note: the `lighttpd.conf` also contains the following rule:
```
$HTTP["url"] =~ "^/grav_path/(.git|cache|bin|logs|backup|tests)/(.*)" {
    url.access-deny = ("")
}
```

I think this rule is redundant, because the other rule already catches all dotfiles?


rkusa:
I think both rules could also be replaced by this simple one:

```
$HTTP["url"] =~ "/\." {
    url.access-deny = ("")
}
```

Maybe even better would be to return a 404 instead of a 403? Because with a 403 you at least know that there is such a file.

Let me know what you think and I can update the PR accordingly.

rhukster:
Cheers.

PR #695, opened 2016-02-22
Adding Tests - On going

Vivalldi:
Close as you need tests. If closed I'll create another pull

~Cheers :sparkles:


rhukster:
Nice! The more tests the merrier!  Thanks!

PR #698, opened 2016-02-22
Allow authentication for proxy

flaviocopes:


rhukster:
@flaviocopes did you ever receive any update son this?  I think it should fix those authentication situations.

flaviocopes:
No haven't got any reply / confirmation that the OP problem was solved

rhukster:
Merging, we can fix if it doesn't solve the problem 100%. We'll never know unless someone tries it :)

PR #721, opened 2016-03-08
Fixes a typo in months of the year

mikewink:


rhukster:
Hey Mike, this actually needs to be fixed in Crowdin or it will get lost in the next update: https://crowdin.com/translate/grav-core/5/en-de

PR #725, opened 2016-03-11
Developer CLI Tools

Vivalldi:
The components (themes and plugins) need some loving care and some updating. basically just need to get more options for them to use. If another plugin template is added update the `newplugin` command to prompt for plugin type.


Vivalldi:
@rhukster this can be merged. Once I have a few more things I'll create another pull request but I feel this would be good for v1.1

rhukster:
Ok sounds good, i'll take a look at this today and hopefully get it merged in.  Another great feature for Grav 1.1. Thanks!

flaviocopes:
The bootstrap start-up theme is including the bootstrap js / css, should we just tell somehow the user to install the Bootstrapper plugin instead? Or we need to update those libraries.

rhukster:
I'm actually going to rewrite this 'bootstrap' theme today so that it's something else entirely.. non-bootstrap .. just a very simple basic theme.

PR #731, opened 2016-03-16
Translate dates automatically in Twig filter

flaviocopes:
The Twig `date` filter does not translate dates automatically for us.
This PR lets `|date()` translate the date based on the locale.


rhukster:
Looks good flavio, feel free to merge into develop.

PR #743, opened 2016-03-23
Update nginx.conf

mrhein:
if is evil see https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

### DIFF
--- a/webserver-configs/nginx.conf
+++ b/webserver-configs/nginx.conf
@@ -10,10 +10,9 @@ server {
     ## Begin - Index
     # for subfolders, simply adjust:
     # `location /subfolder {`
-    # and the rewrite to use `/subfolder/index.php`
+    # and the try_file to use `/subfolder/index.php` and `/subfolder/index.html`
     location / {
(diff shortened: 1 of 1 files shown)


rhukster:
Looks good. I will test when I get home from vacation this weekend. Thanks!

rondlite:
the .js javascript header found in example config should be editted to. Otherwise all .json uri's are parsed as files and you can not clear cache of preview your work in the admin panel.

rhukster:
@rondlite can you provide a code snipped with .js fix?

rondlite:
@rhukster soz should have posted with previous post
`location ~ \.js$ {
    add_header  Content-Type    application/x-javascript;
}
`
just add EOL after the .js

also, not sure if is needed but my nginx.conf I made like this (from the caddy example):

`location / {
        try_files $uri $uri/  /index.php?_url=$uri;
          }
`

rhukster:
i've added the updated to try_files manually.

PR #745, opened 2016-03-24
Refactored the Grav classes load and process methods to follow clean coding standards

toovy:
When we evaluated Grav the first class I've looked into was Grav/Common/Grav. I loved the container and the central process chain, but the code seemed to be "copy & paste". Nearly the same code was repeated over and over again. The open-closed-principle was not applied. I've refactored the process and load method using two arrays as basis to load the needed services dynamically into the container and to use them in the process method in a more generic way. New services and processes can now be added without modifying existing mode (only the arrays would need to be modified).


system/src/Grav/Common/Grav.php
```diff
@@ -215,83 +115,13 @@ public function process()
...
-        $this->output = $this['output'];
-        $this->fireEvent('onOutputGenerated');
-        $debugger->stopTimer('render');
+        // process all processors (e.g. config, initialize, assets, ..., render)
+        foreach ($this->processors as $processor):
+          $processor = $this[$processor];
+          $this->measureTime($processor->id, $processor->title, function($debugger) use ($processor) {
+            $processor->process($debugger);
+          });
+        endforeach;
```
mahagr: We aren't in a template file here, please use `{ ... }` instead. :)

toovy: As the other code uses the classical notation { ... } I have changed it. But I've not see any guideline that say the alternative syntax should only be used in templates. Happy to learn something. :)

system/src/Grav/Common/Grav.php
```diff
@@ -520,6 +350,96 @@ public function shutdown()
...
+
+        $container['debugger'] = new Debugger();
+        $debugger = $container['debugger'];
+
+        // closure that measures time by wrapping a function into startTimer and stopTimer
+        // The debugger can be passed to the closure. Should be more performant
+        // then to get it from the container all time.
+        $container->measureTime = function($timerId, $timerTitle, $callback) use ($debugger) {
+          $debugger->startTimer($timerId, $timerTitle);
+          $callback($debugger);
```
mahagr: I do not like having $debugger as a parameter. The parameter seems to have no purpose, which is why I ended up using a lot of time trying to understand reasoning behind it. I see it being used in 2 processors out of 13, but then again, its already available through the container. If its just there because of optimization, there are better ways to optimize code.

toovy: You've got a point. I'll change that.

system/src/Grav/Common/Grav.php
```diff
@@ -520,6 +350,96 @@ public function shutdown()
...
+     * Register all services
+     * Services are defined in the diMap. They can either only the class
+     * of a Service Provider or a pair of serviceKey => serviceClass that
+     * gets directly mapped into the container.
+     *
+     * @param  Container $container
+     *
+     * @return void
+     */
+    protected static function registerServices($container) {
```
mahagr: I'm not sure why these functions are static; they are only used from `Grav::load()`, which already has container class available.

system/src/Grav/Common/Grav.php
```diff
@@ -520,6 +350,96 @@ public function shutdown()
...
+     * Register all services
+     * Services are defined in the diMap. They can either only the class
+     * of a Service Provider or a pair of serviceKey => serviceClass that
+     * gets directly mapped into the container.
+     *
+     * @param  Container $container
+     *
+     * @return void
+     */
+    protected function registerServices($container) {
```
mahagr: $container = $this :)

mahagr:
I cannot find anything else to be fixed. :)

I like the change, even if right now there are not real benefits of having the array except the code refactor to make it more abstract. I'm thinking if we should allow plugins to use the new protected methods you added.. Hmm.. Also at some point we might want to allow `setup.php` to mess up with the DI container (to make real use from having an array), but that is not in the scope of this change.

toovy:
You are right, the changes will make the system more open in future, there is no immediate benefit that e.g. a feature adds. I'm off the opinion that the core also needs refactorings like this one, else Grav might end up with a codebase like other successful CMS that are nearly unmaintainable *cogh* Wordpress *cogh*. ;)

mahagr:
There are multiple classes in Grav that I would refactor right away if I had the time to do it. One of the most painful ones are Page and Pages classes, which have experienced the most changes since we first implemented them. I'm just afraid the changes I would make would also break B/C, so they just need to wait a bit longer.

Changes like this one are really welcome. Still I'd like to wait to have extra pairs of eyes checking out this code and the performance implications it may have (I doubt there are any, but its something we also need to consider).

@rhukster I know you're at your vacation, but if you have some extra time, this will be one of the most interesting things you could look into. :)

toovy:
@mahagr sure, that all makes sense. As we might use Grav for more clients we also might investigate even more.

toovy:
Oh I see what you mean, 2526 locs...

rhukster:
I really like the ideas behind this refactor.  Definitely does improve readability. I can see that down the line having more public API methods to allow for plugins to interact directly with the processors could be very useful.

I only have a couple of days left on my vaca and i'll be back home.  I'll look at this more thoroughly this weekend.  Thanks!

rhukster:
I know we've made a few changes that is causing some conflicts now.  Would you mind updating this PR with the latest changes in the develop branch and i'll do some testing?  

Thanks!!!

toovy:
@rhukster I'll try to look into the conflicts next week

toovy:
@rhukster merged, tests are ok (except the 5 that did go wrong before) and the demo system runs.

rhukster:
Ok, i've tested on my local and everything seems to be fine.  Performance differences are negligible which was my only real concern.  Thanks for your hard work on this, i think it's going to make Grav very flexible going forward.

toovy:
Happy to help :)

flaviocopes:
P.S. tests should be good now. 
They were failing when run with a user/ folder that had at least one plugin (inside a site folder, for example), but working in Travis / when run inside the clean Grav git repository.

PR #753, opened 2016-03-27
Added the ability to inline css and js code

mrhein:
```
{% block javascripts %}
            {% do assets.addJs('theme://js/lazysizes.js') %}
            {% do assets.addJs('theme://js/loadpreload.js') %}
        {% endblock %}
{{ assets.js('head', true) }}
```
results in a javascript inline tag same with css and style tag


rhukster:
You are changing the method signature, this will break backwards compatibility. Any new params should be added in the attributes array only.

PR #754, opened 2016-03-27
Added the ability to inline css and js code

mrhein:
`{{ assets.css('head', {'inline':true}) }}` now results in inline tag same with css style tag

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -77,6 +77,7 @@ class Assets
     protected $timestamp = '';
     protected $assets_dir;
     protected $assets_url;
+	protected $attributes;
 
     // Default values for pipeline settings
     protected $css_minify = true;
@@ -519,6 +520,7 @@ class Assets
         $this->css = array_reverse($this->css);
         $this->inline_css = array_reverse($this->inline_css);
 
+		$this->attributes = $attributes;
         $attributes = $this->attributes(array_merge(['type' => 'text/css', 'rel' => 'stylesheet'], $attributes));
 
         $output = '';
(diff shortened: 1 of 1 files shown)


system/src/Grav/Common/Assets.php
```diff
@@ -77,6 +77,7 @@ class Assets
     protected $timestamp = '';
     protected $assets_dir;
     protected $assets_url;
+	protected $attributes;
```
rhukster: From just reading over the code i'm not really seeing where this protected instance variable is used. Below you set the attributes to this value, and then check it for `inline` but why not just do it with the local `$attributes` var?  Is there some persistent need?

rhukster:
ok, i guess first please let me know what issue this solves?  There is already a separate array for inline JS and CSS and you can add code via `addInlineCss()` and `addInlineJs()` methods.

mrhein:
You can't get every css or js that was added maybe from a plugin as inline element ... and further i can add anther method that lazyload js and css by parsing out the core required css and so on but this will let me investigate more time ... but this is the first step.

flaviocopes:
So this will allow to print the content of any CSS/JS file added, in an inline `<script>` /  `<style>` tag, instead of linking the file itself

mrhein:
this will only affect pipelined stuff so if you activate pipline css and/or pipline js and give the option .css/.js({"inline":true}) it will output a <script>/<style> tag with all js or style in one blog this is very usefull if you don't use a creapie framework like bootstrap or jquery and only have a few KB of js and css so you can have a small size of page size without extra request next day's i will bring a extreme small and very fast font loader ... after that i will investigate time to ... read the comment above ...

rhukster:
Seems i've updated in Assets to the point where this doesn't merge now.  Would you mind updating this with latest commits and i'll test it?

rhukster:
@mrhein if you would like this in Grav 1.1, please update your PR with latest commits so it merges.  Cheers!

OliverO2:
Here's another PR addressing the issue: https://github.com/getgrav/grav/pull/1377

flaviocopes:
Closing this PR as was not updated for a year, and unlikely to be merged flawlessly now.

PR #759, opened 2016-04-01
Add params, query and fragment when redirecting to a language or default route

flaviocopes:
Fix an issue in the Grav lang redirect. E.g. I call `/order/id:order-20160320-202557-668641/token:2mgb4D27S3`, if multilang is ON, and the lang is prefixed, Grav loads the page `/en/order/`, losing all the parameters and additional URL segments


rhukster:
One thing we could do to optimize.  Rather than check if `$url == '/'` just `ltrim($url, '/')`

PR #773, opened 2016-04-06
Fix environment configuration for localhost testing

### LINKED ISSUE #772: Environment-Based Language Handling doesn't automatically route users to the correct site-version

flaviocopes:
Fixes #772

### DIFF
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -157,14 +157,26 @@ class Uri
         return $this->name;
     }
 
+    /**
+     * Return a string with the current environment name.
+     *
+     * If the server IP is a local server address (127.0.0.1 or ::1), and/or
+     * the host name is set to 'localhost', returns 'localhost'.
+     *
+     * Otherwise it returns the server name as set by self::buildHostname()
+     *
+     * @return string
+     */
     private function buildEnvironment()
     {
-        // set hostname
-        $address = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : '::1';
+        $address = isset($_SERVER['SERVER_ADDR']) ? $_SERVER['SERVER_ADDR'] : '::1';
 
-        // check for localhost variations
         if ($this->name == 'localhost' || $address == '::1' || $address == '127.0.0.1') {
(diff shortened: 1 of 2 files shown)


system/src/Grav/Common/Uri.php
```diff
@@ -157,14 +157,26 @@ private function buildHost()
...
         if ($this->name == 'localhost' || $address == '::1' || $address == '127.0.0.1') {
-            $env = 'localhost';
+            if ($this->name != 'localhost') {
+                $env = $this->name;
+            } else {
+                $env = 'localhost';
+            }
         } else {
             $env = $this->name;
         }
```
smz: @flaviocopes Hi, Flavio! Sorry, but aren't all of the above conditional statements just equivalent to "$env = $this->name;" ?

flaviocopes: Good point, I tricked myself. Actually to allow virtual hosts on localhost the whole method can be boiled down to 

```
        return $this->name;
```

smz: :wink: ... also (nitpicking...), in the Unit Tests, wouldn't it be better to use example.(com | org | net) domain names instead of _foobar _and/or _google_?

flaviocopes: Yes it's just some random URL testing.
Feel free to improve the tests if you want (once this is merged, so there's no conflicts)

smz: OK, will do once merged! How are the _house rules_ here? PR against "master" or "develop"?

flaviocopes: We follow/use git-flow.

`master` is the latest release (stable) branch. 
`develop` is the working branch. PRs are automatically sent to the `develop` branch, so no worries.

This is for Grav core and all plugins/themes.

You can find some more info on https://github.com/getgrav/grav/blob/develop/CONTRIBUTING.md

smz: Got it! Thanks and... ciao! Sergio

rhukster:
Closing as i resolved this with a compromise solution that still treats 127.0.0.1 and ::1 as localhost, but uses hostname for environment otherwise

PR #775, opened 2016-04-06
This allows a frontmatter.yaml file to be included in a page folder

rhukster:
any 'frontmatter.yaml` file will be merged into the page header. 

### Caveats with this approach:
With this approach, in the admin this will be merged in transparently and shows up as if it were in the page frontmatter itself, and then when you save, it is in fact saved back into the page.  

Doing anything else differently is going to be much more complex because we'll have to handle two ways of generating header data (one for admin ignoring this file, and one for frontend).  However it's not even that easy because this data is potentially cached, meaning we'll have to maintain two versions in cache, leading to possible inconsistencies.

This is a pretty advanced feature and would not be something that could be set via Admin so am not sure if this is a problem or not.

Also it is fine if you use **Expert** mode because that reads and saves the frontmatter in the page header directly.

### DIFF
--- a/system/src/Grav/Common/Cache.php
+++ b/system/src/Grav/Common/Cache.php
@@ -264,6 +264,15 @@ class Cache extends Getters
     }
 
     /**
+     * Setter method to set key (Advanced)
+     */
+    public function setKey($key)
+    {
+        $this->key = $key;
+        $this->driver->setNamespace($this->key);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


rhukster:
This PR is regarding  Issue #775

rhukster:
Ok i've found a solution to my caveat, i've actually set it so the cache on admin is independent of site.  This should fix some other obscure issues too.

PR #778, opened 2016-04-07
Add ability to link to file on gist

Xobb:
Now when the file is passed to the gist filter only that file will be displayed, not the whole gist.

Example usage:

```{{ gist("000000000000000000000000", "somefile.php") }}```

Backward compatible by the way.


rhukster:
Good little addition.. thanks.

PR #801, opened 2016-04-23
Allow page blueprints in plugins

flaviocopes:
Allow page blueprints in plugins. Also invert order of loading templates, themes load last and have the "final word". Both on templates and in blueprints.

Cache everything if cache is enabled.

Example usage in a plugin:

```php
<?php
class MyPlugin extends Plugin
{
    /* ... */

    /**
     * @return array
     */
    public static function getSubscribedEvents()
    {
        return [
            'onGetPageBlueprints'     => ['onGetPageBlueprints', 0],
            'onGetPageTemplates'      => ['onGetPageTemplates', 0],
        ];
    }

    /* ... */

    /**
     * Add page blueprints
     *
     * @param Event $event
     */
    public function onGetPageBlueprints(Event $event)
    {
        /** @var Types $types */
        $types = $event->types;
        $types->scanBlueprints('plugins://myplugin/blueprints/pages/');
    }

    /**
     * Add page template types.
     *
     * @param Event $event
     */
    public function onGetPageTemplates(Event $event)
    {
        /** @var Types $types */
        $types = $event->types;
        $types->scanTemplates('plugins://myplugin/templates');
    }

    /* ... */
}
```

### DIFF
--- a/system/src/Grav/Common/Page/Pages.php
+++ b/system/src/Grav/Common/Page/Pages.php
@@ -486,13 +486,44 @@ class Pages
     public static function getTypes()
     {
         if (!self::$types) {
-            self::$types = new Types();
-            self::$types->scanBlueprints('theme://blueprints/');
-            self::$types->scanTemplates('theme://templates/');
 
-            $event = new Event();
-            $event->types = self::$types;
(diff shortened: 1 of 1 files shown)


rhukster:
Going to merge his @flaviocopes ???

PR #806, opened 2016-04-27
Make sure there's an end of line after each file added to the pipeline

flaviocopes:
Fixes an issue with files ending with a line comment, but no new line.

Also fixes (in a case I found) the already-seen JSMin fatal issue with
"Regex not terminated" error.


nepomuk13:
This fix solve the problem indeed, thanks!

PR #841, opened 2016-05-19
Install plugins and themes in correct multisite folders

Sommerregen:
When having multisite folders with different user folders and one installs a plugin or a theme, Grav always installs it in the "root" folders like `user/plugins` or `user/themes`.

This PR addresses the issue (cf. https://github.com/getgrav/grav-plugin-admin/issues/319) and fixes it with full backward-compatibilty and without any change to the Grav GPM package system.

However in the future I advise to use streams in the package system, which are resolved to the respective folders in the grav installs of the users. This PR is a step to this direction and (I guess) will be tackled in **Grav v2.x**.

### DIFF
--- a/system/src/Grav/Common/GPM/Remote/AbstractPackageCollection.php
+++ b/system/src/Grav/Common/GPM/Remote/AbstractPackageCollection.php
@@ -50,6 +50,11 @@ class AbstractPackageCollection extends BaseCollection
 
         $this->fetch($refresh, $callback);
         foreach (json_decode($this->raw, true) as $slug => $data) {
+            // Temporarily fix for using multisites
+            if (isset($data['install_path'])) {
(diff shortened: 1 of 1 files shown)


rhukster:
Yes we definitely plan on moving the CLI stuff to streams.  Probably before Grav 2.x, Maybe even Grav 1.2.

rhukster:
Thanks btw.

hi-ko:
As mentioned in getgrav/grav-plugin-admin/issues/319 the easiest work around we now use is to create symlinks for plugins, themes in sites. We do the same for pages/pagetrees from other sites we want to "reuse" in other sites.
Unfortunately this does not solve the CLI issue which we need for automated deployments (and e.g. reindex search)

Issue #842, opened 2016-05-20
Improper work of language switcher


rkrejcii:
I have multilingual website under Grav which uses grav-plugin-langswitcher.

Actually I have following problem:

- website (let's call it `mywebsite.com`) has default language set to English, default language is without a prefix

- when user switches to another language, the URL properly uses language prefix (such like `/it`) and shows properly the content from `.it.md` pages

- however when I now open another tab in the same browser and access `mywebsite.com`, English version is displayed again. What I expect instead of it is redirection to the last language version (like `mywebsite.com/it/`) I'm using in another tab. And I expect such behaviour even if I restart the browser, unless I'm not browsing in anonymous mode.

Here is my configuration from `system.yaml`:

```
...
languages:
  supported:
    - en
    - el
    - it
  include_default_lang: false
  translations: true
  translations_fallback: true
  session_store_active: true
  http_accept_language: false
  override_locale: true
home:
  alias: /home
  hide_in_urls: false
...
session:
  enabled: true
  timeout: 1800
  name: myskin
  secure: false
  httponly: true
...
```

Thank you for any hint how to achieve of demanded behaviour which is common in another CMS's (such like WordPress with WPML).

flaviocopes:
Seems there's an issue. The PR https://github.com/getgrav/grav/pull/843 fixes the issue, for **Grav 1.1**

rkrejcii:
@flaviocopes: Thank you for fast reaction, appreciated. Any idea when Grav 1.1 or at least this one fix will be released?

rhukster:
This should be fixed now for next beta release.

PR #843, opened 2016-05-20
Load `session` prior to `uri` as the initialization of Uri does some session handling for the active language

flaviocopes:
Refs https://github.com/getgrav/grav/issues/842


rhukster:
I don't see any issues with this.

PR #845, opened 2016-05-23
Move url and session init prior to twig init

flaviocopes:


flaviocopes:
Closed as could not recreate the original issue

PR #851, opened 2016-05-27
translate(): apply sprintf even if translation key missing in yaml

fsalomon:
Let's say someone wants to use English text as translation keys instead of constants. In that case, the en.yaml would stay empty. I know that's more effort to maintain if texts need to be changed, but it's faster to prototype this way.

The output of {{ "There are %d monkeys in the %s"|t(12, "London Zoo") }} should always be with replacements applied, regardless the contents of the yaml files.

### DIFF
--- a/system/src/Grav/Common/Language/Language.php
+++ b/system/src/Grav/Common/Language/Language.php
@@ -377,6 +377,10 @@ class Language
             }
         }
 
+        if (count($args) >= 1) {
+            return vsprintf($lookup, $args);
+        }
+        
         if ($html_out) {
             return '<span class="untranslated">' . $lookup . '</span>';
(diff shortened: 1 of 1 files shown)


flaviocopes:
Also re #852 I'm not sure we should allow using plain text instead of language placeholder constants. Let's hear other opinions.

rhukster:
i really don't see the point of having a plain english translation.  If you don't want to translate yet, simply put english.  If you are ready to translate, use a key in the format we are using (for consistency). 

That said, no we have not specifically forced a key structure, or format, and I don't think there's much point in doing that.. we are just using a simple syntax to make this process as easy as possible.

We'll work on documenting it better.

PR #852, opened 2016-05-27
bugfix: translation key couldn't contain dot

fsalomon:
Let's say someone wants to use English text as translation keys instead of constants. I know that's more effort to maintain if texts need to be changed, but it's faster to prototype this way.

{{ "This is a sentence."|t }} or {{ "e.g."|t }} wasn't working before due to "." being used as a separator by RocketTheme\Toolbox\ArrayTraits\NestedArrayAccess. This patch fixes it.

### DIFF
--- a/system/src/Grav/Common/Config/Languages.php
+++ b/system/src/Grav/Common/Config/Languages.php
@@ -41,4 +41,9 @@ class Languages extends Data
     {
         $this->items = array_merge_recursive($this->items, $data);
     }
+    
+    public function getTranslation($lang, $key, $default = null) 
+    {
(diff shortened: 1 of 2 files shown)


flaviocopes:
Also re #852 I'm not sure we should allow using plain text instead of language placeholder constants. Let's hear other opinions.

fsalomon:
@flaviocopes imho that's not the question. You have been allowing it all along. If you wanted to prevent it, you'd have to force translation keys to match /^[A-Z\.\-]+$/. It's too late to do that now, though.

rhukster:
Closed with same reason as #851

PR #899, opened 2016-06-14
Force SSL on a page if system.force_ssl is enabled

flaviocopes:


rhukster:
is `system.force_ssl` a valid option yet?

flaviocopes:
No, must be set manually, once the PR is confirmed to work I'll add it to the blueprint and add the language strings

rhukster:
We should add at least an empty config entry in `system.yaml`, also the blueprint should be updated to allow these to be set from admin.

Issue #898, opened 2016-06-14
dateRange without end date problem


Sogl:
Hello!

This issue related to:
https://github.com/getgrav/grav/issues/274
https://github.com/getgrav/grav/pull/275

I'm trying to show some posts from blog to in a home modular page depending to start date and found a bug with `dateRange` filter. Blog has 2 posts with 2013 and 2016 years.

My section `frontmatter.yaml`:
```yaml
events:
    items:
        '@page': '/events'
    dateRange:
        start: 1/1/2012
        end: 1/1/2017
```

Markdown:
```twig
{% set events_collection = page.collection('events') %}

{% for event in events_collection %}
    <p>{{ event.title }}</p>
{% endfor %}
```

With this settings I see both posts, it's fine.

But without end date in `frontmatter.yaml`:
```
events:
    items:
        '@page': '/events'
    dateRange:
        start: 1/1/2012
```

I see no posts at all. 

Env:
* Grav v1.1.0-rc.1

flaviocopes:
Mmm I tried but I can't replicate, it works fine to me.

Sogl:
It replicates also with `field: header.date`.

How to see full request to my `/events`?

flaviocopes:
What's an example of a child page? Do you define the date there?

Sogl:
>What's an example of a child page? Do you define the date there?

```
title: Post with a Background Image
date: 25.10.2013
```

I don't know what happened, but it's working on my home Mac machine after `git pull` 😕 
Cache is always disabled.

Need to check on Windows machine where it happen.

rhukster:
Please check the Grav version specifically, see if they are in sync.

flaviocopes:
Yep I also got it working by using the format of the child page you use (setting the date manually)

Sogl:
Hah... it reproduces on my Windows machine in the same way. No matter what date format is setting in post.

Sogl:
I think I found a spot where problem starts:
https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Page/Collection.php#L273

Setting `end: now +21 years` is working, but `end: now +22 years` is not.

flaviocopes:
Ok so this problem is specific to 32-bit systems, where the dates end on `2038-01-19 03:14:07`, and 22 years point us over that date - would have never thought about that.

Testing the PR

flaviocopes:
Merged PR

PR #902, opened 2016-06-15
Remove end date check if not specified (better 32-bit support)

Sogl:
Related to https://github.com/getgrav/grav/issues/898

I can't find better way to remove it with DRY principle :expressionless:
With smaller code we need to check a value for each `$this->items` iteration.

Also it can be used with `PHP_INT_MAX` with previous version of this code.

### DIFF
--- a/system/src/Grav/Common/Page/Collection.php
+++ b/system/src/Grav/Common/Page/Collection.php
@@ -270,17 +270,30 @@ class Collection extends Iterator
     public function dateRange($startDate, $endDate = false, $field = false)
     {
         $start = Utils::date2timestamp($startDate);
-        $end = $endDate ? Utils::date2timestamp($endDate) : strtotime("now +1000 years");
+        $end = $endDate ? Utils::date2timestamp($endDate) : false;
 
         $date_range = [];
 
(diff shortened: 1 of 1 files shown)


flaviocopes:
Want to try this solution? (moved the `if` inside the loop)

```php
    public function dateRange($startDate, $endDate = false, $field = false)
    {
        $start = Utils::date2timestamp($startDate);
        $end = $endDate ? Utils::date2timestamp($endDate) : false;

        $date_range = [];

        foreach ($this->items as $path => $slug) {
            $page = $this->pages->get($path);
            $add = false;
            
            if ($page !== null) {
                $date = $field ? strtotime($page->value($field)) : $page->date();

                if ($end) {
                    if ($date >= $start && $date <= $end) {
                        $add = true;
                    }
                } else {
                    if ($date >= $start) {
                        $add = true;
                    }
                }
                
                if ($add) {
                    $date_range[$path] = $slug;
                }
            }
        }

        $this->items = $date_range;

        return $this;
    }

```

Sogl:
It looks better, but brings +2 `if` conditions on each item iteration...

flaviocopes:
Merged yours

PR #906, opened 2016-06-16
Respect `enable_asset_timestamp` settings for pipelined Assets

Sommerregen:
I don't know if this is intentional, but the enabled `enable_asset_timestamp` settings is not respected for pipelined assets. Even if I disable this setting the cache key is always added.

I know why the cache key should be present due to browser cache invalidation of changes. However what I don't understand is then, why this setting currently only affects non-pipelined assets (see e.g. [Assets.php#L556](https://github.com/Sommerregen/grav/blob/4cec9851a1f51d78894ee2b5cab28570a3e343e8/system/src/Grav/Common/Assets.php#L556)).

This PR is for the lazy-man. Instead of opening an issue we can discuss it here or if this behavior is intentional, feel free to close it.

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -668,17 +668,14 @@ class Assets
      */
     protected function pipelineCss($group = 'head')
     {
-        /** @var Cache $cache */
-        $cache = Grav::instance()['cache'];
-        $key = $cache->getKey();
-
         // temporary list of assets to pipeline
         $temp_css = [];
 
         // clear no-pipeline assets lists
         $this->css_no_pipeline = [];
 
-        $uid = md5(json_encode($this->css) . $this->css_minify . $this->css_rewrite . $group . $key);
+        // Compute uid based on assets and timestamp
+        $uid = md5(json_encode($this->css) . $this->css_minify . $this->css_rewrite . $group . time());
         $file =  $uid . '.css';
         $inline_file = $uid . '-inline.css';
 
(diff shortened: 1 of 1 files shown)


flaviocopes:
I think it's simply missing in that case, no real reason behind that. 

But it should also take in consideration the `enable_asset_timestamp: false` case, right?

Sommerregen:
@flaviocopes That's a good point! I'll wait for @rhukster maybe he can shed some more light on it. Concerning the `enable_asset_timestamp: false` case, it runs as follows: `$this->timestamp` is `''` as this is the default (L80) in (L180) it gets set to the Grav cache key.

Sommerregen:
Ok, I fully understand. Wasn't aware that the hash is based on the files involved. Thought it was somehow also derived from Grav cache key... But you are right ([L677](https://github.com/Sommerregen/grav/blob/4cec9851a1f51d78894ee2b5cab28570a3e343e8/system/src/Grav/Common/Assets.php#L677)) and I really like your idea! :-D I have updated my PR. Hope you like it.

rhukster:
Actually I don't think it should be the cache key, rather an actual time stamp of now

rhukster:
looks good to me!

Sommerregen:
After a good sleep, I don't think that having such a timestamp with `time()` is stable enough not to build the pipeline any time again a user requests a page. I think we should gather all modified times in the `addJs` and `addCss` methods. For local assets that's easy and for remote ones providing a default (a constant, like 0).

rhukster:
Actually you are right, this won't work correctly because the file will be generated every time. We need a last generated timestamp that is actually stored in the `assets/` folder I think.  Then we can reuse this timestamp to build the filename when looking to see if the file exists.

rhukster:
i've switched it back to cache key for the time being so it will at least not recreate the pipeline everytime.

PR #908, opened 2016-06-16
Optimize PR #902

Sommerregen:
This PR optimizes #902 according to the DRY principle. Now only two instead of three if conditions are needed in each loop and redundant code was removed.


flaviocopes:
👍

Sogl:
That's something that I want :wink: :+1:

PR #913, opened 2016-06-21
Add user to config

Vivalldi:
I have a feeling that this is too "simple" in that no user specific config is merged with other configs but i stayed away from doing that so that plugin developers can choose whether to use user based config (for example the login plugin redirect) or if the want to use overarching config.

### DIFF
--- a/system/src/Grav/Common/User/User.php
+++ b/system/src/Grav/Common/User/User.php
@@ -29,7 +29,9 @@ class User extends Data
      */
     public static function load($username)
     {
-        $locator = Grav::instance()['locator'];
+        $grav = Grav::instance();
+        $locator = $grav['locator'];
+        $config = $grav['config'];
 
         // force lowercase of username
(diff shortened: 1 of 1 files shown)


flaviocopes:
Can you describe the end goal, what will this change allow to do?

Vivalldi:
This will allow user specific config to be retrieved with `config->get('user.field.subfield`);` in the same way that plugin, theme, and site config is grabbed.

That being said it is still possible in plugins to retrieve user config by using `$grav['user']['field']['subfield'];`

flaviocopes:
I think it can be useful, letting @rhukster the final word :)

rhukster:
is this reliable between pages?

Vivalldi:
Between pages? As in, if I log in and then navigate to another page, and then another?

rhukster:
Exactly... i merged it then thought about this as there is not going to be any persistence to the compiled configuration.

PR #917, opened 2016-06-25
Feature: Better assets pipelining

Sommerregen:
This PR is an enhancement of PR #906, where we have used the cache key as a stable key for pipeline generation. Now whenever an asset has changed the pipeline will be regenerated again. This is possible, because a modification time is added to each (local) asset. Remote assets are ignored so far, since it is difficult to actually retrieve a reliable modification time...

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -361,7 +361,7 @@ class Assets
             'asset'    => $asset,
             'priority' => intval($priority ?: 10),
             'order'    => count($this->js),
-            'pipeline' => (bool)$pipeline,
+            'pipeline' => (bool) $pipeline,
             'loading'  => $loading ?: '',
             'group'    => $group ?: 'head',
             'modified' => $modified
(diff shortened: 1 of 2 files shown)


MoisesCardenas:
I encountered and issue on how the $uid is been generated and want to get some feed back, on our current environment we  have two servers under a load balancer, the two servers contains an exact copy of the Grav website and this had been working really well until we enabled the Pipeline feature, we started getting 404 on css and on js even that the assets where inserted on the html "/assets/f51423af2c4b570cd1e444013972b42c.css", I did some tests and I found the problem:

`$uid = md5(json_encode($this->css) . $this->css_minify . $this->css_rewrite . $group);` 
and
`$uid = md5(json_encode($this->js) . $this->js_minify . $group);`

The arrays **$this->css** and **$this->js** contains a property called **"modified"** which will always be different for each server and will make the **$uid** to be different for the same resources, this caused a weird behavior where some times the page loads with **uids** names of resources that only exists on server 2 and vice versa, therefore the 404.

I was able to fix this by removing the **json_encode($this->css)** and **json_encode($this->js)** from the md5 after that the uid names on the assets was matching regardless of the server  so no more 404 and the page was loading correctly, I think this is not and ideal solution, any thoughts on this?

flaviocopes:
I think you need session persistence to make sure that users are always directed to the same server during the same session, otherwise you'll also run into cookies / session issues. The same issue applies to cached resources like processed images, which are stored with a random name.

MoisesCardenas:
@flaviocopes that in deed is the way to go I complete forgot that AWS ELB has a feature called sticky sessions which will make users to always hit the same server during the session, I enable it and now the website is working like a charm.

Issue #892, opened 2016-06-12
Feature Request: Add more options to `bin/gpm update` command


Sommerregen:
Currently, `bin/gpm update` updates all packages, which have changed during the last update. Updating certain package classes like "only update all plugins" or "only update all themes" is not possible. What I'd like to see is to have options `--plugins` and `--themes` for the update command to do something like this

```
bin/gpm update --plugins
bin/gpm update --themes
```

A `--quiet` command may also be useful, which installs the plugin/theme without the need of any user interaction.

rhukster:
Good options 👍 

PR would be lovely! :)

PR #921, opened 2016-06-30
Feature/more options gpm #892

rebangm:
Hello,

This PR add options to update only plugins or themes when you 're running bin/gpm update.
Only plugins : bin/gpm update --plugins
Only themes : bin/gpm update --themes

PR Idea from issue https://github.com/getgrav/grav/issues/892, when PR merged fixes https://github.com/getgrav/grav/issues/892

PS : I don't know codeception, and there is no command Test, so I don't write UpdateCommandTest.

Bye

### DIFF
--- a/system/src/Grav/Common/GPM/GPM.php
+++ b/system/src/Grav/Common/GPM/GPM.php
@@ -166,19 +166,28 @@ class GPM extends Iterator
     /**
      * Returns an array of Plugins and Themes that can be updated.
      * Plugins and Themes are extended with the `available` property that relies to the remote version
+     * @param $list_type_update specifies what type of package to update
      * @return array Array of updatable Plugins and Themes.
      *               Format: ['total' => int, 'plugins' => array, 'themes' => array]
      */
-    public function getUpdatable()
+    public function getUpdatable($list_type_update)
     {
-        $plugins = $this->getUpdatablePlugins();
-        $themes  = $this->getUpdatableThemes();
-
+        $total_counts = [];
+        $list_to_update = [];
+        if ($list_type_update['plugins']) {
+            $list_to_update['plugins'] = $this->getUpdatablePlugins();
+            $total_counts[] = count($list_to_update['plugins']);
+        }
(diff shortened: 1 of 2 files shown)


Vivalldi:
What happens if you don't pass an argument?

Just by briefly looking at it, there is no logic to handle if neither is set

rebangm:
Yes, you're right, i need to handle when none of two options are set.
I'll fix it.

Thx

rebangm:
@Vivalldi change done, same behavior like I said previously and now 
```bin/gpm update ``` -> update plugins and themes

**Comment by Sommerregen** [comment]
@rebangm Nice! However your code currently does not conform with the PSR0/2 standards (see http://www.php-fig.org/psr/psr-2/). Can you change that, too?

**Comment by rebangm** [comment]
@Sommerregen I'll check my IDE, I have PSR-2 configuration, maybe a mistake from me 
I'll use Php-CS-Fixer to check.

Thx

**Comment by rebangm** [comment]
@Sommerregen Done, my bad !! I don't put my pre-commit git hook for PSR-2 compliant.
lots of lines changed with my hook, not only my code.
Tell me if I revert commit and fix PSR-2 only on my code.

Thx

**Comment by Sommerregen** [comment]
@rebangm Having a Php-CS-Fixer run in the background is fine and I like it. However @rhukster is the man who has decide it on the very end.

As I told you I'll check the code. If everything is fine and @rhukster back he will then merge it to the core.

P.S.: @rhukster Would be fine to run Php-CS-Fixer as a hook to the Grav repository, too!

**Comment by Vivalldi** [comment]
Well there are a few things that break out of standard for readability, such as,
```
$someArray['withKey`]              = 'value';
$someArray['withAnotherKey']  = 'anotherValue';
```
The fixer is going to change it to simply be
```
$someArray['withKey`] = 'value';
$someArray['withAnotherKey'] = 'anotherValue';
```

The code works fine for me 👍

flaviocopes:
Fine for me!

PR #922, opened 2016-06-30
add ext-zip requirement

rebangm:
Hello,

I had ext-zip requirement in composer.json due to ZipArchive Usage.
When ext-zip isn't install, bin/gpm doesn't work.

Do you have a check config script, same as Symfony when installing for first time ?

bye


flaviocopes:
The Problems plugin runs the first page load and checks for other requirements, but not for Zip support, but it should do it.

PR #940, opened 2016-07-16
update some vietnamese translations

minhchu:


MATsxm:
@minhchu 
Please use the Crowdin platform to edit Translations.
Vietnamese is already available here: 
https://crowdin.com/project/grav-core
and also for the Admin plugin here:
https://crowdin.com/project/grav-admin
Thank you for your precious help,

minhchu:
Thanks. I understand. I'll use the crowdin

minhchu:
btw, can I translate the  `title` and `Path` ?

MATsxm:
@minhchu 
Nope, you **NEVER** have to translate keys (you won't be able on Crowdin anyway) as they are **not** variable and the unique reference you want to work with. It is the "word" that will let the sytem knows what it has to "replace" then it must be ALWAYS the same.
The value it is a variable that can change, depending on the language for example.

**in a way**, you are saying to the system: each time you see `Title` in the code, then use: `Hello world` if the site is set in english **OR** `Bonjour le monde` if the same site is set in french.
If you translate the key `Title`, the system won't be able to know what to use.

Not sure I'm really clear on this one ;)
Nevermind, on crowdin, you will be able to translate only the values, and this is the CORRECT way to deal with.

flaviocopes:
Thanks @MATsxm for clearing this up, as you said translations go in Crowdin. Closing

minhchu:
Hi again. I dont know why but I still can edit the `title` and `Path` on crowdin: https://crowdin.com/translate/grav-core/5/en-vi

MATsxm:
@minhchu 
you can **NOT** and must **NOT**
Keys must NOT be translated.
Please translate all the strings available and you will see that everything will be perfect for your site in vietnamese.

minhchu:
I understand what you said. It's just my curiosity that crowdin didnt raised any warning when I was doing so.

MATsxm:
Agree that the context and the explanation should be added to avoid confusion...
That's the job of the T9n Project Manager but this is another question... :wink:

PR #948, opened 2016-07-20
Improved `authorize` Twig extension

Sommerregen:
This PR addresses https://getgrav.org/forum#!/general:hiding-certain-tabs-from-no and enhances the `authorize` Twig extension method with full backward-compatibility. Thus, it can be used as usual with

```php
authorize(['admin.super', 'admin.pages']);
authorize('admin.super');
```

or with the new changes with

```php
authorize([
  'admin' => [
    'super' => true
  ],
  'site' => [
    'user' => false
  ]
]);
```

which makes it possible to set access rights for tabs via blueprints as

````yaml
assets:
  type: tab
  title: Assets
  security:
    admin:
      super: true
    site:
      user: false
  fields:
     ...
```

### DIFF
--- a/system/src/Grav/Common/Twig/TwigExtension.php
+++ b/system/src/Grav/Common/Twig/TwigExtension.php
@@ -729,11 +729,15 @@ class TwigExtension extends \Twig_Extension
     }
 
     /**
-     * Authorize an action. Returns true if the user is logged in and has the right to execute $action.
+     * Authorize an action. Returns true if the user is logged in and
+     * has the right to execute $action.
      *
-     * @param string $action
-     *
-     * @return bool
+     * @param  string|array $action An action or a list of actions. Each
+     *                              entry can be a string like 'group.action'
+     *                              or without dot notation an associative
+     *                              array.
+     * @return bool                 Returns TRUE if the user is authorized to
+     *                              perform the action, FALSE otherwise.
      */
     public function authorize($action)
     {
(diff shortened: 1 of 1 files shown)


flaviocopes:
👍 thanks!

flaviocopes:
Just to clear things,

```
    admin:
      super: true
    site:
      user: false
```

means a tab is visible to users with `admin.super` who do not have `site.user` set to true? Or what is the logic there?

Sommerregen:
@flaviocopes Good point! I tried to adapt the logic of the Login plugin. Atm it means what you said. Only users with `admin.super` are allowed to access the tab. Users with `site.user` not. I see this is redundant, because all other roles do not have access either... and it is not the same as the login plugin provides :-/ ...

Maybe it is better to replace [L749-756](https://github.com/getgrav/grav/pull/948/files#diff-ccac52608947fd35efb86786423e0527R753) with

```php
foreach ($action as $key => $rules) {
  $prefix = is_int($key) ? '' : $key . '.';
  $rules = $prefix ? (array) $rules : [$rules => true];
  foreach ($rules as $rule => $value) {
    if ($this->grav['user']->authorize($prefix . $rule) === $value) {
      return true;
    }
}
```

like [the logic from the login plugin](https://github.com/getgrav/grav-plugin-login/blob/develop/login.php#L440-L444).

flaviocopes:
My question is, in what case setting a value as false is a good use case? I don't see one at hand, but maybe you do, as you wrote the PR :)

Issue #247, opened 2015-08-06
Chapter about Nginx at learn.getgrav.org


kris4de:
I had some trouble with the preconfigured nginx.conf on my DigitalOcean Ubuntu VM and I think one chapter about configuration of Nginx would be helpful.

Some lines didn't work for my virtual host (or server block) configuration and others are not needed, I think.

```
worker_processes  1;

events {
    worker_connections  1024;
}
```

According to [this full example](http://wiki.nginx.org/FullExample), these lines are already default and not needed.

```
http {
    include       mime.types;
    default_type  application/octet-stream;
    sendfile        on;
    keepalive_timeout  65;
```

This surrounding `http` block throws errors and doesn't work with virtual host configuration – Nginx can't restart.

```
location / {
            root   html;
                index  index.php;
        if (!-e $request_filename){ rewrite ^(.*)$ /index.php last; }
    }
```

In default configuration, the first two lines inside `location /` are already present. But outside of `location /`. And I am not sure whether they belong in there, because it's not working properly – 404 errors (not Grav, Nginx) are the result.

As a conclusion, at least one chapter about Nginx configuration would be really appreciated.

austinbutler:
I'd agree on having a little more NGINX in the manual. At the very least linking to the NGINX conf file included with the package itself: https://github.com/getgrav/grav/blob/develop/nginx.conf

Once I realized it was there, I had it working in seconds.

rhukster:
Would be awesome if someone who knows Nginx better than me could draft up a version of this documentation that we can add to the learn docs??

DarrylDias:
If you are trying to configure Grav with NGINX running on Ubuntu VPS on Digital Ocean. I would recommended editing the `/etc/nginx/sites-enabled/default` file instead of directly editing the `/etc/nginx/nginx.conf`. I have created a [gist](https://gist.github.com/DarrylDias/2a04dba9688e131eaced). That contains my current NGINX configuration that is running my current Grav [website](http://darryldias.me).

greasan:
@DarrylDias I wrote about virtual host config. The `nginx.conf` is untouched. But thanks for your current configuration. I will have a look at it.

simonhaenisch:
@rhukster I could help with that, where would you place it in the docs? Mabybe in Basics > Installation or Cookbook or Troubleshooting?

flaviocopes:
[...]

PR #975, opened 2016-08-11
better nginx config, renamed

simonhaenisch:
This is a better Nginx configuration. I did it as part of creating a page at [getgrav/grav-learn](https://github.com/getgrav/grav-learn) about how to use Grav with Nginx. See issue #247 and PR getgrav/grav-learn#302 

I renamed `nginx.conf` to just `nginx`, because `nginx.conf` is the name of the main Nginx config file and naming it like that could lead to people replacing the main config with this file, which would break the Nginx config. This file is supposed to be copied into `/etc/nginx/sites-available`.


simonhaenisch:
I just saw that you prefer to have a `feature/` branch instead of directly merging to `develop`. Do you want me to change it, i.e. open a new pull request?

flaviocopes:
Actually, feature branches are cool for moving forward something, internally mostly, but for contributions, GitHub PRs are best in this way - it's mostly the same thing.

rhukster:
will review this during the coming week.  Thanks!

simonhaenisch:
I've done a few improvements (IPv6, deny system files, deny hidden files) and a fix regarding url parameters. Before there was

```nginx
try_files $uri $uri/ /index.php?_url=$uri;
```

First I thought that the `_url` parameter was necessary but after looking into the Grav source code I found out that is not true. So I removed the `?_url=$uri` part, but then got 404 for urls with parameters. Adding a random parameter (like `?foo=bar`) made it work again, but the url parameters where not applied (e.g. image crop), so it seems this always "overrides" the real url parameters, because they are never passed. After some more digging I found the [Nginx Docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#variables) and now I know how to correctly pass the url parameters:

```nginx
try_files $uri $uri/ /index.php$is_args$args;
```

`$is_args` adds the `?` only if there are arguments.

simonhaenisch:
I just read <https://learn.getgrav.org/hosting/vps/digitalocean#configure-nginx-connection-pool> and there the thing with `try_files` also needs to be fixed.

rhukster:
Yah I was going to review this nginx conf and merge it with the docs next.

rhukster:
I also prefer to to use user/group for nginx permissions whenever possible for simplicity

simonhaenisch:
Sry, what do you mean by that?

simonhaenisch:
any updates on this, will the PR be merged eventually? I'd like to get rid of my fork one day :nerd_face: 

I don't get why it failed the check, is it possible to re-run the test?

Issue #1016, opened 2016-09-01
Retina images in combination with manual derivatives


fredrikekelund:
I want to generate srcset attributes for the images I'm using so that they both scale down on small displays and up on retina displays. Grav's methodology for this seems logical enough, in a first sweep I appended "@2x" to the filenames, to get it working on retina displays. That worked fine. Then I wanted them to scale down as well, so I manually called the `derivatives` method as well. This made the images scale down on mobile, **but** it also put a limit on the maximum width, so that the largest source only had half the width of the original image.

I'm not sure exactly where the flawed logic is here, but it seems to be in how the combination of `$alternatives` and `$derivatives` is treated in the `srcset` method of the `ImageMedium` class (see [here](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Page/Medium/ImageMedium.php#L207-L221)).

The `$alternatives` array should probably either be emptied if we use the derivative method manually, or srcset shouldn't work exclusively with either the `$alternatives` array or the `$derivatives` array

rhukster:
This is a bit gnarly and it's super hard to debug on my non-retina display.  I really need a new computer!

fredrikekelund:
The problem should be reproducible anyway, even if the difference is hard to tell. Here's how I encountered the issue:
1. Add a wide picture, 2500px in my case
2. Append "@2x" to the filename
3. Call the derivatives method in the twig template like `derivatives(500, 2500, 500)`
4. Check the srcset attribute, the widest version of the image will be just 1250px, half of the original image

rhukster:
I've added a Hacktoberfest tag to see if we can get some help with this as I'm a bit stretched with other things at the moment.

fredrikekelund:
This one is hopefully addressed by my PR, so I think we just need to iron that issue out to be able to close this. Maybe this isn't that much of low hanging fruit either

fredrikekelund:
Pregenerated alternatives and alternatives that are generated using ImageMedium#derivatives now use the same underlying method, which should make things more predictable.

PR #1035, opened 2016-09-12
ImageMedium#derivatives now works with image filters

fredrikekelund:
My original motivation for writing this PR was the problem described in #1016, where using the `derivatives` method on an image in a Twig template wouldn't work well in combination with the other method of producing responsive images in Grav - appending a pixel density indicator to the image's filename (eg. "@2x").

The underlying problem here also meant that image filters would not work in combination with image derivatives. They would work with image alternatives, which is what was generated from appending "@2x" to the filename of the image, but it wouldn't work with derivatives.

This commit fixes both the first and the second problem by making it so that the ImageMedium#derivatives method doesn't add image variants to a separate array, but rather to the same array as it would have if the filename method was used.

I've made some line comments to highlight a few things in the code as well!

### DIFF
--- a/system/src/Grav/Common/Page/Medium/ImageMedium.php
+++ b/system/src/Grav/Common/Page/Medium/ImageMedium.php
@@ -228,7 +228,7 @@ class ImageMedium extends Medium
         // Clear any alternatives that have already been implied by the image's
         // filename (eg. ones that had "@2x" or similar appended to their
         // filenames)
-        $this->clearAlternatives();
+        $this->reset();
 
         while ($width <= $max_width) {
             $ratio = $width / $this->get('width');
--- a/system/src/Grav/Common/Page/Medium/ImageMedium.php
+++ b/system/src/Grav/Common/Page/Medium/ImageMedium.php
@@ -239,12 +239,20 @@ class ImageMedium extends Medium
             }
 
             $derivative = MediumFactory::fromFile($this->get('filepath'));
-            $derivative_width = $derivative->get('width') * $ratio;
-            $derivative_height = $derivative->get('height') * $ratio;
-            $derivative->resize($derivative_width, $derivative_height);
-            $derivative->set('width', $derivative_width);
-            $derivative->set('height', $derivative_height);
-            $this->addAlternative($ratio, $derivative);
+
+            // It's possible that MediumFactory::fromFile returns null if the
+            // original image file no longer exists and this class instance was
+            // retrieved from the page cache
+            if (isset($derivative)) {
+                $derivative_width = $derivative->get('width') * $ratio;
+                $derivative_height = $derivative->get('height') * $ratio;
+
+                $derivative->resize($derivative_width, $derivative_height);
+                $derivative->set('width', $derivative_width);
+                $derivative->set('height', $derivative_height);
+
+                $this->addAlternative($ratio, $derivative);
+            }
         }
 
         return $this;
--- a/system/src/Grav/Common/Page/Medium/ImageMedium.php
+++ b/system/src/Grav/Common/Page/Medium/ImageMedium.php
@@ -66,11 +66,6 @@ class ImageMedium extends Medium
     ];
 
     /**
-     * @var array
-     */
-    protected $derivatives = [];
-
-    /**
      * @var string
      */
     protected $sizes = '100vw';
@@ -197,28 +192,19 @@ class ImageMedium extends Medium
      */
     public function srcset($reset = true)
     {
-        if (empty($this->alternatives) && empty($this->derivatives)) {
+        if (empty($this->alternatives)) {
             if ($reset) {
                 $this->reset();
             }
+
             return '';
         }
 
-        if (!empty($this->derivatives)) {
-          asort($this->derivatives);
-
-          foreach ($this->derivatives as $url => $width) {
-              $srcset[] = $url . ' ' . $width . 'w';
-          }
-
-          $srcset[] = $this->url($reset) . ' ' . $this->get('width') . 'w';
+        $srcset = [];
+        foreach ($this->alternatives as $ratio => $medium) {
+            $srcset[] = $medium->url($reset) . ' ' . $medium->get('width') . 'w';
         }
-        else {
-          $srcset = [ $this->url($reset) . ' ' . $this->get('width') . 'w' ];
-          foreach ($this->alternatives as $ratio => $medium) {
-              $srcset[] = $medium->url($reset) . ' ' . $medium->get('width') . 'w';
-          }
-        }
+        $srcset[] = $this->url($reset) . ' ' . $this->get('width') . 'w';
 
         return implode(', ', $srcset);
     }
@@ -232,31 +218,36 @@ class ImageMedium extends Medium
      * @return $this
      */
     public function derivatives($min_width, $max_width, $step = 200) {
-      $width = $min_width;
+        $width = $min_width;
 
-      // Do not upscale images.
-      if ($max_width > $this->get('width')) {
-        $max_width = $this->get('width');
-      }
-
-      while ($width <= $max_width) {
-        $ratio = $width / $this->get('width');
-        $derivative = MediumFactory::scaledFromMedium($this, 1, $ratio);
-        if (is_array($derivative)) {
-          $this->addDerivative($derivative['file']);
+        // Do not upscale images.
+        if ($max_width > $this->get('width')) {
+            $max_width = $this->get('width');
         }
-        $width += $step;
-      }
-      return $this;
-    }
 
-    /**
-     * Add a derivative
-     *
-     * @param  ImageMedium $image
-     */
-    public function addDerivative(ImageMedium $image) {
-      $this->derivatives[$image->url()] = $image->get('width');
+        // Clear any alternatives that have already been implied by the image's
+        // filename (eg. ones that had "@2x" or similar appended to their
+        // filenames)
+        $this->clearAlternatives();
+
+        while ($width <= $max_width) {
+            $ratio = $width / $this->get('width');
+            $width += $step;
+
+            if ($ratio == 1) {
+                continue;
+            }
+
+            $derivative = MediumFactory::fromFile($this->get('filepath'));
+            $derivative_width = $derivative->get('width') * $ratio;
+            $derivative_height = $derivative->get('height') * $ratio;
+            $derivative->resize($derivative_width, $derivative_height);
+            $derivative->set('width', $derivative_width);
+            $derivative->set('height', $derivative_height);
+            $this->addAlternative($ratio, $derivative);
+        }
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


system/src/Grav/Common/Page/Medium/MediumFactory.php
```diff
@@ -119,8 +119,8 @@ public static function scaledFromMedium($medium, $from, $to)
         }
 
         $ratio = $to / $from;
-        $width = (int) ($medium->get('width') * $ratio);
-        $height = (int) ($medium->get('height') * $ratio);
+        $width = $medium->get('width') * $ratio;
```
fredrikekelund: This isn't directly related to the PR, but I've previously found that downscaled images sometimes got a white stripe at the far right edge. Not rounding the pixel values sent to Gregwar/Image fixed that

system/src/Grav/Common/Page/Medium/ImageMedium.php
```diff
@@ -232,31 +218,36 @@ public function srcset($reset = true)
...
+
+        while ($width <= $max_width) {
+            $ratio = $width / $this->get('width');
+            $width += $step;
+
+            if ($ratio == 1) {
+                continue;
+            }
+
+            $derivative = MediumFactory::fromFile($this->get('filepath'));
```
fredrikekelund: Even though it would be cleaner to use `MediumFactory::scaledFromMedium` here, I seemed to encounter some problems with it, presumably since it operates on the same original medium. Theoretically, it might be beneficial for performance to avoid the extra disk write that is required by that method.

system/src/Grav/Common/Page/Medium/ImageMedium.php
```diff
@@ -232,31 +218,36 @@ public function srcset($reset = true)
...
-     * Add a derivative
-     *
-     * @param  ImageMedium $image
-     */
-    public function addDerivative(ImageMedium $image) {
-      $this->derivatives[$image->url()] = $image->get('width');
+        // Clear any alternatives that have already been implied by the image's
+        // filename (eg. ones that had "@2x" or similar appended to their
+        // filenames)
+        $this->clearAlternatives();
```
fredrikekelund: This highlights an issue with the `derivatives` method. When using it in Twig templates, any filters applied before I call it won't affect the derivatives. It would of course be best if the order in which I called the methods didn't matter, but maybe we ought to call `$this->reset()` here to make sure the original medium and the derivatives at least look the same when rendered?

system/src/Grav/Common/Page/Medium/ImageMedium.php
```diff
@@ -197,28 +192,19 @@ public function cache()
...
-          asort($this->derivatives);
-
-          foreach ($this->derivatives as $url => $width) {
-              $srcset[] = $url . ' ' . $width . 'w';
-          }
-
-          $srcset[] = $this->url($reset) . ' ' . $this->get('width') . 'w';
-        }
-        else {
-          $srcset = [ $this->url($reset) . ' ' . $this->get('width') . 'w' ];
```
fredrikekelund: Previously, calling `$this->url($reset)` here would reset the current medium and remove the registered alternatives. I wasn't sure what the purpose was, but reversing the order so that the `url` method of each alternative is called before the original medium's method did the trick

rhukster:
Wow this is the nicest documented PR I've ever received. Love it!  Will review and test.

fredrikekelund:
Thanks! :) The changes so far did fix the fact that general image operations weren't applied to derivatives, but I found that I stumbled a bit with the "@2x" filename plus the derivatives method, it's not working quite the way I wanted it to just yet. I'm working on it now, so will hopefully push another commit pretty soon!

fredrikekelund:
@rhukster alright, I've pushed another commit. Hopefully that should do it!

system/src/Grav/Common/Page/Medium/ImageMedium.php
fredrikekelund: I've done some testing with regards to this. In a case where I have an image with a filename that ends with "@2x", and I call `derivatives()` on that medium, then the fact that the medium is reset by default when calling `srcset()` causes two full-width copies to be generated if `url()` is also called on the medium. This is clearly an edge case, but since I was a bit unsure of what the reset here was for anyway, I thought I should mention it.

rhukster:
Sorry got busy this past week.. will check this out soon.

system/src/Grav/Common/Page/Medium/ImageMedium.php
```diff
@@ -232,31 +218,36 @@ public function srcset($reset = true)
...
+        while ($width <= $max_width) {
+            $ratio = $width / $this->get('width');
+            $width += $step;
+
+            if ($ratio == 1) {
+                continue;
+            }
+
+            $derivative = MediumFactory::fromFile($this->get('filepath'));
+            $derivative_width = $derivative->get('width') * $ratio;
```
fredrikekelund: I've been using this fork for my own project, and I've run across an error on this row a few times. `MediumFactory::fromFile` seems to have returned null, presumably because `$this->get('filepath')` in turn returned null. Both times this happened was after I had removed content, and the problem went away by restarting the `php-fpm` service on the server, so I suspect Grav has somehow cached the content associated with a page and then runs into an error.

Does this seem like a separate issue, or do we need to check to ensure the filepath isn't null or something here?

rhukster: There is a check in the first line of that `Medium::fromFile()` method that checks the file exists, it returns null if not, so probably at least need to check for null before using `$derivative`.

fredrikekelund: Hmm, I see. How come a medium can be constructed in the first place if the corresponding file doesn't exist? I mean, if `MediumFactory` calls `file_exists`, how come the `Media` class that assembles page media doesn't?

fredrikekelund: Let me know if you want me to add a null check here or if it'd better to make changes further up the hierarchy, @rhukster!

system/src/Grav/Common/Page/Medium/ImageMedium.php
fredrikekelund: After some pretty basic profiling of a very image heavy page, I'm a bit concerned with this method. Reading the image from disk every time seems to have a significant impact on performance, almost on par with the image processing itself.

This PR saves an additional disk write, which *could* be beneficial for performance, I guess. Since reading/writing is slow, avoiding touching disk is of course good, but with `MediumFactory::scaledFromMedium`, the file that we're reading might be getting incrementally smaller, which could also be beneficial for performance if the original image is very large.

rhukster: I appreciate you digging into this.  It was originally developed by another user and i'm not as familiar with this section of Grav as the others.

Are you saying your PR improves the performance by not writing so much?

fredrikekelund: I haven't done any benchmarking between the current implemention and this patch, but yes this could improve performance.

While `MediumFactory::scaledFromMedium` (which is used in the current implementation) performs both a write and a read in this step, this patch performs just a read. The difference is that this patch always reads the original file, while the old implementation only reads the original file once, writes it several times, and then reads the copy it wrote when adding the derivative. If the original file is very large, I can imagine that performance would be better by not reading *that* file several times, but I don't really know without measuring.

system/src/Grav/Common/Page/Medium/ImageMedium.php
fredrikekelund: Alright, getting through this was a real ordeal, but it looks like I figured out what the issue was.

At first I thought it was PHP that was returning errant results when listing the contents of a directory, but it turns out that it was Grav's page object caching mechanism that didn't pick up on updates to just the media files of a page object. I described this issue in #1080 as well. 

`Media::__construct` (which initializes a page's media), is allowed to continue even if `MediumFactory::fromFile` returns null. This patch doesn't take that risk into account, and IMO, it shouldn't either. If the original `ImageMedium` was initiated, it should reliably point to an actual file.

Would be awesome if we could work through this final issue and get this PR merged, @rhukster!

fredrikekelund: If the cache is working as intended and it's a conscious convention to return null when the image can't be found, then we should probably check if the method call returned null. But in that case maybe we can have a discussion about how the caching mechanisms can be clarified in the docs (but that's better done in #1080).

fredrikekelund:
@rhukster I've pushed another commit that accounts for the risk of the image derivatives being generated from a cached ImageMedium instance and the original file no longer existing. I've also verified that this works (it results in an "error" image in the browser). This should hopefully be good for a merge now!

fredrikekelund:
One more fix! I've been testing out this patch a lot recently, and it's worked well for me. But I realized that with pre-generated responsive image alternatives, it would auto-generate an "@1x" alternative if such a file wasn't found. I've fixed that now so that the first image alternative it looks for is "@2x"

rhukster:
BTW you might want to merge in this commit: https://github.com/getgrav/grav/commit/31e358ca7c4e603a5517b27142fbffa64a22c9a3

It adds the correct 'prettyname' to the generated alternative images.

```
<img alt="Retina Image" src="/grav-retina/images/d/2/5/2/9/d2529b606ad3a797704c18099c574995d72b7eda-48a4355cc289c3a5f2520c1cf804c9e0b4e50078-sierra1x.jpeg" srcset="
/grav-retina/images/d/2/5/2/9/d2529b606ad3a797704c18099c574995d72b7eda-48a4355cc289c3a5f2520c1cf804c9e0b4e50078-sierra1x.jpeg 1706w, 
/grav-retina/images/2/a/a/c/b/2aacb1f0128f13fec18ecc7105f09c2ffd83bb85-sierra3x.jpeg 5120w, 
/grav-retina/images/5/d/f/f/6/5dff6396db4b21d969ede476ad209de8c876ebdc-05784e5c11e5ea1dc0770b0b75b520bd48cc9375-sierra2x.jpeg 3413w" sizes="100vw">
```

This is the same example as above.

rhukster:
Btw another commit (https://github.com/getgrav/grav/commit/92401de443c278fad29f06d3c53b9de73d74b2c6) fixes the issue with the increasing hashing when you append operations..results look better now:

```
<img alt="Retina Image" src="/grav-retina/images/4/8/a/4/3/48a4355cc289c3a5f2520c1cf804c9e0b4e50078-sierra1x.jpeg" srcset="
/grav-retina/images/4/8/a/4/3/48a4355cc289c3a5f2520c1cf804c9e0b4e50078-sierra1x.jpeg 1706w, 
/grav-retina/images/f/1/2/b/8/f12b8d3f8d049dc7ada651fd98fe22848e82361f-sierra3x.jpeg 5120w, 
/grav-retina/images/0/5/7/8/4/05784e5c11e5ea1dc0770b0b75b520bd48cc9375-sierra2x.jpeg 3413w" sizes="100vw">
```

fredrikekelund:
I knew it was going to bite me sooner or later that I forgot to do this in a feature branch... I've pushed a fix for the problem you described as well as merged your changes from `develop`, but I did it in a separate branch, so I'll go ahead and open a new PR!

PR #1038, opened 2016-09-13
New `bin/gpm direct-install` command

rhukster:
## Overview
This allows for direct installation of a local ZIP package or a remote URL based ZIP package.  

## Purpose

This command is intended for testing or troubleshooting where you might want to install a particular version of a plugin or a theme.  

## Usage

For example to install v1.0.1 of the **Cors** plugin:

```
bin/gpm direct-install https://github.com/getgrav/grav-plugin-cors/archive/1.0.1.zip
```

To install a local plugin:

```
bin/gpm direct-install /Users/bob/downloads/grav-plugin-archives.zip
```

This also works for Grav core, however, if you install a version prior to the introduction of `bin/gpm direct-install` you won't be able to use the command after you 'downgrade'

```
bin/gpm direct-install https://github.com/getgrav/grav/releases/download/1.1.4/grav-update-v1.1.4.zip
```

## Notes

1. This command will not allow installation on top of **symlinked directories** (that also goes for symlinked installations of Grav itself
2. This command **does not pay attention to dependencies**.  Please use `bin/gpm install` for that.

### DIFF
--- a/system/src/Grav/Console/Gpm/DirectInstallCommand.php
+++ b/system/src/Grav/Console/Gpm/DirectInstallCommand.php
@@ -120,19 +120,6 @@ class DirectInstallCommand extends ConsoleCommand
 
             if ($type == 'grav') {
 
-                $this->output->write("  |- Checking destination...  ");
-                if ($this->gravSymlinked()) {
-                    $this->output->write("\x0D");
-                    $this->output->writeln("  |- Checking destination...  <yellow>symbolic link</yellow>");
-                    $this->output->writeln("  '- <red>ERROR: symlinks found...</red> <yellow>" . GRAV_ROOT."</yellow>");
-                    $this->output->writeln('');
-                    exit;
-
-                } else {
-                    $this->output->write("\x0D");
-                    $this->output->writeln("  |- Checking destination...  <green>ok</green>");
-                }
-
                 $this->output->write("  |- Installing package...  ");
                 Installer::install($zip, GRAV_ROOT, ['sophisticated' => true, 'overwrite' => true, 'ignore_symlinks' => true], $extracted);
             } else {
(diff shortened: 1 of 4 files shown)


system/src/Grav/Console/Gpm/DirectInstallCommand.php
```diff
@@ -0,0 +1,381 @@
...
+     * Determine if Grav is symlinked (system is all we need to check)
+     *
+     * @return bool
+     */
+    protected function gravSymlinked()
+    {
+        if (is_link(GRAV_ROOT . DS . 'system')) {
+            return true;
+        }
+    }
```
w00fz: This is not needed, `Installer` as an internal method to verify validity of destination:

```php
Installer::isValidDestination(GRAV_ROOT . '/system');
if (Installer::IS_LINK === Installer::lastErrorCode()) { return false; } // Grav is symlinked
```

system/src/Grav/Console/Gpm/DirectInstallCommand.php
```diff
@@ -0,0 +1,381 @@
...
+     */
+    protected function isRemote($file)
+    {
+        if (filter_var($file, FILTER_VALIDATE_URL))
+        {
+            return true;
+        } else {
+            return false;
+        }
+    }
```
w00fz: This can be changed to the following, just looks cleaner:

```php
return (bool) filter_var($file, FILTER_VALIDATE_URL);
```

system/src/Grav/Console/Gpm/DirectInstallCommand.php
rhukster: Removed

system/src/Grav/Console/Gpm/DirectInstallCommand.php
rhukster: yup, cleaner.

Issue #1044, opened 2016-09-14
Errors in Admin Page - Plugins


synchro7:
I'm using local server "Xampp".
`Twig_Error_Runtime thrown with message "An exception has been thrown during the rendering of a template ("GPM not reachable. Please check your internet connection or check the Grav site is reachable") in "partials/plugins-list.html.twig" at line 19."

Stacktrace:
#31 Twig_Error_Runtime in C:\xampp\htdocs\grav-admin\vendor\twig\twig\lib\Twig\Template.php:402
#30 RuntimeException in C:\xampp\htdocs\grav-admin\system\src\Grav\Common\GPM\GPM.php:473
#29 Grav\Common\GPM\GPM:findPackage in C:\xampp\htdocs\grav-admin\vendor\twig\twig\lib\Twig\Template.php:604
#28 call_user_func_array in C:\xampp\htdocs\grav-admin\vendor\twig\twig\lib\Twig\Template.php:604
#27 Twig_Template:getAttribute in C:\xampp\htdocs\grav-admin\cache\twig\b6\b6a16b9e99993d5305205028e010e1f471b3dfa16d9791de3725566abe70287b.php:72
#26 __TwigTemplate_812dd152f02b4c4e995fb2a2e134fb73f18fa1fc2bdaa0c164751063205c25e5:doDisplay in C:\xampp\htdocs\grav-admin\vendor\twig\twig\lib\Twig\Template.php:387
#25 Twig_Template:displayWithErrorHandling in C:\xampp\htdocs\grav-admin\vendor\twig\twig\lib\Twig\Template.php:355
#24 Twig_Template:display in C:\xampp\htdocs\grav-admin\cache\twig\4d\4d52b4af49aed4b34218dfba0d4aacb0e49c5ec1836447acc79aafdc1f78bf2f.php:218
#23 __TwigTemplate_d45efbb2ecfd1b6d175e48517c93ba1d3d4dbe975bc59085cd8b1e31d7a923f9:block_content in C:\xampp\htdocs\grav-admin\vendor\twig\twig\lib\Twig\Template.php:167
#22 Twig_Template:displayBlock in C:\xampp\htdocs\grav-admin\cache\twig\bc\bcedad5a9d9c096c0555639fe81c8eae8ad587d56962c88e545ddb97fbb1e4cb.php:205
[...]

Issue #1008, opened 2016-08-26
GPM not reachable


dimlau:
[...]

Issue #850, opened 2016-05-27
Cannot connect to the GPM


FrenchW:
Hi, 
On each admin page, I have the red popup : "Cannot connect to the GPM"
I've red the help page here https://learn.getgrav.org/troubleshooting/common-problems and if it does not resolve the issue, it's said to keep in touch with you.

Hosting is in an OVH.com shared server (said "Mutualisé" in french)

# Configuration
## System
`System : Linux webm333.start.ha.ovh.net 3.14.60-grsec-hosting-web-3.14 #1 SMP Wed Feb 17 13:41:21 UTC 2016 x86_64`
`Build Date : Apr 29 2016 07:00:47`
`Configure Command : './configure' '--with-openssl' '--with-mysqli=mysqlnd' '--with-pdo-mysql=mysqlnd' '--prefix=/usr/local/php7.0' '--enable-fpm' '--enable-opcache' '--with-gd' '--enable-intl' '--enable-gd-native-ttf=/usr' '--with-png-dir=/usr' '--enable-wddx' '--with-config-file-path=/usr/local/php7.0/etc' '--enable-sigchild' '--enable-short-tags' '--disable-rpath' '--enable-libgcc' '--with-zlib=/usr' '--enable-bcmath' '--enable-calendar' '--with-curl=/usr' '--with-gdbm=/usr' '--enable-ftp' '--with-gettext=/usr' '--with-imap' '--with-kerberos' '--with-imap-ssl' '--with-mcrypt=/usr' '--with-mhash=/usr' '--with-jpeg-dir=/usr' '--enable-exif' '--enable-sysvsem' '--enable-sysvshm' '--enable-sysvmsg' '--enable-zip' '--enable-inline-optimization' '--enable-soap' '--enable-mbstring' '--enable-mbregex' '--with-gnu-ld' '--with-libxml-dir' '--enable-shared=yes' '--enable-static=yes' '--with-xsl=/usr' '--enable-sockets' '--with-xmlrpc' '--with-iconv' '--with-imagick' '--enable-pdo' '--with-pdo-pgsql=/usr' '--with-xpm-dir=/usr' '--with-pcre-regex' '--with-freetype-dir=/usr' '--with-pspell' '--with-pgsql=/usr' '--with-bz2' '--with-gmp' 'EXTENSION_DIR=/usr/local/php7.0/lib/php-extensions'`

## Curl
`cURL support : enabled`
`cURL Information : 7.38.0`
`Age : 3`

## OpenSSL
[...]

PR #1053, opened 2016-09-19
SSL Verify Peer

w00fz:
## Overview
In some occasions we've noticed people unable to connect to GPM. After further investigations it turns out that on some systems (Windows only has been confirmed so far), the SSL certificate cannot be verified, causing GPM to fail (references: #1044, #1008 and potentially #850).
The to go solution is to disable the peer verification from `cURL` and `fopen`.

## Purpose
This PR adds 2 new optional system config: `system.gpm.verify_peer` (default: `true`) and `system.gpm.method` (default: `auto`).

## `system.gpm.verify_peer`

When enabled, the SSL certificate gets verified by either fopen or cURL methods. If disabled the checks is skipped (this is not advised but might be the only viable solution for some people)

## `system.gpm.method`

By default is set to `auto`. Auto automatically tests if the `fopen` mechanism is available and if it is gets used. Otherwise it tries to see if `cURL` is available and if it is use that instead.

Usually `auto` works fine but there might be some instances where some might want to force either methods. This new setting will allow to do so.

## Notes
 * As long as the changes reported above, this PR also makes some code rearrangement for the Proxy, it should just work the same but might be good to get this reviewed/tested.
 * This PR also introduces settings for Admin, new lang have been added with https://github.com/getgrav/grav-plugin-admin/commit/0a2b1d790f4cb4b189837c215eeda158a75e44dd.

### DIFF
--- a/system/blueprints/config/system.yaml
+++ b/system/blueprints/config/system.yaml
@@ -982,6 +982,27 @@ form:
                     label: PLUGIN_ADMIN.PROXY_URL
                     help: PLUGIN_ADMIN.PROXY_URL_HELP
 
+                gpm.method:
+                    type: toggle
+                    label: PLUGIN_ADMIN.GPM_METHOD
+                    highlight: auto
+                    help: PLUGIN_ADMIN.GPM_METHOD_HELP
+                    options:
+                        auto: PLUGIN_ADMIN.AUTO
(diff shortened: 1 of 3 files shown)


w00fz:
Missing code ref. c4f71c9ddab5d3066948d2b16ba2047ef51cb89c

PR #1055, opened 2016-09-19
Added possiblity to connect to redis via a unix socket

baltendo:
When a socket path is specified it should be used and by default host port with the default values.

### DIFF
--- a/system/src/Grav/Common/Cache.php
+++ b/system/src/Grav/Common/Cache.php
@@ -196,14 +196,14 @@ class Cache extends Getters
 
             case 'redis':
                 $redis = new \Redis();
-                $socket = $this->config->get('system.cache.redis.socket');
+                $socket = $this->config->get('system.cache.redis.socket', false);
 
-                if (isset($socket)) {
+                if ($socket) {
                     $redis->connect($socket);
                 } else {
                     $redis->connect($this->config->get('system.cache.redis.server', 'localhost'),
                     $this->config->get('system.cache.redis.port', 6379));
-                }              
+                }
 
                 $driver = new DoctrineCache\RedisCache();
                 $driver->setRedis($redis);
--- a/system/src/Grav/Common/Cache.php
+++ b/system/src/Grav/Common/Cache.php
@@ -196,8 +196,15 @@ class Cache extends Getters
 
             case 'redis':
                 $redis = new \Redis();
-                $redis->connect($this->config->get('system.cache.redis.server', 'localhost'),
+                $socket = $this->config->get('system.cache.redis.socket');
+
+                if (isset($socket)) {
+                    $redis->connect($socket);
+                } else {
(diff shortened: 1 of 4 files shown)


system/src/Grav/Common/Cache.php
```diff
@@ -196,8 +196,15 @@ public function getCacheDriver()
 
             case 'redis':
                 $redis = new \Redis();
-                $redis->connect($this->config->get('system.cache.redis.server', 'localhost'),
+                $socket = $this->config->get('system.cache.redis.socket');
```
w00fz: Could you change this to default to `false`:

```php
$socket = $this->config->get('system.cache.redis.socket', false);
```

system/src/Grav/Common/Cache.php
```diff
@@ -196,8 +196,15 @@ public function getCacheDriver()
...
+
+                if (isset($socket)) {
```
w00fz: Then you can just check for `if ($socket) {`

w00fz:
Actually I changed my mind, in the blueprints it should go before server and port, since in your code unix socket takes precedence over server:port.

baltendo:
Thank you for your review! I added the config option to the blueprint, but I don't know the blueprint stuff so I copied it from the redis server setting. Especially the 'label' and 'help' are not clear to me what they are doing. I didn't specify a placeholder. Please let me know if something is missing.

w00fz:
This looks good to me now. Those constants in help and label refer to the admin plugin language and are used for translation. Unfortunately nothing that can be added in here for that, has to be done in the admin plugin repo on the `en.yaml` file only (we use crowdin for translations).
It's something I can take care of unless you want to try.

Thanks!

baltendo:
Thanks again for the review. I am closing this pull request. It would be great if you could add the stuff to the admin plugin.

baltendo:
Thanks again for the review and it would be great if you could update the admin plugin. As I am quite new to github and contributing to other projects, how does it go on now? Are you merging the commits?

w00fz:
We just need to merge your PR with grav. I didn't do it yet as I wanted to have the other devs also take a look at it and give the ok. 

/cc @rhukster @flaviocopes

rhukster:
Well. From the way I read it:

1. `redis.socket` is always defined in system.yaml
2. Because it's always define, `redis.socket` will always take precedence no matter if you have a `redis.server` defined in your user config. You would have to explicitly set `redis.socket` to false for it to pick up `redis.server`

This doesn't look backwards compatible to me.

w00fz:
Yah the `redis.socket` is new

rhukster:
You know i meant `cache.redis.socket: false` right? :)

baltendo:
So I added a placeholder in the blueprint and added the requested part in the system/config/system.yaml with a comment.

rhukster:
Looks good, thanks for you contribution!

baltendo:
Thanks!

PR #1060, opened 2016-09-23
Adds a check for the header 'HTTP_X_FORWARDED_PROTO'

Perlkonig:
Adds a check for the header 'HTTP_X_FORWARDED_PROTO' to discover whether the scheme should be 'https'. This is necessary in some shared hosting or load balancing situations where the SSL negotiation happens upstream.

### DIFF
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -125,6 +125,8 @@ class Uri
         // set the base
         if (isset($_SERVER['HTTPS'])) {
             $scheme = (strtolower(@$_SERVER['HTTPS']) == 'on') ? 'https://' : 'http://';
+        } elseif ( (isset($_SERVER['HTTP_X_FORWARDED_PROTO'])) && (strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) === 'https') ) {
+            $scheme = 'https://';
         } else {
             $scheme = 'http://';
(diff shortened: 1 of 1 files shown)


w00fz:
Hmmm, I'm not too sure about this. Looks like `HTTP_X_FORWARDED_PROTO` can be forged by clients and wouldn't make it very secure on our end.

I've seen people suggesting to handle this by configuring their server as:

```apache
SetEnvIf HTTP_X_FORWARDED_PROTO "^https$" HTTPS=on
SetEnvIf HTTP_X_FORWARDED_PROTO "^http$" HTTPS=off
```

If you have other official resources to point at and where we can look at that would help us establish whether it is ok to add this or not.

Perlkonig:
I didn't come across that .htaccess solution in my searching, so I'm exploring that. Turns out `SetEnvIf` isn't working in this hosting environment, but I'm looking at rewrite solutions (which scare me because there's a lot of rewriting going on). If I find a solution, I'll submit a pull request for the .htaccess file, perhaps. Nothing is ever simple, it seems! :)

Perlkonig:
So the following appears to be working. I'll submit a new pull request for the .htaccess file instead. Thanks for your patience.

```
RewriteCond %{HTTP:X-Forwarded-Proto} https
RewriteRule .* - [E=HTTPS:on]
```

PR #1062, opened 2016-09-23
ParsedownGravTrait `addBlockType` and `addInlineType` enhancements

Sommerregen:
This PR extends the API for Grav plugins to access the Parsedown parser directly. First it avoids an exception, when an index is set, but no block type `$type` exists in the array. And second it allows to register Markdown extensions, which don't have a marker (The built-in `code` block is such an example, where indentation yields formatting the paragraph as code).


rhukster:
Looks good!

PR #1083, opened 2016-09-30
Proposal to add 'wordwrap' to list of built-in twig filters

Perlkonig:
I'm proposing to include this because it's a natural extension of the already existing `truncate` filters.
- Fixed two typos
- Added backwards-compatible parameters to `safeTruncate` function.
- Wrote a `wordwrap` filter that leverages `safeTruncate` to wrap a long string at a given length in a word-safe manner.

Thanks!

### DIFF
--- a/system/src/Grav/Common/Twig/TwigExtension.php
+++ b/system/src/Grav/Common/Twig/TwigExtension.php
@@ -79,7 +79,7 @@ class TwigExtension extends \Twig_Extension
             new \Twig_SimpleFilter('regex_replace', [$this, 'regexReplace']),
             new \Twig_SimpleFilter('safe_email', [$this, 'safeEmailFilter']),
             new \Twig_SimpleFilter('safe_truncate', ['\Grav\Common\Utils', 'safeTruncate']),
-            new \Twig_SimpleFilter('wordwrap', 'wordwrap'),
+            new \Twig_SimpleFilter('wordwrap', ['\Grav\Common\Utils', 'wordwrap']),
             new \Twig_SimpleFilter('safe_truncate_html', ['\Grav\Common\Utils', 'safeTruncateHTML']),
             new \Twig_SimpleFilter('sort_by_key', [$this, 'sortByKeyFilter']),
             new \Twig_SimpleFilter('starts_with', [$this, 'startsWithFilter']),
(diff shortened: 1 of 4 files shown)


Perlkonig:
Will do. About to leave for vacation, so it will be a week or two.

rhukster:
@Perlkonig any chance you could add tests per previous request??

Perlkonig:
Yup. Sorry for the delay.

Perlkonig:
Actually, I'm a moron. I did not realize PHP had a built-in `wordwrap` function. What a crazy language! How it has survived this long I will never know. I changed the PR to simply use the built-in function. I added a test, but I don't know how to invoke it. Let me know what else you need.

Perlkonig:
And of course, travis died ☹️ Not sure what I did to cause that.

Perlkonig:
Yeah, the problem is with the test code, but I don't see how to fix it.

flaviocopes:
The test fails because of calling a method you removed

screen shot 2016-10-24 at 12 15 46

mahagr:
I don't think that `wordwrap()` in PHP is UTF8 safe. Try wordwrap with `ääääää...` to test if it wraps to the half of the size. In the worst case word will be wrapped to 1/4 of the wanted size.

Perlkonig:
Well that's no good! Will triple check. Are you saying that the truncate functions in Utils *are* UTF8 safe? If so, I guess I'll revert to a hand written version. PHP never ceases to amaze me.

Perlkonig:
Thanks, Flavio. "wordwrap" was still registered as an extension, so I assumed it would still work. I'm not familiar with this testing framework. Thanks for fixing.

Perlkonig:
Sure enough, not UTF safe. I'll test the `truncate` functions and find a solution later today.

mahagr:
I have no idea (been while since I've have looked at them) if the existing functions are UTF8 safe, but if they are not, someone will likely report a bug sooner or later to fix them. Its not a big issue on most languages that use latin character set with accents here and there, but for languages like Russian, Greek, Hebrew, Arabian, Chinese etc its going to be major issue.

Perlkonig:
The truncate functions do indeed use the `mb_*` functions, so they should be safe. I'll go back to the hand-coded version and add some accents and cyrillic tests.

Perlkonig:
Recoded `wordwrap`. I can't use `truncate` because it forces a break of long words at the end of a string. The `wordwrap` function shouldn't try to hyphenate because hyphenation rules are byzantine. The multibyte tests aren't passing though. Trying to figure that out.

flaviocopes:
What's wrong with Twig's `wordwrap` http://twig.sensiolabs.org/doc/extensions/text.html#wrapping-words? you wrote about it in https://github.com/Perlkonig/grav-plugin-twig-extensions/blob/master/README.md#omitted-modules

Perlkonig:
It's not word safe (thus the "not useful" comment).

Perlkonig:
Well that's no good! Will double check.

Perlkonig:
I hate reneging on a PR, but I don't have the expertise to figure out why the multibyte tests are failing. Sorry for the trouble.

Issue #1067, opened 2016-09-24
Feature request: Force JSON output regardless of URL extension


Perlkonig:
I have a plugin that returns a status reports. It should always be JSON (using the `*.json.twig` template). I don't want to force the client to append `.json` to the URL. I just want to programmatically trigger the response as JSON. It's currently not possible to do this.

w00fz:
This is kind of the pattern we follow though. I think it would be somewhat confusing and inconsistent if suddenly you could return different output formats from different extensions. `.json` could now suddenly return HTML, `.html` could now suddenly return RSS/XML. etc.
 
Have you considered forcing a redirect for your html route to the `.json` version of it? This way even if the client reaches the regular route, it will still automatically get redirected to the `json` version of it.

rhukster:
I'm ok with adding support for this @w00fz but it will probably not be something in admin. It's a very advanced feature that is not going to be used often.

Perlkonig:
I'm thinking from a web service perspective. URLs are supposed to be opaque. One URL represents one *resource*, which can have multiple *representations*. The client is supposed to be able to send an `Accept` header that tells the server what *representation* they want (e.g., `text/html`, `application/json`, etc.). The URL represents the *resource* itself and shouldn't change. 

So with the Webmention plugin, when you submit a mention, you can optionally be given a URL to check later for the status of your request. If the client is a person, they may want an HTML response. They'd just check it from their browser (automatically a `text/html` request) and get something formatted and pretty. If the status check is done by a script, it would prefer `application/json`. It signals that in the `Accept` header. It shouldn't have to change the URL to get that output.

As it stands, I am not aware of a method for me to, from a plugin, force a response in a specific MIME type. I can change it's template, but I can't  force Grav to use the `*.json.twig` over the `*.html.twig` from within the plugin. I admit this may not be a common need, but it is conformant with HTTP best practices.

w00fz:
Ok I can see some valid points here and I can see a couple things we could improve this, imo.

I think that, and I am not too sure about this as I need to double check the code, we implicitly assume a direct hit to a route is HTML and that it looks for the `.html.twig` template for it, not there fails or defaults to `default.html.twig`, based on the page type.

[...]

PR #1088, opened 2016-10-01
Proof of concept. Still can't throw arbitrary error...

Perlkonig:
DO NOT MERGE! This is proof of concept. The problem is I still can't figure out how to throw a specific error, even from the core code. Any help appreciated.
- Adds a new object to Grav\Common\Utils called [AcceptHeader](https://github.com/adoy/Accept-Header-Parser). License is embedded in the file.
- Significantly alters `Grav\Common\Twig\Twig::processSite()` as follows:
  - If a format is set either by page header or extension, it is honoured—no failsafe.
  - Otherwise, the `Accept` header is analyzed and processed in preference order.
    - The mime type `*/*` is recognized as matching `html`. This should be failsafe enough for modern user agents.
  - If a matching template is found somewhere in the search path, great. Move on.
  - Otherwise, throw an error (ideally a proper `406 Not Acceptable`).
- Right now it only supports `html`, `json`, `yaml`, and `csv` templates. Not sure if there are more that we should add.

My tests so far show this approach does work. The only think it's not doing yet is throwing the correct error. Even though I get an error message, the response code is still `200 OK`, which is not in fact OK.

Thanks for putting up with my recent code binges. I'm not a PHP native. Hopefully my coding style doesn't offend you 😃 

Related to issue #1067


Perlkonig:
Of course I blew something up 😧 I'll see if I can figure out what the Travis reports mean.

w00fz:
I think a lot of that code could be simplified if you use the `mime` method Grav offers ([see here](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Grav.php#L218-L247))

w00fz:
Also for security reasons we should prevent `yaml` requests to go through. We don't want this to be exploited and potentially get access to config files, data files, password files and the likes.

Perlkonig:
How so? The new code happens when `$format` isn't specified. The `mime` method seems to go from extension to mime type, but what I need to do is go from mime type to extension.

system/src/Grav/Common/Utils/AcceptHeader.php
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * This library is free software; you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation; either version 2 of
+ * the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU Lesser General Public License for more details.
+ */
```
w00fz: I don't think we can accept this if the license is LGPL. Grav is under MIT license.

Perlkonig: Fair enough. If that's not OK, I'll code something from scratch.

Perlkonig:
Does .htaccess not already block direct requests for YAML files? Again, the content negotiation would only happen if there was no file extension specified in the URL. We already have the [Page as Data](https://github.com/btopro/page-as-data) plugin, for example.

Perlkonig:
Sure. I could look at moving that logic over there. I'll do that in the next commit.

Perlkonig:
I just realized I failed to set the `Content-Type` header, too. Next commit.

Perlkonig:
Moved that logic into `Grav.php` and set the `templateFormat` so the correct `Content-Type` header is sent.

system/src/Grav/Common/Utils/AcceptHeader.php
Perlkonig: Latest commit removed this code. I wrote my own parser/sorter from scratch without looking 😄

system/src/Grav/Common/Grav.php
```diff
@@ -247,6 +247,31 @@ public function mime($format)
...
+                return 'atom';
+            case 'application/rss+xml':
+                return 'rss';
+            case 'application/xml':
+                return 'xml';
+        }
+        return null;
+    }
+
+    /**
```
Perlkonig: Done.

flaviocopes:
To throw a specific error, use `throw new \RuntimeException('Page Not Found', 404);` for example.

Perlkonig:
That doesn't work. If you look at the current commit, if no output is generated, I do indeed throw a `RuntimeException` with a 406 code, but all I get is an error message dump onto a blank page with a response code of `200 OK`.

w00fz:
I might be wrong but it doesn't look like the `q` parameter for priority is doing anything. It gets set to `1.0` but on multi `Accept` headers it's always `1.0`. I see the logic in the sorter for taking `q` into account but I fail to see where this is ever going to be set differently than `1.0`.

Perlkonig:
[The spec explains.](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) The agent provides q values. The code assigns a default q value of 1.0 if none is provided. Test it with the following header, for example:

```
application/json;q=0.5, text/html;q=0.75, application/xml
```

The correct order would be

- `application/xml`
- `text/html`
- `application/json`

w00fz:
I see, I missed the part where you assign the original `q` back with `floatval` if it was passed in.

w00fz:
Hey @Perlkonig I changed a bit your code, just to simplify it. I also removed some extra stuff you had that I don't think we need. The `culled` is just extra checks and loops we don't really need and doesn't really cause any issue. The sort will also take care of it grouping them at the same level. At the end of the day, you get what you need.

I removed extra details in your result array, now it's flat and only has 3 entries: `mime`, `priority` (q), `index`. Anything else wasn't used anyway.

I could just push these but I'm going to share them here first so you can see what I'm doing 😉 

```php
<?php
/**
 * Parses the `Accept` header and produces an array of acceptable
 * MIME types in preference order. If $cull is true, then only the
 * first instance of each MIME type is returned (you only care about
 * the types themselves and not the parameters).
 *
 *
 * @return mixed
 */
public static function parseAccept() {
    $result = [];

    // Break at commas
    $header = explode(',', $_SERVER['HTTP_ACCEPT']);
    
    // Look at each entry and isolate paramaters
    $result = array_map(function($entry) use ($index) {
        list($mime, $priority) = explode(';', $entry . ';q=1.0');
        $priority = explode('=', $priority)[1];
        
        return [
            'mime' => $mime,
            'q' => $priority,
            'index' => $index++
        ];
    }, $header);

    // Sort result
    usort($result, array(self, "sortAccept"));
    
    return $result;
}

/**
 * Sorts the array generated by `parseAccept` into preference order based on the spec.
 *   - q value
 *   - specificity
 *     - use of wildcards
 *     - number of parameters
 *   - order of appearance
 *
 * NOTE: Inequalities are reversed because we want the first element in the resulting
 *       array to be the one with highest priority (descending order).
 *
 * @return mixed
 */
private static function sortAccept($a, $b) {
    if ($a['q'] !== $b['q']) {
        return $a['q'] > $b['q'] ? -1 : 1;
    }

    $aAny = strpos($a['mime'], '*');
    $bAny = strpos($b['mime'], '*');
    
    if ($aAny !== $bAny) {
        return $aAny > $bAny ? -1 : 1;
    }
    
    return $a['index'] > $b['index'] ? -1 : 1;
}
```

As you can see is much simplified but does the same thing.

Perlkonig:
TLDR: Thanks for this! I'm totally fine with pushing this to the branch.

* Thanks for cleaning things up. I'm not a PHP native, so my style is not ideal, I know.
* I did the "cull" thing because if multiple of the same mime type does exist with different parameters, then the actual content negotiation will retry the same template multiple times. Not sure what's cheaper: the cull loop or the failure of the `render()` command. Probably the latter because multiple types shouldn't frequently appear.
* I agree that most of the parameters aren't needed. I just wanted my first pass to be as inclusive as possible. I should note, though, that according to the spec and the way my sort worked, mime types aren't necessarily grouped. See last example in the spec, for instance. But again, Grav doesn't need the other parameters, so we could skip the specificity sort for now.

rhukster:
What's the status of this PR? Is it something that is still being worked on?

Perlkonig:
I don't know if @w00fz has more to add. I submitted it as a proof of concept. I still can't throw a specific HTTP error. If we can get it to a place where you think it will work, you should feel free to merge it, despite the perhaps-over-emphatic title. I just think it would be useful for Grav to be able to respond to a well-formed client request. As it stands, either the user has to add an extension or the page file itself has to have the response type in the front matter.

Perlkonig:
Closing because it's old, and there's obviously no real interest. I'm going to replicate the throwing errors problem and submit that as an issue first. If that gets resolved (or is already resolved), then I'll submit a new PR against the current code base.

Issue #1064, opened 2016-09-24
Error 500 page exposes site details to visitors


gitname:
Even with the **Display errors** option set to `No`... 

![image](https://cloud.githubusercontent.com/assets/7143133/18805339/a98c4fd8-81bf-11e6-9039-4c317466db6f.png)

..., the [Error 500 page](https://github.com/getgrav/grav/blob/master/system/src/Grav/Common/Errors/Resources/layout.html.php) (example shown below) still exposes more information than I would like it to while in Production.

Specifically, it does 2 things I want to change:

1. It has the **known appearance** of a Grav Error 500 page (providing a clue to the visitor that the site uses Grav and is vulnerable to any outstanding vulnerabilities in Grav)
2. It shows the **PHP error information** (providing a clue to the visitor of potential vulnerabilities in the site)

![image](https://cloud.githubusercontent.com/assets/7143133/18805359/66d0821c-81c0-11e6-9548-8285e27b06e2.png)

I would like to be able to customize the page, to disable it completely, or for there to be a Grav option I can use to indicate whether I want it to contain the PHP error information or not.

Perlkonig:
I concur. When "display errors" is set to false, it should return a bare `500` and let the web server handle it. I'm looking through the source code but don't understand Whoops enough to know what the alternatives are. What would happen if you just returned `Handler::Quit` in [this function](https://github.com/getgrav/grav/blob/6c042b48043336203b138e2e76cf2b3a11dbda33/system/src/Grav/Common/Errors/SimplePageHandler.php#L29)? Why does Grav want to handle it in the first place? I'll do a little digging.

Perlkonig:
A bare `Handler::Quit` seems to work. Something to consider, anyway.

rhukster:
Done.

PR #1089, opened 2016-10-02
Generate a bare 500 response when errors:display is false. This commi…

Perlkonig:
…t does not remove the Resources folder at this time, though they could conceivably be deleted if this approach was accepted.

Related to Issue #1064


Perlkonig:
Another option is to leverage the Errors plugin in the `handle()` function, but I didn't have time right this second to test how that might work. I'll try tomorrow. Ideally it would check for a customized 500 page in the Errors plugin and return a bare `Handler::QUIT` otherwise.

Perlkonig:
As far as I can tell, the SimplePageHandler doesn't have access to the surrounding Grav object, so perhaps there is no easy customization possible. But with "error:display" set to false, I think a bare 500 is best anyway.

Perlkonig:
The latest commit cleans up the `Resources` folder and minimizes the `SimplePageHandler.php` file as much as works. My admittedly basic tests seemed to work fine. 

* The principle here is that there is no reason for Grav to waste resources on this. If `system.errors.display` is truly set to false, then don't do anything but emit an empty 500.
* Most (all?) modern web servers already offer a server-level option for customizing error pages.
* I tried numerous methods for leveraging the Errors plugin or some other mechanism, but I couldn't come up with something clean. Again, if `system.errors.display` is false, it's probably best to keep it simple and just return the 500.

Anyway, for your consideration. Thanks for your time!

rhukster:
I think this change would cause a lot of upset people.  The custom error page is not just for 500 errors, it's for anything that Grav would throw an error at.  I think most webserver provided 500 errors are very ugly, and don't provide any information, causing lots of extra detective work.  So i'm really not interested in simply removing that. 

If you can find a way to add toggle an **optional** "let webserver handle" errors, then sure I'll consider that, but not simply removing functionality that is there already.  Changes need to be upgrade-safe and not change fundamental behavior.  They can add, even modify (to a degree) current behavior, but not simply remove it.

Perlkonig:
But I think the point of setting `system.errors.display` to `false` is exactly to *not* "provide any information." The site owner can always set the display to `true` and get a full stack trace.

If you're open a new system setting, then I can do that. Ideally I'd turn `system.errors.display` to a three-option switch, but that's not backwards compatible. It just feels hackish to have a `system.errors.display.no_really` setting :smile:

flaviocopes:
Agree, we cannot remove such functionality, which I consider useful. 
An alternative way would be letting the theme provide a custom error page. Or have the ability to change the strings that provide the error message (which now is hardcoded to english - that way one could remove the error message if desired).

Perlkonig:
I tried finding a way to customize, but apparently the handler needed a standalone HTML page. When the error is called, it means you can't rely on Grav to generate anything. Something blew up. New PR on its way.

PR #1091, opened 2016-10-04
Adding 'system.errors.verbosity' to eventually replace 'system.errors.display'

Perlkonig:
* Introduces three levels of verbosity:
  * 2: Full stack trace
  * 1: Current Grav error page
  * 0: Bare 500 response

* Backwards compatible
  * If `verbosity` isn't set, then it looks at `display`.
  * Otherwise, `verbosity` is taken *instead of* `display`.
  * If accepted, I suggest marking `display` as deprecated for a release or two and documenting the change. 
  * Adopting this approach will not break legacy installs. Not updating your config file results in zero change in Grav behaviour.

References Issue #1064 and closed PR #1089. Thanks!

### DIFF
--- a/system/src/Grav/Common/Errors/BareHandler.php
+++ b/system/src/Grav/Common/Errors/BareHandler.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * @package    Grav.Common.Errors
+ *
+ * @copyright  Copyright (C) 2014 - 2016 RocketTheme, LLC. All rights reserved.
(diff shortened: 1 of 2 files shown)


rhukster:
Thanks!

rhukster:
as you can see i tweaked your PR a little.  basically i removed the new `verbosity` option and reused the existing `display` one. By also changing the values to `1, 0, -1` I was able to maintain backwards compatibility with the old true (1) and false (0) values.

PR #1092, opened 2016-10-04
Added autoplay and hide controls methods

juanparati:
VideoMedium doesn't have a public interface that allows to set some basic video tag attributes like "autoplay" or to remove the "controls" attribute. 

I decided to implement two public methods that can supply the missing functionality for add the attribute "autoplay" and remove the "controls" attribute.

### DIFF
--- a/system/src/Grav/Common/Page/Medium/VideoMedium.php
+++ b/system/src/Grav/Common/Page/Medium/VideoMedium.php
@@ -30,6 +30,33 @@ class VideoMedium extends Medium
         ];
     }
 
+
+    /**
+     * Set video autoplay on
+     *
+     * @return $this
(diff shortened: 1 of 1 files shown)


flaviocopes:
Thanks, can you provide an example usage to put in the docs?

juanparati:
An example using Twig:

`video.resize(429,270).autoplay().hideControls().loop().html('','','align-'~page.header.video_align)`

Sommerregen:
@rhukster I think this is a precursor of #1442 and not needed anymore as #1442 has already been merged https://github.com/getgrav/grav/commit/849130621a8929e6d56a0cd104e3f76ab90efb37 .

PR #1125, opened 2016-10-21
Don't truncate HTML if content length is less than summary size

### LINKED ISSUE #1114: Utils::truncateHTML removes whitespace when generating summaries
I have set my **Summary Size** high, so that only pages which have a **Summary Delimiter** have summaries generated on the list page. 

Unfortunately, in this instance when comparing `page.summary` and `page.content`, short articles which span multiple paragraphs which should otherwise have `page.summary == page.content` do not. This can be shown by dumping the variables:

```
{{ page.summary|var_dump }}
{{ page.content|var_dump }}
```

outputs 

```
string(53) "<p>This is a post.</p><p>It spans multiple lines.</p>"
string(54) "<p>This is a post.</p>
<p>It spans multiple lines.</p>"
```

I have developed a workaround by editing `/system/src/Grav/Common/Page/Page.php:summary()` to compare the string length of `$content` vs `$size`, so if content is shorter than the summary cut-off, just return the content as is:

```
		if (strlen($content) < $size) {
	         	         return html_entity_decode($content);
		} else {
	         	         $summary = Utils::truncateHTML($content, $size);
	         	         return html_entity_decode($summary);
		}
```

Is this a bug, or am I just missing something obvious? Is there a more graceful workaround?

adamroe:
Fixes #1114 by comparing size of `$content` against the summary size. If a `$content` is short and doesn't require a summary to be generated, return it's original form.

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -506,7 +506,8 @@ class Page
             $size = 300;
         }
 
-        $summary = ((strlen($content) < $size) ? $content : Utils::truncateHTML($content, $size));
+        $summary = Utils::truncateHTML($content, $size);
+		
         return html_entity_decode($summary);
     }
 
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -506,8 +506,7 @@ class Page
             $size = 300;
         }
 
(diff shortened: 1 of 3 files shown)


adamroe:
Hi @rhukster, good call. I couldn't see anywhere else in the codebase where TruncateHtml was called, but placing the sanity check in the Utils:truncateHtml makes sense going forward. I have updated my pull request.

Issue #1051, opened 2016-09-18
Feature suggestion: full control over image sizes with ImageMedium#derivatives


fredrikekelund:
I would like to suggest that the `ImageMedium#derivatives` method be able to take an array instead of just max, min and step arguments. That array would allow for flexible control over what image widths to resize to.

This is useful because often times you're looking for an incremental increase in image widths - for mobile users you might want to really optimize for performance, but for desktop users, maybe all you need is one or two large versions that won't ever be too small.

I could easily submit a PR for this, but wanted to get a green light first.

rhukster:
Sounds useful.. green light given :)

fredrikekelund:
Cool! :) I'll see if I can submit a PR for this some time in the coming week

PR #1133, opened 2016-10-23
Added ability to pass an array of explicit widths to ImageMedium#derivatives

fredrikekelund:
Allows for more precise control than the min-width, max-width and step parameters. As previously discussed in #1051.

### DIFF
--- a/system/src/Grav/Common/Helpers/Excerpts.php
+++ b/system/src/Grav/Common/Helpers/Excerpts.php
@@ -282,8 +282,15 @@ class Excerpts
 
         // loop through actions for the image and call them
         foreach ($actions as $action) {
-            $medium = call_user_func_array([$medium, $action['method']],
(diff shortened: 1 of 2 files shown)


fredrikekelund:
@rhukster it used to be pretty straightforward in Twig or PHP, but I realized that specifying an array as a function argument wasn't Markdown compatible, so I pushed a commit to fix that. Now, testing it should be a simple as doing this in a page's Markdown:

```md
![](test.jpg?derivatives=[640,1500,2500])
```

rhukster:
Ok thanks, will test this shortly!

rhukster:
Sorry for the delay, I have been uber busy.  Will test this during the upcoming holiday.  Cheers.

rhukster:
I got it tested and looks good! Thanks, merged.

Issue #1169, opened 2016-11-11
Unable to perform case-insensitive ordering of a collection when the order is based on a custom header field.


parcand:
The `$sort_flags` parameter is always set to `SORT_REGULAR ` when ordering a collection based on a custom header field.

This is quite problematic since there are many situations where data is expected to be sorted case-insensitively. 

Adding an optional `sort_flags` parameter to the `order` method of the `Grav\Common\Page\Collection` class could help solve that issue.

flaviocopes:
Ok I've done a bit of research into this, and came up with this PR https://github.com/getgrav/grav/issues/1173

The thing to note is this: If the `intl` PHP extension is loaded, only these flags are available:
https://secure.php.net/manual/en/collator.asort.php as we use PHP's Collator class to sort.

Otherwise, if `intl` is not loaded, you can use the PHP standard sorting flags (https://secure.php.net/manual/en/array.constants.php) and add

```
content:
    ....
    order:
        by: header.title
        sort_flags:
            - SORT_NATURAL
            - SORT_FLAG_CASE
```

to order case insensitive.

parcand:
Wow, that was fast, and your solution is great. Thanks a lot!

rhukster:
Also we are going to improve collections in Grav 2.0

PR #1173, opened 2016-11-15
Allow to override sorting flags for page header-based or default ordering.

flaviocopes:
If the `intl` PHP extension is loaded, only these flags are available:
https://secure.php.net/manual/en/collator.asort.php. Otherwise, you can
use the PHP standard sorting flags
(https://secure.php.net/manual/en/array.constants.php)
[#1169](https://github.com/getgrav/grav/issues/1169)


parcand:
This is great! Thanks a lot! 👍

PR #1174, opened 2016-11-17
Add 2 new language values for French

amtur8:
set the name and native name for these two.
(fr-FR and fr-CA)


MATsxm:
+1 :+1:

MATsxm:
@flaviocopes @rhukster 
Following this, it would be great to "target language" to fr-CA for the Core and Admin plg.
Thanks

PR #1178, opened 2016-11-18
Fix Page Collections problem with @page.modular

haagvivi:
Loading the page children before call modular on the Collection.

In order to view the complete example issue you can refer to [https://getgrav.org/forum#!/general:cant-get-modular-collectio](https://getgrav.org/forum#!/general:cant-get-modular-collectio)

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -507,7 +507,7 @@ class Page
         }
 
         $summary = Utils::truncateHTML($content, $size);
-		
+
         return html_entity_decode($summary);
     }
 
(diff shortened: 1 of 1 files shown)


rhukster:
Thanks I'll look at this soon and review.

mahagr:
There are two kinds of children: modular and non-modular. After the fix both will be returned, which is not correct.

flaviocopes:
I tried adding a non-modular child, it's not picked up, as it's calling `$results->modular();`. Not sure if I "got" the problem you notice

haagvivi:
Yes, the modular function juste keep the modular pages.

PR #1180, opened 2016-11-20
Add Chinese language.

lndj:
Add Chinese language & the test.


flaviocopes:
Thanks @lndj, for translations of core and Admin we use Crowdin, sign up here https://crowdin.com/project/grav-core and submit your translation there, thanks! I will merge the test separately after it's done.

flaviocopes:
Closed for inactivity

PR #1194, opened 2016-12-07
Added warning to `update` command when new Grav version is available

Perlkonig:
After [coming up in the forum](https://getgrav.org/forum#!/general:error-after-updating-plugin), I've added a warning to the `gpm/update` command if it notices that a new Grav version is available. My understanding is that best practice is to `bin/selfupgrade` *before* `bin/update`.

### DIFF
--- a/system/src/Grav/Console/Gpm/UpdateCommand.php
+++ b/system/src/Grav/Console/Gpm/UpdateCommand.php
@@ -11,6 +11,7 @@ namespace Grav\Console\Gpm;
 use Grav\Common\GPM\GPM;
 use Grav\Common\GPM\Installer;
 use Grav\Console\ConsoleCommand;
+use Grav\Common\GPM\Upgrader;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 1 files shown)


rhukster:
The only issue I have with this, is there is a minor chance that the dependencies are such that you shouldn't update Grav first.  I can't believe this would really ever be the case, but it 'might', and your warning is kinda fatal if you decline. 

maybe it should be a strong warning, but let you continue if you ignore?

Perlkonig:
It does let you continue if you want. It's up to you. You can answer "Y" to the question.

I could clarify "If you continue without *also* updating Grav, blah..." The real issue seems to be updating plugins that require the newer version of Grav, but the developer wasn't aware of that and didn't add a dependency. Also, since plugins update more frequently than Grav itself, people may be in a habit of updating plugins without even realizing that an updated version of Grav is available.

rhukster:
Ok, cool, i misread it then.  Yeah, it's good practice to update Grav first, so the warning is good.

rhukster:
BTW in future, please add a CHANGELOG.md entry too as part of the PR.. saves me doing it as a separate.

Perlkonig:
Will do that in the future. Sorry!

PR #1208, opened 2016-12-12
fixed array handling

ChrisGitIt:
if the input array looks like this: 
array('valA','valB') (not a keyed array)

The return result before my change was:
array(array('valA'),array('valB'))

The return result after my change is:
array('valA','valB')

(here is an example that got bug me: https://getgrav.org/forum#!/theme-development:no-select-multiple-in-backe)


rhukster:
Thanks!!

PR #1209, opened 2016-12-13
TwigExtension - detect (in)human agent

gigago:
Added function ishuman() to TwigExtension.


flaviocopes:
The change in system/src/Grav/Common/Taxonomy.php is not related I think? 

Also, please update your branch as it seems there are some conflicts with the current develop.

gigago:
Taxonomy is unrelated, don't know why it shows up here. I tried to update the branch, which appears to have resolved the conflicts.

rhukster:
Hmm. you appear to have pushed files with merge conflicts.  Take a look at the "Files changed" tab.

gigago:
Seems to be ok now.

flaviocopes:
Usage:

{% if ishuman() %}
human
{% else %}
not human
{% endif %}

Avoid instantiating Browser as it's already added to the Grav object:

```php
    public function isHuman()
    {
        return $this->grav['browser']->isHuman();
    }
```

But, since the same functionality can be achieved using 

```twig
{% if browser.ishuman %}
human
{% else %}
not human
{% endif %}
```

I'm a bit unsure if it's worth this shortcut.

gigago:
Sounds reasonable to me. Frankly, I didn't realize it could be solved this way. Since this is functionally equivalent and trivial to code, I'll take no for an answer.

rhukster:
Yah I think that using `browser.isHuman` is already simple enough.  Probably just need better documentation.

I've added this line item to: https://github.com/getgrav/grav-learn/issues/356

PR #1218, opened 2016-12-16
Added router to support built-in PHP web server

ftrrtf:
I was surprised when I saw that there is no way to run grav on the built-in server, which is ideal for development, 'cos grav doesn't need a database, etc. So I'm a little researched this issue and found a solution. 
Used for the past several months for me, no any problems noticed, everything works as well.

How to run:
```
php -S localhost:8000 router.php
```

Related issues: 
https://github.com/getgrav/grav-plugin-admin/issues/2
https://github.com/getgrav/grav-plugin-admin/issues/3

### DIFF
--- a/router.php
+++ b/router.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * @package    Grav.Core
+ *
+ * @copyright  Copyright (C) 2014 - 2016 RocketTheme, LLC. All rights reserved.
+ * @license    MIT License; see LICENSE file for details.
+ */
+
+if (PHP_SAPI !== 'cli-server') {
+    exit('This script cannot be run from browser. Run it from a CLI.');
+}
+
+if (is_file($_SERVER['DOCUMENT_ROOT'] . DIRECTORY_SEPARATOR . $_SERVER['SCRIPT_NAME'])) {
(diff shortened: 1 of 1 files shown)


flaviocopes:
The problem is not that Grav does not run on the built-in webserver. It does, and it works fine.

But Admin, and some other plugins, do not because of https://github.com/getgrav/grav-plugin-admin/issues/3#issuecomment-128079164.

We therefore added a check inside Admin to prevent it to run on the built-in webserver. 

BUT your solution makes it work, very nice! Just not sure if / how to handle this, as people do not always read the documentation before firing the built-in webserver, so we'll get those problems back when they run `php -S localhost:8000`. Maybe we could have this documented somewhere, we point out to this in the error we give, and we add a setting to unlock the built-in webserver in the admin config yaml.

rhukster:
Thanks for this PR @ftrrtf, i've made a few minor modifications to ensure this can't be run without the router.  Please check my PR: #1219 

If it's all good for you we can merge that one.

ftrrtf:
@rhukster 👍

rhukster:
Closing this one.. please move discussions to #1219

PR #1219, opened 2016-12-16
Reworked PHP CLI router PR #1218

rhukster:
This is a slight reworking of the original #1218 PR that added a custom `router.php` to enable support via the built-in CLI command. 

This moves the router under `system/` folder and also provides an error if you try to use the built-in webserver without the router option.


flaviocopes:
Thanks @ftrrtf 👍

Issue #1253, opened 2017-01-16
Always add an alt attribute to images for valid markup


indigoxela:
Hi,
an _alt_ attribute is required for valid image html markup.

It would be cool, if public function parsedownElement would always add it, even if not set by users in markdown.

Instead of:
`<img src="some.jpg" />` (invalid)
insert
`<img alt="" src="some.jpg" />` (valid)

To me it looks like a tiny change in system/src/Grav/Common/Page/Medium/Medium.php:
```
236a237,239
>             else {
>                 $attributes['alt'] = '';
>             }
```

This also outputs an empty alt attribute for any images handled with methods like cropZoom.
Right now alt attributes are lost for those images, even if set by users in markdown for the original image (in a blog article...).

Or is there a good reason to skip alt attributes if empty and drop them if image is changed?

flaviocopes:
Adding an empty alt tag is not a bad idea I think, while an empty alt tag is not really useful per se, it helps passing HTML validation for those that are interested in it.

Alt attribues _are_ preserved when processing images:

```
![ALT test](test.jpg?cropZoom=100,100))
```

⬇️ 

```
<img alt="ALT test" src="/grav-develop/images/9/3/4/5/9/93459bbe7f7babd39d7f46e164bc470825e32cce-test.jpeg">
```

Maybe you are testing with cache enabled and don't see your changes reflected.

indigoxela:
> Alt attribues are preserved when processing images:

I must be more precise:
Yes, if adding images directly in markdown, alt attributes are preserved.

But if for instance for an overview page of a collection, all the child page images are processed in a template loop, all alt attributes are lost.

As an example: I have a template "blog_item.html.twig".
In a loop I get the first image of all blog item pages to render them as small linked images:
```
    <a href="{{ page.url }}">
    {{ page.media.images|first.cropZoom(200, 200).html }}
    </a>
```
In the resulting page listing, all alt attributes are gone.

Btw.: alt attributes are also gone in search result pages (plugin simplesearch).

flaviocopes:
Ok in this case you're iterating over `page.media` which is an array of the media included in the page. Is the alt text provided in an image meta file? https://learn.getgrav.org/content/media#metafiles or you just defined it inside the page?

indigoxela:
> Is the alt text provided in an image meta file?

The alt text is not provided in a meta file but directly in markdown:
`![the image alt](image.jpg)`

By the way - many thanks for having a look.

[...]

PR #1262, opened 2017-01-19
Provided empty alt if '' is supplied

A----:
See issue #1253 .

`![] (...)` now outputs ``
Same for `image.html('', '')`

`image.html('')` would output ``, same as before. At least it's now possible to choose if alt attribute should appear.

### DIFF
--- a/system/src/Grav/Common/Page/Medium/Medium.php
+++ b/system/src/Grav/Common/Page/Medium/Medium.php
@@ -229,7 +229,7 @@ class Medium extends Data implements RenderableInterface
         }
 
         if (empty($attributes['alt'])) {
-            if (!empty($alt)) {
+            if (!empty($alt) || $alt === '') {
                 $attributes['alt'] = $alt;
             } elseif (!empty($this->items['alt'])) {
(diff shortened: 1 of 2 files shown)


rhukster:
This is failing checks because we have some tests in `tests/unit/Grav/Common/Markdown/ParsedownTest.php` that tests expected in and out for image tags.

These will need to be updated to provide expected output.

A----:
Better? :)
I had some issues running all tests locally, my environment wasn't properly set I guess. I had to rely on travis for some, hence the three commits.

rhukster:
Thanks for the tests! all merged now. FYIW, to get test running locally you should run from a base Grav install.  make sure you have the development vendor libs:

```
$ bin/composer.phar install
```

Then run the tests:

```
$ vendor/bin/codecept run
```

A----:
That's what I did, but I had a some failed expectations. Filenames were not generated correctly, trailing slashes in links, …
This has to be my enviromnent though, and I didn't want to spend too much time on such a small fix.

Anyway, thanks for the merge.

PR #1275, opened 2017-01-27
Don’t set default param_sep, use system constant

rhukster:
Basically this removes the default value from the `system.yaml` file and sets it based on the system `PARAM_SEPARATOR` constant if the value is not already set/overridden in user `system.yaml`.

### DIFF
--- a/system/config/system.yaml
+++ b/system/config/system.yaml
@@ -5,7 +5,7 @@ wrapped_site: false                         # For themes/plugins to know if Grav
 reverse_proxy_setup: false                  # Running in a reverse proxy scenario with different webserver ports than proxy
 force_ssl: false                            # If enabled, Grav forces to be accessed via HTTPS (NOTE: Not an ideal solution)
 custom_base_url: ''                         # Set the base_url manually
-param_sep:                                  # Automatically is selected based on PATH_SEPERATOR, but can be manually chosen. Use ';' for Apache on windows
+param_sep: ':'                              # Will automatically set correctly for windows
 
 languages:
   supported: []                             # List of languages supported. eg: [en, fr, de]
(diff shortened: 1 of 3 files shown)


zzzzBov:
Had some time to test this. It works ok so long as you don't change configuration in admin. If you save any admin changes, the field will save over the empty value with whatever value was selected.

To fix admin, I attempted to update `system/blueprints/config/system.yaml`:

```
param_sep:
    type: select
    size: medium
    label: PLUGIN_ADMIN.PARAMETER_SEPARATOR
    classes: fancy
    help: PLUGIN_ADMIN.PARAMETER_SEPARATOR_HELP
    default: ''
    options:
      '': '(autodetect)'
      ':': ': (for *Nix)'
      ';': '; (for Windows)'
```

However an empty string key didn't appear to work (I've hardly ever used yaml). The spec mentions being able to use `?` for a complex mapping, so I attempted to use:

```
options:
  ? : '(autodetect)'
  ...
```

But that was interpreted as a value of `'?'`. I don't know if that's something where grav should be updating the yaml parser in use.

In any event, I updated to use:

```
options:
  'auto': '(autodetect)'
  ':': ': (for *Nix)'
  ';': '; (for Apache running on Windows)'
```

But even with a value of `'auto'` initially, it appears that the `/admin/config/system` page is pulling from the latest config value set by `PATH_SEPARATOR` rather than the initial value of `'auto'` (which made sense once I thought about it).

So then I attempted to just nuke the entire `param_sep` section from the `system.yaml` blueprint, but even doing that it appears that saving the config page saves the value stored in `system.param_sep`.

At this point I've spent far more time on the issue than I care to. I can't tell you what the right solution is for backwards compatibility for your users, but I can tell you that I'm going to use:

```
$this->container['config']->init();
$this->container['config']->set('system.param_sep', PATH_SEPARATOR);
return $this->container['plugins']->setup();
```

PR #1292, opened 2017-02-04
Fix blueprints slug validation

flaviocopes:


flaviocopes:
Re https://github.com/getgrav/grav-plugin-admin/issues/955

Issue #1293, opened 2017-02-06
Unable to to fetch cache contents when saving on the command line


johnnoel:
I'm building a custom plugin for one of our clients and part of the requirement is caching some information (from an external API) and then using that information on subsequent requests.

Using the CLI functionality, I've created a command that can be run on a schedule; it downloads the data and then saves it to the Grav cache. However, when trying to retrieve that information on a request (so not on the command line), the cache always reported a miss.

After some investigation this seems to be because the [Cache](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Cache.php#L123) class generates a key according to the [Uri](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Uri.php#L66) hostname. On the command line this is short-circuited to 'localhost', while on a normal web request, this is obviously whatever the hostname is.

To reproduce: create a plugin with a CLI Command that saves some data to the Grav cache; then set the plugin to listen on any event and try to fetch the same data.

Expected behaviour: Data cached on the same Grav system would be accessible regardless of environment.

Workaround: The most straightforward way of circumventing this behaviour is to set the global cache key to something known prior to saving or fetching data e.g.

```php
$originalKey = $grav['cache']->getKey();
$grav['cache']->setKey('my-cache-key');
// do caching operations
$grav['cache']->setKey($originalKey);
```

If the Uri class adhered to the `custom_base_url` in the system.yaml configuration file then this would be a good fix while retaining existing functionality. Otherwise making sure that the cache prefix in system.yaml is used verbatim, or perhaps introducing a cache namespace configuration option would, again, solve the problem.

rhukster:
Ok I've looked into this and found a couple of things:

1) A plugin that sets the cache key early (`onPluginsInitialized()`) and then used for both web/cli works.

2) The caveat with the above is that APC/APCu/XCache etc have seperate cache instance for each PHP process (web/cli) so even with the same cache key, they caches are independent between cli and web.  

So, i've attached a test plugin to show how you can set the cache key for both CLI and web.  It includes a simple CLI command:

```
bin/plugin cli-cache-test cache
```

to store a value in cache, then the plugin itself retrieves that and then dumps it via `dump()` command.   This plugin works with latest Grav if you have the cache setting set to `file` as this is a shared cache between both and it works fine.

Here's the plugin zip (just extract into your `user/plugins/` dir to test: 
[...]

PR #1294, opened 2017-02-06
Fixes to allow not just ‘default/file’ cache in CLI

rhukster:
A fix for CLI to better integrate with non-volatile cache drivers: https://github.com/getgrav/grav/issues/1293


rhukster:
thanks! merged!

PR #1357, opened 2017-03-18
Fix class for Config/Site "Default language"-field

OleVik:
`vsmall` is an incorrect class, and it should presumably be `x-small` so that the field renders the same size as  "Delimiter", used in /admin/config/site. This is how it renders currently (on larger resolutions):

image

Creds to Ricardo for pointing it out on Slack.


ricardo118:
I found another two of the same issue
npkmren

For Default ordering it has class `long` which is also incorrect, and Append URL extension has just a blank space, no size defining class

PR #1361, opened 2017-03-19
Ensure Windows-compliant filenames

OleVik:
Adds an array (`$illegal`) of disallowed characters, and does a simple `str_replace` on filename prior to writing the package to /tmp, in InstallCommand->downloadPackage(). Source: [SO](http://stackoverflow.com/a/3380159/603387) and [MSDN](https://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx), tested on Windows 10 x64, PHP v7.0.7 ran via Caddy v0.9.3.

Other considerations could be added to `$illegal`, but at current they are superfluous. Did not instantiate as a Class-property as only downloadPackage() runs `file_put_contents`.


rhukster:
@OleVik I see this fix is only in the GPM CLI command, is this not an issue when installing/updating on windows via admin?

rhukster:
I found it.. updated myself, going to test...

OleVik:
Good point, hadn't tested from Admin.

PR #1377, opened 2017-03-23
Adding the ability to inline CSS and Js code via Asset Manager

OliverO2:
This pull request makes Grav's Asset Manager support '_inline_' as a new loading option for CSS and Js assets. It also tries to simplify things by extracting common code from addCss() and addJs() into a new helper function (though the same could still be done for output and pipeline functions).

[Inlining CSS](https://developers.google.com/speed/docs/insights/OptimizeCSSDelivery) and Js resources can speed up rendering, particularly over mobile networks. It is not always feasible to directly include assets into the template, for example, where Sass-complied CSS is used or where CSS or Js is imported from external sites to load fonts.

Enabling any CSS and Js asset to become inline makes it easy to optimize for speed and to reduce [flashes of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) without having to re-structure or re-process asset files.

A slightly simplified real world use case:

`system.yaml` uses collections to group assets by loading method:
```
assets:
  collections:
    css-inline:
      - 'http://fonts.googleapis.com/css?family=Ubuntu:400|Open+Sans:400,400i,700'
      - 'theme://css-compiled/app.css'
    js-inline:
      - 'https://use.fontawesome.com/<embedcode>.js'
    js-async:
      - 'theme://foundation/dist/assets/js/app.js'
      - 'theme://js/header-display.js'
```

The template inserts these into different asset groups depending on loading requirements (defaulting to _inline_):
```
        {% block stylesheets %}
            {% do assets.addCss('css-inline') %}
            {% do assets.addCss('css-link', {'group': 'head-link'}) %}
        {% endblock %}
        {{ assets.css('head-link') }}
        {{ assets.css('head', {'loading': 'inline'}) }}
        {% block javascripts %}
            {% do assets.addJs('js-inline') %}
            {% do assets.addJs('js-async', {'group': 'head-async'}) %}
        {% endblock %}
        {{ assets.js('head-async', {'loading': 'async'}) }}
        {{ assets.js('head', {'loading': 'inline'}) }}
```
See also: https://github.com/getgrav/grav/pull/754

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -255,28 +255,32 @@ class Assets
     }
 
     /**
-     * Add a CSS asset.
+     * Add an asset to its assembly.
      *
      * It checks for duplicates.
      * You may add more than one asset passing an array as argument.
+     * The third argument may alternatively contain an array of options which take precedence over positional
+     * arguments.
      *
-     * @param  mixed $asset
-     * @param  int   $priority the priority, bigger comes first
-     * @param  bool  $pipeline false if this should not be pipelined
-     * @param null   $group
+     * @param  array   $assembly the array assembling the assets
+     * @param  mixed   $asset
+     * @param  int     $priority the priority, bigger comes first
+     * @param  bool    $pipeline false if this should not be pipelined
+     * @param  string  $loading  how the asset is loaded (async/defer/inline, for CSS: only inline)
+     * @param  string  $group    name of the group
      *
      * @return $this
      */
-    public function addCss($asset, $priority = null, $pipeline = true, $group = null)
+    public function addTo(&$assembly, $asset, $priority = null, $pipeline = true, $loading = null, $group = null)
     {
         if (is_array($asset)) {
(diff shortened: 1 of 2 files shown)


OleVik:
Could elaborate on how this differs from the already existing `AddInlineCSS` and `AddInlineJS`? Does it, like 754, add the ability to fully inline the entirety of the file in a `<style>`- or `<script>`-tag?

OliverO2:
@OleVik Yes, exactly: This PR adds the ability to inline the entirety of the file's (or an entire pipeline's) contents in a `<style>`- or `<script>`-tag. It works for local files as well as remote URLs.

rhukster:
This PR looks promising.  I'm traveling for a week so won't really have chance to thoroughly test this until I get back.

OliverO2:
@rhukster Thanks! Don't worry, take your time. I've included some test cases and I'm already using this on a production site.

rhukster:
Looks good 👍

OliverO2:
Cool, thanks! :smiley:

PR #1383, opened 2017-03-26
CSS pipeline rewrite URL fix for webserver root-based paths

### LINKED ISSUE #1382: CSS URLs from webserver root get prepended with base path in asset pipelining (and break)
I've had a few CSS url scheme values get mangled to an invalid path when being being rewritten for the asset pipeline. So, for example, a header background image that I had in my dev environment, without pipelining, did not render when I pushed up to the server environment (with pipelining and minification). The path in the pipelined CSS is like `url('/user/themes/antimatter/css-compiled/user/themes/antimatter/<restofpath>')`. I am using SCSS if that makes any difference to the problem (don't think so judging from the code).

Here are the changes to make to reproduce the bug, apply them to the develop branch:

1) **user/themes/antimatter/scss/template/_custom.scss:** add these lines:

```css
#body {
  background-image: url('/user/themes/antimatter/images/logo.png');
}
```

2) toggle `css_pipeline` in **user/config/system.yaml** and probably clear the caches

I'm submitting a pull request following that works around this issue.

hughbris:
Fixes #1382

In this function, there are now three `if` clauses in a row without `else` and with the same action in their body. Could perhaps be refactored. For now, I've just added that third, making it look crowded :(

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -1255,18 +1255,8 @@ class Assets
 
             $old_url = $matches[2];
 
-            // ensure link is not rooted to webserver
-            if (strpos($old_url, '/') === 0) {
-                return $matches[0];
-            }
-
-            // ensure this is not a data url
-            if (strpos($old_url, 'data:') === 0) {
-                return $matches[0];
-            }
-
-            // ensure this is not a remote url
-            if ($this->isRemoteLink($old_url)) {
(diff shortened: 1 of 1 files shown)


hughbris:
Github UI confused me as usual - I think my amendment has come across.

Thanks for the tip. Thought I was being consistent with `if` stanza below. I've used startsWith() and also changed the data URL test to use that. Also consolidated all three predicates because it was bugging me. Not sure what your style rules are around long lines etc.

Issue #1200, opened 2016-12-11
Install specific version of a plugin


c33s:
currently it is not possible to install a plugin of a specific version.
```
php bin\gpm install mobile-detect --version 1.0.0
Grav Package Manager version 1.1.8
```
this makes it impossible to use grav in an ci environment. if i develop a grav application locally (with vendors and plugins in `.gitignore`) and commit my changes to the my repo where the ci-server will takeover (gitlab-ci), which has the sourcecode from my repo and then calls the build & deploy commands:

`gitlab-ci.yaml`:
```
  - composer install --no-scripts
  - php bin/grav install
  - php bin/gpm install admin
  - php bin/gpm install devtools
  - php bin/gpm install landio
  - php bin/gpm install cachebuster
  - composer install --no-interaction --no-progress --optimize-autoloader  --ignore-platform-reqs
  - lftp -u username,password hostname -e "set ftp:ssl-allow true; set ssl:verify-certificate no; mirror --delete --reverse ./ targetdirectory"
```

in this case plugins can have a different version on production as on the local development server. depending on when the plugins where installed locally (can be a month ago), it can lead to a major gap between the plugin versions.

it should be possible to define the version i want to install from a plugin:

```
php bin/gpm install admin --version=1.2.4
```

flaviocopes:
Not nice as `--version=`, but there's 

`bin/gpm direct-install` which also allows you to install a local package.

Install a plugin:

`bin/gpm directinstall https://github.com/getgrav/grav-plugin-admin/archive/1.2.5-rc.4.zip`

c33s:
@flaviocopes thanks for your nice workaround, this of course helps very much.

for the future it would be awesome to manage this in the .dependencies file directly

related: #880

akeif:
I'm very much interested in completing this feature request. Dev instance up and running. Let's see what I can do.

akeif:
@rhukster @w00fz Can we discuss the following Slack conversation before I keep working on this issue? I need to know what you guys want to do with this issue. Right now, when the CLI install a package, it download the latest version for each dependencies. If I'm to add a feature to download older versions, I should download the specified versions from the package blueprints.yaml

```
akeif [8:14 PM] 
@rhuk @w00fz About issue #1200 `bin/gpm install` fetch packages dependencies
from `https://getgrav.org/downloads/plugins.json`.
Is there a way to request an older plugin version data? I'm pretty sure it's not possible.
[...]

PR #1384, opened 2017-03-26
[#1200] Clarified `install` and `direct-install`

akeif:
To help user that need to install specific version of a package.
So far it wasn't absolutely clear how to achieve and some
users ended up asking for help or opening issues asking
for the feature.


akeif:
Alternative solution following this: https://github.com/getgrav/grav/issues/1200#issuecomment-289073259

PR #1397, opened 2017-04-01
Fix #1389 related to `bin/gpm direct-install`

### LINKED ISSUE #1389: Direct Install lacks options and cleanup
Thought I'd tag @akeif for this, since he's been very involved in getting the `directinstall` command working. Whilst the command works well, with only minor infidelities I thus far have blamed on Windows and haven't tracked down fully in debug, two points are worthy of immediate consideration: The command does not register common options, and folders are not deleted on failure or success.

For the former, `force`, `all-yes`, and `destination` are not available. Just like the rest of GPM, having them available cuts down on development-time significantly. For the latter, I noticed that regardless of whether the installation halts or succeeds, the temporary folders in `/tmp/` are not deleted - at least none have been thus far.

akeif:
Add --all-yes and --destination options

Fix a problem where temporary package files will not be
cleaned upon command error.

### DIFF
--- a/system/src/Grav/Console/Gpm/DirectInstallCommand.php
+++ b/system/src/Grav/Console/Gpm/DirectInstallCommand.php
@@ -15,6 +15,7 @@ use Grav\Common\GPM\Installer;
 use Grav\Common\GPM\Response;
 use Grav\Console\ConsoleCommand;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 class DirectInstallCommand extends ConsoleCommand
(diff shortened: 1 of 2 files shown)


akeif:
@OleVik This PR is ready to review :tada:

OleVik:
Will check it out tomorrow evening.

rhukster:
Looks good from just looking at it.  Will test it myself later.  

BTW, could you also add your folder/file cleanup to the admin GPM command to keep things consistent? https://github.com/getgrav/grav-plugin-admin/blob/develop/classes/gpm.php#L203

akeif:
Sure @rhukster. I'll tackle this out in a few days.

akeif:
@OleVik One more detail: I didn't add `--force`. I didn't find any cache hits. Check [this comment](https://github.com/getgrav/grav/issues/1389#issuecomment-290777124) for full details.

akeif:
@rhukster Done https://github.com/getgrav/grav-plugin-admin/pull/1047

OleVik:
`all-yes` works well, but I did notice an oddity in temp-folder creation: One temp folder is created for the .zip-package, another for its extracted contents - presumably they belong in the same temporary cache. Though since both are removed in the process it's largely irrelevant. Had no infidelities on Windows either; all native functions for extracting and moving, and subsequently cleaning up, appear to be working as they should.

Looks good to me.

akeif:
@OleVik It's the basic and intended behavior.
* copy .zip in tmp/ and run some checks
* extract and check
* install
* cleanup

PR #1404, opened 2017-04-04
Cache now respects params / query

kboduch:
Kinda mandatory, don't you think?

Now http://localhost:8888/home?o=9 or http://localhost:8888/home/o:9 are cached.

### DIFF
--- a/system/src/Grav/Common/Cache.php
+++ b/system/src/Grav/Common/Cache.php
@@ -120,7 +120,7 @@ class Cache extends Getters
         }
 
         // Cache key allows us to invalidate all cache on configuration changes.
-        $this->key = ($prefix ? $prefix : 'g') . '-' . substr(md5($uri->rootUrl(true) . $this->config->key() . GRAV_VERSION),
+        $this->key = ($prefix ? $prefix : 'g') . '-' . substr(md5($uri->buildUri() . $this->config->key() . GRAV_VERSION),
                 2, 8);
 
         $this->driver_setting = $this->config->get('system.cache.driver');
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -113,7 +113,7 @@ class Uri
      *
      * @return string
      */
-    private function buildUri()
+    public function buildUri()
     {
         $uri = isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : '';
 


rhukster:
The cache key is primarily used to cache content, and the params/queries are generally things that are used in Twig templates.  Twig has its's own caching mechanism that does not make use of the cache key.  That's why you can use params on on pages that have cached content as long as those params are used by plugins/themes etc and not the content itself. I guess that's why this issue has never really come up in the past.

That said, there are certainly scenarios when the content and therefore the cached content, could depend on params/queries, so this is probably a worthwhile change.  I'll check it out to make sure there's no other implications.

kboduch:
Yeah, in my case I use params to display portion of images like so:

	{%- set offset = uri.param('o') is not null ? uri.param('o') : 0 -%}
	{%- set limit = uri.param('l') is not null ? uri.param('l') : 9 -%}

	{%- for p in page.media.images() | sort | slice(offset, limit) -%}
	    {{- p.url -}} <br>
	    {{- p.cropZoom(200,200) -}}
	{%- endfor -%}

	<a href="{{ base_url }}/o:{{ offset + limit }}">Next</a>

And without changes I made I get the same page no matter what param is passed.

rhukster:
Is that logic in your page content (`.md`)? or in a twig template (`.html.twig`)?

kboduch:
It's twig template

rhukster:
Ok, i'll check it out.. thanks.

mahagr:
I think it would be better to define this inside the page header and maybe allow plugins to hook into it as well.

Sommerregen:
@rhukster  @mahagr I vote against caching all queries, since it enables an attack vector to fill the Grav cache to an unlimited size. I tested the above code directly in a twig partial template and it works without problems.

rhukster:
Yah, i have been able to replicate the issue, but it's very specific.  Ie, only when you rely on manipulating page content itself with params or queries.  This is pretty rare, and it will cause every combination of param/query to cache as a new page.  On a search page for example, this would be a  nightmare.

So I agree with @mahagr and @Sommerregen it probably should be a page-level option that is not in effect by default.  I'm going to look at how best to achieve this.

Sommerregen:
@rhukster There exists the `cache_enable: false` page header option for such cases, i.e., the content should be manipulated by some queries or params. Maybe that's the best approach?

mahagr:
Header could have a list of query params which affect on the caching.

rhukster:
Yah the more I think about it, page-specific use of params/querystrings can be handled with `never_cache_twig: true` in the page header.  This allows the raw content of the page to be cached, but twig in the page will still be processed.  This should work fine without any extra work.

PR #1411, opened 2017-04-08
Fixed what I think may have been an oversight.

Moonlight63:
getRawContent and setRawContent were structured to get and set the content AFTER the content had already been processed. I ran into this problem while developing a plugin for my own personal use. If this was not an oversight and was in fact the intended behavior, then the wording is very misleading. lol.

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -756,16 +756,6 @@ class Page
         $cache_id = md5('page' . $this->id());
         $cache->save($cache_id, ['content' => $this->content, 'content_meta' => $this->content_meta]);
     }
-
-    /**
-     * Needed by the onPageContentProcessed event to get the raw page content
-     *
-     * @return string   the current raw page content
-     */
-    public function getRawContent()
-    {
-        return $this->raw_content;
-    }
     
     /**
      * Needed by the onPageContentProcessed event to get the page content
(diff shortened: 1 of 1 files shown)


rhukster:
There is a method that returns raw_content, but it's called `rawMarkdown()`.  I don't remember how/why this naming was chosen. In retrospect it does look odd.

Moonlight63:
I didn't see the rawMarkdown function. That is strange naming, lol. In my case this makes even less sense because I am making a plugin that will allow storing more than just markdown in the page. Hence, the raw_content may not actually be markdown at all. I tried searching to see if any core functions were using the get/setRawContent functions and couldn't find any. If a name switch doesn't hurt anything, this might make a lot more sense.

rhukster:
You can't just change the method name of `rawMarkdown()`, it's being used by admin plugin and probably others. I would suggest creating aliases, but you can't just change stuff that exists without considering how it impacts existing plugins and usages by people.

Moonlight63:
My apologies, I didn't check to see if the rawMarkdown function was being used anywhere. As I said, it was only a proposed change if it wouldn't have a large negative impact. I'm not trying to cause an inconvenience for anyone.

rhukster:
Don't forget to check: https://learn.getgrav.org/api/

It's not fully flushed out as our docblock headers need improving, but gives you a high-level view.

PR #1416, opened 2017-04-10
Allow `user/account.yaml` overrides + implemented more robust theme init

Sommerregen:
I stumbled across this minor issue, when trying to extend the base `user/accounts.yaml` configuration. It turned out that there were some edge cases, that `loadLanguages` [Themes.php#L256](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Themes.php#L256) was called, but `theme://` was actually not registered yet. This PR fixes this. It further allows now overriding the `user/account.yaml` file that will be shown in Admin plugin and allows to use the `theme://` scheme inside the actual theme itself via, e.g.

```yaml
# Grav streams
streams:
  schemes:
    blueprints:
      type: ReadOnlyStream
      prefixes:
        '':
          - 'theme://blueprints'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## xx/xx/2017
 
 1. [](#improved)
+    * Allow `user/accounts.yaml` overrides and implemented more robust theme initialization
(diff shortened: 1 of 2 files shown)


rhukster:
Thanks, i was actually aware of this, but hadn't had time to look into it.

PR #1458, opened 2017-05-01
[WIP] Web Server Config URL Rewrite Rules White List

ScottHamper:
Hey All,

When looking into the current `.htaccess` file's URL rewrite rules, I noticed a couple small issues:
- Periods are not escaped when black listing specific files (e.g., `LICENSE.txt`), meaning that other files/folders are perhaps unintentionally blacklisted (e.g., `LICENSEatxt`; likely not really an issue in practice).
- The `.yml` file type extension is not black listed (although `.yaml` is), meaning files such as `/codeception.yml` are accessible via HTTP requests.

This pull request so far simply addresses these two specific issues, however the last issue above has bothered me in a broader way. In general, it is often preferable (from a security point of view) to use a white list instead of a black list when it comes to filtering things (e.g., firewall rules, sanitizing input to prevent XSS, etc). The example of `.yaml` vs `.yml` mentioned above really highlights the limitations/issues with using a black list approach, and I think switching to a primarily white list based approach could be an easy win for improving Grav's default configuration.

In exploring this idea with a website I'm currently working on, I've so far determined that files need to be served from these folders:
- `/` (usually only for a few specific files such as `robots.txt`, `favicon.ico`, etc)
- `/assets`
- `/images`
- `/system/assets`
- `/system/images`
- `/user/pages`
- `/user/plugins`
- `/user/themes`

There may be additional areas that I haven't caught yet, but these are the most obvious sources. With the current URL rewrite rules, it's possible to serve up files in all sorts of irrelevant directories. For example, visitors can get specific version information about installed dependencies through files like `/vendor/composer/installed.json`.

So a simple place to start with a white list might be something along the lines of:

```
# Whitelist specific folders
RewriteCond %{REQUEST_URI} !^/[^/]*$
RewriteCond %{REQUEST_URI} !^/\.well-known/
RewriteCond %{REQUEST_URI} !^/(assets|images)/
RewriteCond %{REQUEST_URI} !^/system/(assets|images)/
RewriteCond %{REQUEST_URI} !^/user/(pages|plugins|themes)/
RewriteRule .+ error [F]
```

This rule would be placed before the existing black list rules, since we still want to avoid serving up specific types of files even under these directories. The clearest example is with plugins - they're a giant mix of code vs assets, and it's likely impossible to know ahead of time what folders under any specific plugin are safe to serve (though perhaps there is a convention for this already established in the documentation? I haven't dived into plugin internals yet).

This specific rule still wouldn't solve issues like serving up `/codeception.yml`, but it's a step in the right direction. The tricky part about a white list is coming up with something that provides value while still working for the overwhelming majority of existing/future website setups (without requiring additional configuration overhead on the website maintainer).

For the sake of completeness (and extra ideas), on the site I'm working on, I've locked things down even further by using the following (so far; still testing/in-development):

```
## Begin - Security
# Whitelist specific folders/files
RewriteCond %{REQUEST_URI} !^/\.well-known/
RewriteCond %{REQUEST_URI} !^/(assets|images)/
RewriteCond %{REQUEST_URI} !^/system/(assets|images)/
RewriteCond %{REQUEST_URI} !^/user/pages/
RewriteCond %{REQUEST_URI} !^/user/(plugins|themes)/.*\.(jpg|jpeg|gif|png|css|css\.map|js|eot|svg|ttf|woff|woff2)$
RewriteCond %{REQUEST_URI} !^/(index\.php|favicon\.ico|favicon\.png|favicon\.gif|robots\.txt|humans\.txt)?$
RewriteRule .* error [F]

# Blacklist specific file types in all subfolders
RewriteRule /.*\.(txt|md|yaml|yml|php|pl|py|cgi|twig|sh|bat)$ error [F]

# Blacklist dot files
RewriteRule (^|/)\.(?!well-known) - [F]
## End - Security
```

Here I've gone ahead and white listed only specific file types under the `user/plugins` and `user/themes` folders (I imagine the list of allowed extensions will need to grow if additional plugins are installed, making this likely non-ideal for a default `.htaccess`). Additionally, only specific files in the root directory are accessible. Beyond that, I apply the current file extensions black list.

Anyways, now that I've written all this up - what do you all think? Is this something you'd be interested in pursuing/merging? I'm happy to provide patches for `.htaccess` and `web.config`, but I am unfamiliar with the other server config formats and do not have a testing environment for them.

Regards,
--- Scott

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -56,9 +56,9 @@ RewriteRule .* index.php [L]
 # Block all direct access for these folders
 RewriteRule ^(\.git|cache|bin|logs|backup|webserver-configs|tests)/(.*) error [F]
 # Block access to specific file types for these system folders
-RewriteRule ^(system|vendor)/(.*)\.(txt|xml|md|html|yaml|php|pl|py|cgi|twig|sh|bat)$ error [F]
+RewriteRule ^(system|vendor)/(.*)\.(txt|xml|md|html|yaml|yml|php|pl|py|cgi|twig|sh|bat)$ error [F]
 # Block access to specific file types for these user folders
-RewriteRule ^(user)/(.*)\.(txt|md|yaml|php|pl|py|cgi|twig|sh|bat)$ error [F]
+RewriteRule ^(user)/(.*)\.(txt|md|yaml|yml|php|pl|py|cgi|twig|sh|bat)$ error [F]
 # Block all direct access to .md files:
 RewriteRule \.md$ error [F]
 # Block all direct access to files and folders beginning with a dot
--- a/.htaccess
+++ b/.htaccess
@@ -54,7 +54,7 @@ RewriteRule .* index.php [L]
 
 ## Begin - Security
 # Block all direct access for these folders
-RewriteRule ^(.git|cache|bin|logs|backup|webserver-configs|tests)/(.*) error [F]
+RewriteRule ^(\.git|cache|bin|logs|backup|webserver-configs|tests)/(.*) error [F]
 # Block access to specific file types for these system folders
 RewriteRule ^(system|vendor)/(.*)\.(txt|xml|md|html|yaml|php|pl|py|cgi|twig|sh|bat)$ error [F]
 # Block access to specific file types for these user folders
@@ -64,7 +64,7 @@ RewriteRule \.md$ error [F]
 # Block all direct access to files and folders beginning with a dot
 RewriteRule (^|/)\.(?!well-known) - [F]
 # Block access to specific files in the root folder
-RewriteRule ^(LICENSE.txt|composer.lock|composer.json|\.htaccess)$ error [F]
+RewriteRule ^(LICENSE\.txt|composer\.lock|composer\.json|\.htaccess)$ error [F]
 ## End - Security
 
 </IfModule>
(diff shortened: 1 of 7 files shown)


flaviocopes:
Nice, I think whitelisting is better and this concern has already been raised in the past. As you say, the default `.htaccess` is the more general one and depending on the plugins installed tweaking could be needed, but usually from those folders you serve the usual suspects like js/css and images, and you already have included them.

rhukster:
I think your edits looks good.  Thanks!

ScottHamper:
Hey @rhukster ,

Always glad to help! However, I hadn't intended the pull request to be merged quite yet - none of my commits actually addressed implementing a white list in the server configs. The initial commits were meant to start a discussion around white listing and to figure out the specific implementation details that would work best for the project (before then going on to implement them). Should I go ahead and make these updates for `.htaccess` and `web.config` using the following generalized list?

```
# Whitelist specific folders
RewriteCond %{REQUEST_URI} !^/[^/]*$
RewriteCond %{REQUEST_URI} !^/\.well-known/
RewriteCond %{REQUEST_URI} !^/(assets|images)/
RewriteCond %{REQUEST_URI} !^/system/(assets|images)/
RewriteCond %{REQUEST_URI} !^/user/(pages|plugins|themes)/
RewriteRule .+ error [F]
```

I'm not sure if a merged PR can be re-opened or if I'll have to create a new PR.

rhukster:
yes, please just create a new PR.

PR #1471, opened 2017-05-12
ETag header should be inside double quotes

j000:
It's small change, but still...
https://tools.ietf.org/html/rfc7232#section-2.3


rhukster:
cheers

PR #1472, opened 2017-05-12
Added Exif + Auto generation of meta.yaml files

rhukster:
This works alone, but works best in combination with latest Grav Admin develop branch.  Basically this is disabled by default to ensure you don't unexpectedly get a bunch of generated meta.yaml files.

To enable, simply toggle the `media.auto_metadata_exif` in `system.yaml`.  Every time media is processed for a page this toggle is checked and if enabled, **and** if the file is a `image/jpeg` mimetype **and** if no existing `.meta.yaml` exists, then it reads the EXIF data, and saves the data to `sample-image.jpg.meta.yaml`.

You can then access these like any other meta data, for example:

```
{{ page.media['sample-image.jpg'].meta.camera }}
```

A sample auto-generated meta file can be seen here:

```
FileSize: 738508
MimeType: image/jpeg
height: 1440
width: 1920
aperture: f/5.6
camera: 'COOLPIX P610'
Orientation: 1
horizontalResolution: 72
verticalResolution: 72
software: 'Photos 2.0'
exposure: 1/250
iso: 360
creationdate: null
focalLength: 179.2
ColorSpace: 1
gps: '-1.2231777777778,35.098036111111'
title: 'Hyena Title'
keywords:
    - kenya
    - narok
    - riftvalley
caption: 'Hyena Description'
```

The Exif object is also loaded via the Dependency Injection Container, so it can be used dynamically by other plugins:

```
$exif_data = Grav::instance()['exif']->reader->read('/path/to/your/image.jpg');
if ($exif_data) {
    var_dump($exif_data->getData());
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## mm/dd/2017
 
 1. [](#new)
+    * Added EXIF support with automatic generation of Page Media metafiles
     * Added `Pages::baseUrl()`, `Pages::homeUrl()` and `Pages::url()` functions
     * Added `Debugger::getCaller()` to figure out where the method was called from
     * Added support for custom output providers like Slim Framework
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 1. [](#new)
     * Added EXIF support with automatic generation of Page Media metafiles
+    * Added Twig function to get EXIF data on any image file
     * Added `Pages::baseUrl()`, `Pages::homeUrl()` and `Pages::url()` functions
     * Added `Debugger::getCaller()` to figure out where the method was called from
     * Added support for custom output providers like Slim Framework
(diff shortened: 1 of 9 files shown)


OleVik:
Works well in tests. For Windows users getting errors, the [PHP docs](http://php.net/manual/en/exif.installation.php) state that - of course - the exif-extension must be enabled. Additionally, the mbstring-extension [must be loaded first](http://php.net/manual/en/function.exif-imagetype.php#100734), like this:

```
extension=php_mbstring.dll
extension=php_exif.dll
```

in your php.ini, which is the default order in most php.ini's. Also, production php.ini's tend to have both enabled.

flaviocopes:
I added a little check in https://github.com/getgrav/grav/commit/6b282d845f40b847ac1f62624f1e5ba4f1b381ab to avoid a fatal error when `composer update` is not run

PR #1480, opened 2017-05-17
<testing> Avoid validation on a YAML field, do not parse YAML

flaviocopes:
YAML parsing causes a fatal error if the field is not a string

### DIFF
--- a/system/src/Grav/Common/Data/Validation.php
+++ b/system/src/Grav/Common/Data/Validation.php
@@ -38,6 +38,11 @@ class Validation
             $field['type'] = 'text';
         }
 
+        // If this is a YAML field, stop validation
+        if (isset($field['yaml']) && $field['yaml'] === true) {
+            return $messages;
+        }
+
         // Get language class.
         $language = Grav::instance()['language'];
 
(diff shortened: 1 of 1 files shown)


8book:
This also happens in Gantry

I'm sorry to put this here, but I'm not very familiar with Github.
This tagged collection error still happens when we use the Gantry

mahagr:
@8book Where do you see this happening?

8book:
Actually I was trying to use the same Yaml I do with the other tags. And it was an error.

taxonomy:
	author: exemplo

After testing a lot I found an example in the GRAV documentation that appeared something like below. And there it was!

items:
    '@taxonomy.author': exemplo

But in the midst of so many tests I also discovered that this way when we switch to normal mode and save it gives error because the field "Items" in the Admin tries to rewrite and makes it go wrong.

mahagr:
To me it looks like your issue isn't the one in this issue and that you're trying to do something in a wrong way.

8book:
It's exceedingly simple what I wanted to do just to make the author collection.
Either way works the second way, so an hour is solved

PR #1481, opened 2017-05-17
Avoid to delete images/ folder after using GPM

dctremblay:
When I installed or updated a Grav plugin, it deleted all the cache, but also the "compiled" images,
makes the website very slow (for a relatively complex website).

My initial reason for this patch is because I installed fastcgi_cache in order to make the site still available while
deploying with Ansible, but the images got cleared when the provisioning run the update GPM command,
then showing "missing picture" errors on the website.

This patch nows permit to use the function like :

```php
Grav\Console\ConsoleTrait::clearCache('all');
Grav\Console\ConsoleTrait::clearCache('cache-only');
```

*Please note that those examples are static, just consider the method parameter.*

And then I've linked it after the update and install classes. So when a plugin got updated/installed,
it does only clear the cache instead of clear the cache plus the compiled pictures, agressively slowing the website.

In hope that this will break nothing in the actual core, because I don't know Grav well enough to deal with all the parameters in my mind.

Thanks !

### DIFF
--- a/system/src/Grav/Console/ConsoleTrait.php
+++ b/system/src/Grav/Console/ConsoleTrait.php
@@ -98,14 +98,12 @@ trait ConsoleTrait
      * @return int
      * @throws \Exception
      */
-    public function clearCache($all = [])
+    public function clearCache($scope = 'all')
     {
-        if ($all) {
-            $all = ['--all' => true];
-        }
+        $scope = ['--'.$scope => true];
 
         $command = new ClearCacheCommand();
-        $input = new ArrayInput($all);
+        $input = new ArrayInput($scope);
         return $command->run($input, $this->output);
     }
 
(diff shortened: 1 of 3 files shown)


rhukster:
I can appreciate this PR after dealing with a large image-based site that continually has to re-process large images even when only minor content is changed.  

However, i think the better approach is to have a configuration option that changes the default 'clear cache' behavior.  I hesitate to change the current behavior by default (ie in system/config/system.yaml), but i think that having the ability to not touch mess with the images/ folder when calling `clear()` would be very useful for some use cases.

I think i will test this though because we used to have issues with Gregwar Image caching not picking up changes it should of detected. Perhaps this has been addressed now and we can assume that it will automatically handle things rather than us manually clearing the whole folder out.

rhukster:
Ok i've implemented this as a system option.

PR #1482, opened 2017-05-18
Remove unpublished pages from the translated languages, move into untranslated list

flaviocopes:
Refs https://github.com/getgrav/grav-plugin-sitemap/issues/43


flaviocopes:
Probably need to add a flag as it changes the current behavior

mahagr:
Hmm.. I think it may be a good idea to add a parameter to get all pages even if they aren't published.. Also what if user has permission to access only some translations?

flaviocopes:
Yes it needs a parameter also because Admin relies on this to manage translations. Permission-related, this only returns the routes to other pages, so once you change the language, you can use the usual permissions checks

mahagr:
Yup, same here.

PR #1508, opened 2017-06-01
fix #1501 path is lost when redirecting to detected language

### LINKED ISSUE #1501: Problem with multisite (using setup_subdirectory.php) and multilanguage
Hi, 

I have been trying to get multilanguage support working with multisite using setup_subdirectory.php. After a lot of frustration, Ive managed to boil it down to this issue :

The script below sets up a multisite environment with a valid english page at http://localhost:8080/test/en/home. However, when I navigate to the root url of the _test_ subsite http://localhost:8080/test, the browser is redirected to http://localhost:8080/en and hence delivers the content of the home page of the **_parent_** site http://localhost:8080/en/home. I would have expected it to redirect to http://localhost:8080/test/en and hence deliver the content of the subsite i.e. http://localhost:8080/test/en/home.

Am I missing some configuration or is this a bug?

Thanks

Andy

P.S. I dont see this issue with setup_subdomain.php, but this is due to the fact that a url such as http://test.domain.com:8080 would be correctly routed to http://test.domain.com:8080/en and hence would correctly deliver the content from the subsite http://test.domain.com:8080/en/home

```bash
wget https://getgrav.org/download/core/grav/1.2.4

unzip 1.2.4
cp setup_subdirectory.php grav/setup.php

cd grav

mkdir -p user/sites/test
cp user/pages/01.home/default.md user/pages/01.home/default.en.md

cat << EOF > user/config/system.yaml

languages:
  supported:
    - en
EOF

cp -r user/plugins user/sites/test/
cp -r user/themes user/sites/test/
cp -r user/pages user/sites/test/
cp -r user/config user/sites/test/

php -S 127.0.0.1:8080 system/router.php &

firefox http://localhost:8080/test
```

the4thamigo-uk:
As far as I can see $path contains the original path from the url. Currently $url=='/' which means that the following redirection occurs : 

http://domain/mypage -> http://domain/en

Instead of :

http://domain/mypage -> http://domain/en/mypage

Im not sure whether the same should be done for the if condition immediately preceding this one though? :o/

### DIFF
--- a/system/src/Grav/Common/Service/PageServiceProvider.php
+++ b/system/src/Grav/Common/Service/PageServiceProvider.php
@@ -73,7 +73,7 @@ class PageServiceProvider implements ServiceProviderInterface
                         $c->redirect($url);
                     }
                     if (!$language->isLanguageInUrl() && $language->isIncludeDefaultLanguage()) {
-                        $c->redirectLangSafe($url);
+                        $c->redirectLangSafe($path);
                     }
                 }
                 // Default route test and redirect


rhukster:
Will take a look, thanks.

mahagr:
This change is super-ceded by my pull request #1511, which tackles the whole issue instead of just fixing this particular bug. The fix kind of works but is really not the correct way to do this.

PR #1509, opened 2017-06-02
Nicetime fix: 0 seconds from now -> just now

ShtHappens796:
This fixes the weird message that appears when the posting time equals the current date, for example if a user submits a comment and checks the time their comment was posted right away.

Will probably require the JUST_NOW variable to be added to all translations.


rhukster:
Thanks.

PR #1554, opened 2017-07-04
Quick Fix Summary issues

ash0080:
1. Calculate the string size without any html tags, so now you can get exactly what you wanted length;
2. Support utf8 2 character widths characters, like Chinese, Japanese;
3. This is a quick dirty mod, some associated functions should be rethinked;

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -497,11 +497,14 @@ class Page
 
         // Set up variables to process summary from page or from custom summary
         if ($this->summary === null) {
-            $content = $this->content();
+            $content = strip_tags($this->content());
             $summary_size = $this->summary_size;
         } else {
-            $content = $this->summary;
-            $summary_size = mb_strlen($this->summary);
+            $content = strip_tags($this->summary);
+            // Use mb_strwidth to deal with the 2 character widths characters
+            // Size should not include the html tags
+            // Or like <p>, <blockquote> will make the rendered width totally different
+            $summary_size = mb_strwidth($content ,'utf-8');
         }
 
         // Return calculated summary based on summary divider's position
@@ -510,7 +513,12 @@ class Page
         if (!in_array($format, ['short', 'long'])) {
             return $content;
         } elseif (($format === 'short') && isset($summary_size)) {
-            return mb_substr($content, 0, $summary_size);
+            // Use mb_strimwidth to slice the string
+            if(mb_strwidth($content, 'utf8')>$summary_size){
+                return mb_strimwidth($content, 0, $summary_size);
+            } else {
+                return $content;
+            }
         }
 
         // Get summary size from site config's file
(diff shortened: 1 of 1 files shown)


ash0080:
Compatible with the old version now

Kasparsu:
`$this->home_route = $config->get('system.home.alias');` this was fine
`$this->home_route      = $config->get('system.home.alias');` this is not
if you do not know exactly what psr2 is then read up on it [http://www.php-fig.org/psr/psr-2/](here)
to ease your work you can install some linters for php aswell as. https://github.com/squizlabs/PHP_CodeSniffer and https://phpmd.org/ are useful. If you are using phpstorm you can easily set up code style and lint automatically from Code->Reformat code. Then you do not have to guess what is wanted from you and let the tools point it out.

system/src/Grav/Common/Page/Page.php
```diff
@@ -106,16 +106,16 @@ public function __construct()
...
+        $this->taxonomy  = [];
+        $this->process   = $config->get('system.pages.process');
         $this->published = true;
     }
 
     /**
      * Initializes the page instance variables based on a file
      *
      * @param  \SplFileInfo $file The file information for the .md file that the page represents
-     * @param  string       $extension
```
Kasparsu: why all the extra space? it was fine before. if you do not know exactly what psr2 is then read up on it [http://www.php-fig.org/psr/psr-2/](here) to ease your work you can install some linters for php aswell as. https://github.com/squizlabs/PHP_CodeSniffer and https://phpmd.org/ are useful. If you are using phpstorm you can easily set up code style and lint automatically from Code->Reformat code. Then you do not have to guess what is wanted from you and let the tools point it out.

system/src/Grav/Common/Page/Page.php
```diff
@@ -215,17 +215,17 @@ public function translatedLanguages($onlyPublished = false)
      */
     public function untranslatedLanguages($includeUnpublished = false)
     {
-        $filename = substr($this->name, 0, -(strlen($this->extension())));
-        $config = Grav::instance()['config'];
-        $languages = $config->get('system.languages.supported', []);
+        $filename              = substr($this->name, 0, -(strlen($this->extension())));
```
Kasparsu: why all the extra space? it was fine before. if you do not know exactly what psr2 is then read up on it [http://www.php-fig.org/psr/psr-2/](here) to ease your work you can install some linters for php aswell as. https://github.com/squizlabs/PHP_CodeSniffer and https://phpmd.org/ are useful. If you are using phpstorm you can easily set up code style and lint automatically from Code->Reformat code. Then you do not have to guess what is wanted from you and let the tools point it out.

ash0080: Actually I use Reformat in PhpStorm，
that‘s the exact PSR-1&PSR-2 style predefined，
but seems not same as original code，
I will clean them manually for the next update

rhukster:
Thanks so much!

ash0080:
Thank you all！

PR #1571, opened 2017-07-17
Fix ordering in linux+intl environments

flaviocopes:


emielmolenaar:
Nice! Was having ordering issues and this PR fixes those issues. :+1:

erdinctogulga:
👍

Issue #1574, opened 2017-07-17
Sortable Pages stop working now [Grav v1.3.0 / Admin v1.5.0]


ash0080:
Can not save the order

ash0080:
Noticed that the folder numeric prefix is still working. means change the order in page editor will rename the folders, but Grav can't show them as before. both on frontend and backend

timrobbings:
I am also seeing this behaviour after updating to Grav v1.3.0/Admin v1.5.0.

I _think_ I can see the problem, see [this PR for my proposed fix](https://github.com/getgrav/grav/pull/1575).

Although admittedly, I have a fairly rudimentary understanding of the codebase so people can let me know if I've missed something!

MattAppleton:
Second this, page ordering was working with rc05, newest Admin seems to break it, have tried clearing cache and going again but not working.

ash0080:
@timrobbings It works, Great!

rhukster:
merged, will release later tonight or tomorrow

ash0080:
@rhukster wait, seems still has problem, the fix doesn't work after touching the page in editor again.

rhukster:
if you can provide a PR with a fix that would be appreciated. I don't have any access right now to a linux system that exhibits this behavior.

timrobbings:
I am having some difficulty reproducing the issue but I am happy to have another look later if you can provide more info on the exact sequence of steps you are taking.

ash0080:
![2017-07-18 4 17 19](https://user-images.githubusercontent.com/2803210/28307219-bd03e10c-6bd4-11e7-839d-9f08f274931c.png)

After dragging


![2017-07-18 4 17 27](https://user-images.githubusercontent.com/2803210/28307223-c0232334-6bd4-11e7-934e-231ba5460df7.png)

After saving

zhedar:
Can confirm this, sadly the fix proposed in #1575 doesn't have any effect. I also tried upgrading to core 1.3.0 alone, thus without updating the admin plugin. Still had the same problem though. 

It seems like the pages are ordered by the date of creation. If I create a new page manually or through the admin plugin, it is still numbered correctly but not sorted after the other pages. This happens both in the admin view and the frontend.
<img width="239" alt="bildschirmfoto 2017-07-18 um 14 51 25" src="https://user-images.githubusercontent.com/6633403/28319556-7b777baa-6bce-11e7-99c7-b19f7bce474e.png">

flaviocopes:
[...]

PR #1575, opened 2017-07-17
Fix default sort order

timrobbings:
[This](https://github.com/getgrav/grav/commit/287a51148a3cdc74d70f6dd9e1de0c3d336b8fae) appears to have broken page ordering (see #1574).

Test-information:

Did a quick test on a personal project, clearing cache and reloading
the page fixed the default sort order of the navbar (i.e. it displayed
according to numeric folder prefixes).


rhukster:
Thanks!

Issue #1577, opened 2017-07-20
Sort order for taxonomy filtered collections not working as expected


timrobbings:
I believe this to be a potential bug, and see this behaviour < Grav 1.3.0.

Assume you have the following structure:
```
/home
/home/featured-work
/portfolio
/portfolio/portfolio-item-1
/portfolio/portfolio-item-2
/portfolio/portfolio-item-3
```

/home is a modular page, and the /home/featured-work page is a module that
displays a number of pages with a given taxonomy from the /portfolio collection.

If each portfolio item is manually sorted (i.e. by folder prefix), then viewing
the /portfolio page displays the collection in the *correct* order.

However, when items from this collection are rendered by a page under a sibling
parent, the sort order is *incorrect*.

I believe this is a bug, but it would be good to confirm.

Looking at the code, it appears to be a straightforward fix, sorting the
collection generated from a given taxonomy.

I have tentatively [submitted a PR for review](https://github.com/getgrav/grav/pull/1578) - which I suggest is worth testing
reasonably thoroughly as I am still getting to grips with the codebase and may
have missed some nasty side effects.

Thanks,

Tim

PR #1578, opened 2017-07-20
Sort taxonomy filtered collections

timrobbings:
This relates to #1577.

Assume you have the following structure:
```
/home
/home/featured-work
/portfolio
/portfolio/portfolio-item-1
/portfolio/portfolio-item-2
/portfolio/portfolio-item-3
```

/home is a modular page, and the /home/featured-work page is a module that
displays a number of pages with a given taxonomy from the /portfolio collection.

If each portfolio item is manually sorted (i.e. by folder prefix), then viewing
the /portfolio page displays the collection in the *correct* order.

However, when items from this collection are rendered by a page under a sibling
parent, the sort order is *incorrect*.

I believe this is a bug, but it would be good to confirm.

Looking at the code, it appears to be a straightforward fix, sorting the
collection generated from a given taxonomy.

**Test-information:**

By cp'ing the files over to an existing Grav site and testing that items displayed
by a module on the home page (from a sibling collection) were displayed as
expected.

Other pages are still sorted correctly (e.g. top-level pages in the navbar), it would
probably be sensible if someone else tests and confirms that this hasn't broken things
for them (it would be much appreciated :smiley: ).

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -2634,7 +2634,8 @@ class Page
                 if (!empty($parts)) {
                     $params = [implode('.', $parts) => $params];
                 }
-                $results = $taxonomy_map->findTaxonomy($params)->published();
+                $order = $this->orderBy();
+                $results = $taxonomy_map->findTaxonomy($params, null, $order)->published();
                 break;
         }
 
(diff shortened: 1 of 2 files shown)


timrobbings:
I might just be being a bit thick here, apologies.

I was (perhaps wrongly) expecting `findTaxonomy` results to be sorted by default.

However, it is possible to chain '.order()' after findTaxonomy, i.e. `findTaxonomy({'tag': 'featured'}).order('default')`.

The previous behaviour was _not_ a bug!

So this PR can probably be abandoned (unless anyone particularly wants `findTaxonomy` results to be sorted by default ?).

rhukster:
I would rather not change this behavior as people that need to sort taxonomy can do so with the chaining you mentioned.

PR #1618, opened 2017-08-23
fix Twig dynamic translation

iusvar:
Here is my PR

A little for my poor english, a bit because I thought the solution was not exhaustive (the presence of the other two arguments: array_support and html_out) I hesitated to do this PR. I've done the tests I find necessary, I think it's okay  :smile:

### DIFF
--- a/system/src/Grav/Common/Language/Language.php
+++ b/system/src/Grav/Common/Language/Language.php
@@ -374,9 +374,7 @@ class Language
     {
         if (is_array($args)) {
             $lookup = array_shift($args);
-            if (is_array($args)) {
-                $languages = array_shift($args);
-            }
+            $languages = array_shift($args);
         } else {
             $lookup = $args;
             $args = [];
--- a/system/src/Grav/Common/Language/Language.php
+++ b/system/src/Grav/Common/Language/Language.php
@@ -374,6 +374,9 @@ class Language
     {
         if (is_array($args)) {
             $lookup = array_shift($args);
+            if (is_array($args)) {
+                $languages = array_shift($args);
+            }
         } else {
             $lookup = $args;
             $args = [];


system/src/Grav/Common/Language/Language.php
```diff
@@ -374,6 +374,9 @@ public function translate($args, array $languages = null, $array_support = false
     {
         if (is_array($args)) {
             $lookup = array_shift($args);
+            if (is_array($args)) {
```
mahagr: `$args` is still an array, no need to check it again.

iusvar: You're right, in fact, just the one line with array_shift

rhukster:
thanks

iusvar:
Thank you all

rhukster:
Actually I had to revert this as it broke sprintf style text replacement in translations.  I've created a new twig filter/function that will allow you to do translations in other languages:

https://github.com/getgrav/grav-learn/commit/95507f1e540ef386ac43089f8ce7d79d64a01a23

iusvar:
The `tl` alternative is fine but I have some doubts about how `func_get_args` is used.

Using [func_get_args](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Twig/TwigExtension.php#L612), all arguments will be available in the only variable [$args](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Language/Language.php#L373) (so `$languages`, `$array_support`, `$html_out` are not used).

So it is in the case of [Translation via Twig](https://github.com/getgrav/grav-learn/blame/develop/pages/02.content/11.multi-language/docs.md#L211), and in the case of [Translations with Variables](https://github.com/getgrav/grav-learn/blame/develop/pages/02.content/11.multi-language/docs.md#L248), only the `$args` variable should be evaluated.

In my opinion, the function [translate](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Language/Language.php#L373) should be more or less written:

```
public function translate($args)
    {
        $numargs = count($args);

        $languages = null;
        $array_support = false;
        $html_out = false;

        if (is_array($args[0])) {
            $lookup = array_shift($args[0]);
        } else {
            $lookup = $args[0];
            $args[0] = [];
        }

        if ($numargs==2){
            $languages = $args[1];
        }
        if ($numargs==3){
            $array_support = $args[2];
        }
        if ($numargs==4){
            $html_out = $args[3];
        }

        if ($this->config->get('system.languages.translations', true)) {
            if ($this->enabled() && $lookup) {
                if (empty($languages)) {
                    if ($this->config->get('system.languages.translations_fallback', true)) {
                        $languages = $this->getFallbackLanguages();
                    } else {
                        $languages = (array)$this->getLanguage();
                    }
                }
            } else {
                $languages = ['en'];
            }

            foreach ((array)$languages as $lang) {
                $translation = $this->getTranslation($lang, $lookup, $array_support);

                if ($translation) {
                    if (count($args[0]) >= 1) {
                        return vsprintf($translation, $args[0]);
                    } else {
                        return $translation;
                    }
                }
            }
        }

        if ($html_out) {
            return '<span class="untranslated">' . $lookup . '</span>';
        } else {
            return $lookup;
        }
        
    }
```

Thank you

PR #1621, opened 2017-08-24
Precompression for (pipelined) assets (nginx::gzip_static support).

Eihrister:
Compressing resources on-the-fly adds CPU-load and latency (wait for the
compression to be done) every time a resource is served. nginx offers
another way of doing things, and it's called gzip_static.

http://nginx.org/en/docs/http/ngx_http_gzip_static_module.html

If you enable gzip_static, nginx will look for $filename.gz and serve
that directly, so no extra CPU-cost or latency is added to your
requests, speeding up the serving of your website.

This feature adds support to directly compress asset files when they are
written to disk by the Grav\Common\Assets class. Using this feature you
can also specify the compression level (default to the maximum of 9).

Since you only have to compress every resource only once, using the
maximum compression level is an ideal default.

### DIFF
--- a/system/blueprints/config/system.yaml
+++ b/system/blueprints/config/system.yaml
@@ -794,6 +794,27 @@ form:
                     validate:
                         type: bool
 
+                assets.precompress:
+                    type: toggle
+                    label: PLUGIN_ADMIN.PRECOMPRESS_ASSET
+                    help: PLUGIN_ADMIN.PRECOMPRESS_ASSET_HELP
+                    highlight: 0
+                    options:
+                        1: PLUGIN_ADMIN.YES
+                        0: PLUGIN_ADMIN.NO
+                    validate:
+                        type: bool
+
+                assets.precompress_level:
+                    type: text
+                    size: x-small
+                    label: PLUGIN_ADMIN.PRECOMPRESS_ASSET_LEVEL
+                    help: PLUGIN_ADMIN.PRECOMPRESS_ASSET_LEVEL_HELP
+                    validate:
+                        type: number
+                        min: 1
+                        max: 9
+
                 assets.enable_asset_timestamp:
                     type: toggle
                     label: PLUGIN_ADMIN.ENABLED_TIMESTAMPS_ON_ASSETS
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -90,6 +90,10 @@ class Assets
     protected $css_no_pipeline = [];
     protected $js_no_pipeline = [];
 
+    // Default values for asset precompression.
+    protected $precompress = false;
+    protected $precompress_level = 9;
+
     /**
      * Assets constructor.
      *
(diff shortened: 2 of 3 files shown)


OleVik:
This seems to be specific for Nginx? Whilst handy, it would be superfluous for all other webservers. Also, have you tested it with other webservers to ensure that the settings causes no issue with them?

Eihrister:
Hi Ole,

For other webservers it will not cause problems, as it is merely creating an additional file.

nginx supports it by default without requiring any extra work, however, support is not limited to nginx. Other webservers can also make use of it, if configured accordingly. Some searching can find some ways to set it up:

Lighttpd: http://redmine.lighttpd.net/issues/611
Apache: https://stackoverflow.com/questions/16883241/how-to-host-static-content-pre-compressed-in-apache

I'm sure it could work in any webserver software as long as you configure it properly. 

If I could've done it in a plug-in, I would've. I have purposefully avoided using "nginx" in the code / descriptions of things, as it is what it is: precompression of files. I understand that it might seem unusual to put this in the core, but to me it didn't really seem to have another place. It's either upon creation of said resources, or doing everything by hand externally out of Grav, not knowing when asset files are created.

Daniël

system/blueprints/config/system.yaml
```diff
@@ -794,6 +794,27 @@ form:
...
+                assets.precompress_level:
+                    type: text
+                    size: x-small
+                    label: PLUGIN_ADMIN.PRECOMPRESS_ASSET_LEVEL
+                    help: PLUGIN_ADMIN.PRECOMPRESS_ASSET_LEVEL_HELP
+                    validate:
+                        type: number
+                        min: 1
+                        max: 9
+
```
rhukster: I see you have provided updates to system.yaml with translations for this field, however, those translations need to be added to the `admin/languages/en.yaml` file as the 'source' language file.

system/src/Grav/Common/Assets.php
```diff
@@ -869,6 +898,49 @@ protected function pipelineJs($group = 'head', $returnURL = true)
...
+            }
+
+            // Make the file modification time identical to that of the source file
+            touch($filename . ".gz", filemtime($filename));
+
+            return true;
+        }
+        return false;
+    }
+
```
rhukster: is it not possible to perform this functionality with the `RocketTheme\Toolbox\File\File` class? We use this class throughout Grav to read/save files.  That way it would be more consistent an not need all the manual low-level PHP file calls.

rhukster (changes requested):

jimblue:
Any update about this PR @Eihrister ?

gstrauss:
lighttpd [mod_deflate](https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModDeflate) can compress and save the compressed result in a cache if `deflate.cache-dir` is set in `lighttpd.conf`.  Then, subsequent requests (with Accept-Encoding) will receive the already-compressed file from the cache.

If you pre-compress the files yourself (gzip, brotli, zstd, ...), you can use lighttpd [mod_magnet](https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModMagnet) to run a small lua script for content negotiation, e.g. [content-negotiation.lua](https://redmine.lighttpd.net/projects/lighttpd/wiki/AbsoLUAtion#Content-Negotiation)

PR #1623, opened 2017-08-25
Gregwar/Image supports gaussianBlur, so should Grav

campino2k:
Since grav uses the GD Adapter of Gregwar/Image and has no option to change it, this should be safe to implement, when the latest Version of the Lib is used.


campino2k:
Maybe the composer.json should then change to require minimum version v2.0.21 of Gregwar/Image

rhukster:
FYI Grav already includes 2.0.21

PR #1634, opened 2017-09-02
Fix setting session cookie twice

j000:
Because setting cookie twice bothers me. Now proper options are applied to first one, so no need to set it again.

`RocketTheme\Toolbox\Session` uses `ini_set()` internally anyway, so this should not be a problem, but it can be changed to full `session_set_cookie_params()` call.


rhukster:
I had noticed this, but thought it was my local setup as I was testing login plugin changes.... Reverted.

PR #1683, opened 2017-10-10
Specify type in inlineJS

paulmassen:
This commit allows users to specify type to inlineJS via a 4th parameter . 

This might be useful for those who would like to output encoded json microdata which require to use 
<script type="application/ld+json"> to be valid. 

Exemple usage: $assets->addInlineJs($outputjson, 100,'', "application/ld+json");
which will output: <script type="application/ld+json">


rhukster:
Please be sure to submit a PR for the documentation to make people aware of this addition.

PR #1687, opened 2017-10-12
Update Assets.php

paulmassen:
This is a small addition to the previous change. 
This allows to have multiple type on inlined js. Without this change, if you have something like
```
{% do assets.addInlineJs('Hello', 100,'', "somespecialtype") %}
{% do assets.addInlineJs('Good Morning', 100,'', "anothertype") %}
{% do assets.addInlineJs('Goodbye', 100,'') %} // regular inlined js
{% do assets.addInlineJs('Bonjour', 100,'') %} // regular inlined js
```

It will be rendered like this:

```
<script type="somespecialtype">
Hello 
Good Morning
Goodbye
Bonjour
</script>
```

With this fix, it should display 
```
<script type="anothertype">
Good Morning
</script>

<script type="somespecialtype">
Hello
</script>

// These two inlined js has no type specified, so we can group them on the same tag
<script>
Bonjour 
Goodbye
</script>
```

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -663,14 +663,18 @@ class Assets
 
         // Render Inline JS
         foreach ($this->inline_js as $inline) {
-            if ($group && $inline['group'] == $group) {
+            if (($group && $inline['group'] == $group) && ($inline['type'] == '')) {
+                // concatenate inlined js if type is empty
                 $inline_js .= $inline['asset'] . "\n";
+            } else {
+                // build a script tag if inline type is set
+                $attributeString = " type=\"" . $inline['type'] . "\""; 
+                $output .= "\n<script" . $attributeString . ">\n" . $inline['asset'] . "\n</script>\n";
             }
         }
 
         if ($inline_js) {
-            $attribute_string = isset($inline) && $inline['type'] ? " type=\"" . $inline['type'] . "\"" : '';
-            $output .= "\n<script" . $attribute_string . ">\n" . $inline_js . "\n</script>\n";
+             $output .= "\n<script>\n" . $inline_js . "\n</script>\n";
         }
 
         return $output;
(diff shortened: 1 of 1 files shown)


mahagr:
It looks like the change changes the order in which the scripts appear. It could be a bad thing, though I cannot find any examples right now.

system/src/Grav/Common/Assets.php
```diff
@@ -663,14 +663,18 @@ public function js($group = 'head', $attributes = [])
 
         // Render Inline JS
         foreach ($this->inline_js as $inline) {
-            if ($group && $inline['group'] == $group) {
+            if (($group && $inline['group'] == $group) && ($inline['type'] == '')) {
```
mahagr: Should be rewritten as:

```php 
if ($group && $inline['group'] === $group && $inline['type'] === '') {
```

mahagr (changes requested):
It looks like the change changes the order in which the scripts appear. It could be a bad thing, though I cannot find any examples right now.

paulmassen:
Requested changes applied

mahagr (approved):

mahagr:
What about the ordering? I'm a bit uneasy as the assets are not added in the order they were added by the code.

paulmassen:
Indeed. 
A solution would be to add type to all inlined js. As per this commit: https://github.com/getgrav/grav/pull/1687/commits/5cd745f74f9b03b640e15a28a0147e39b029601b

The cost is that if several inlined js is added, it will output multiple script tags but it is valid.

What do you think?

mahagr:
Will need to discuss this with @rhukster . Multiple tags has the benefit that broken js doesn't break rest of the js, but on the other hand single takes less bytes. Also the default type should be removed in html5.

paulmassen:
@mahagr I took another look at this, and it seems like the ordering is still correct, it was just wrong in my manually typed example. 
There is still the problem of the defaut text/javascript displayed when no type is set, but I'm not sure I will be able to solve that.

PR #1693, opened 2017-10-13
Update CompiledFile.php

vdakalov:
const `CACHE_DIR` already defined with trailing slash

```php
define('CACHE_DIR', ROOT_DIR . 'cache/');
```


mahagr:
Looks OK, though I would really want to get rid of those CONSTs as they do not work in multisite if you want an individual cache location for each site. :)

PR #1709, opened 2017-10-19
fix nonsense logic

benface:


system/src/Grav/Common/Grav.php
```diff
@@ -148,7 +148,7 @@ public function setLocale()
         // Initialize Locale if set and configured.
         if ($this['language']->enabled() && $this['config']->get('system.languages.override_locale')) {
             $language = $this['language']->getLanguage();
-            setlocale(LC_ALL, count($language < 3) ? ($language . '_' . strtoupper($language)) : $language);
+            setlocale(LC_ALL, strlen($language) < 3 ? ($language . '_' . strtoupper($language)) : $language);
```
rhukster: count is a little faster than strlen - https://svachon.com/blog/benchmark-php-strlen-vs-count/

benface: Hmm but you can't pass a string to `count()` can you? It's for arrays.

rhukster: sure you can, works for arrays and strings.. counts chars in strings.

benface: Are you sure??

```
benface:~ benface$ php -r 'echo count("hello world");'
PHP Warning:  count(): Parameter must be an array or an object that implements Countable in Command line code on line 1
```

rhukster: hmm.. you are right! not sure why i thought that. Let me actually look at this real quick.

rhukster:
Ok tested, and it was just always returning true before and treating things as short lang versions.  Your fixes sorts it.  Thanks.

PR #1712, opened 2017-10-22
Remove the dependency with published() in routable() method of Page

clemdesign:
Hello,

I propose this modification for the class Page.
The initial routable() method return a boolean depends on variable "routable" and result of method published(). I propose to remove the dependance with published().

In my project, I would to use the principle of "not published page" so for adjustment of the content, it is interesting for me to preview the final result. In the current version of Grav, a button "preview" is available (In edition mode) but only reachable when page is published, through the routable() method. So maybe I misunderstood anything on why result of routable() depends of published(), but I think a page can be routable and **not** published.

I propose this solution to resolve my issue, and in case of conflict with another feature unknown by me, tell me please.

Thank you,
Clem

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -1410,7 +1410,7 @@ class Page
     /**
      * Gets and Sets whether or not this Page is routable, ie you can reach it
      * via a URL.
-     * The page must be *routable* and *published*
+     * The page must be *routable*
      *
      * @param  bool $var true if the page is routable
      *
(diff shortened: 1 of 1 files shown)


rhukster:
Changing this would have some far-reaching implications.  the `Page::routable()` method is used in many places and the because the check for published state is implicit in this function it could have unintended consequences. 

I need to think about this to come up with a good solution.. However, I cannot accept this PR as it stands.

clemdesign:
I understand the impact. 

So, I waiting for a good solution.
Thank you.

Issue #1729, opened 2017-10-31
[1.3.8] Taxonomies seem to work with lowercase entries only


campino2k:
I had 1.3.7 or 1.3.6  with a taxonomy named "produktwelt" and things like "Produktwelt1". The taxonomylist plugin worked great with that. Upgraded to 1.3.8 and filtering by the taxonomy works only when the taxonomy entry is completely lowercase.

rhukster:
I see you have already provided PRs for taxonomylist plugin.  That's where this needs to be fixed as the change in 1.3.8+ of Grav was to tackle XSS attacks.

campino2k:
That alone does only fix the active indicator. Besides that: Filtering for terms with Uppercase letters returns nothing

PR #1733, opened 2017-11-01
use lowercase comparison for taxonomy terms

campino2k:
Since the URI params only get returned lowercase in 1.3.8, the comparison for the use of the terms of the taxonomy should search in a lowercase version of the pages taxonomy maps.

This will result in a loss of the case sensitivity of taxonomy terms.

Term1 and term1 will be the same.

fixes: #1729

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -2498,7 +2498,7 @@ class Page
                         foreach ($items as $item) {
                             $item = rawurldecode($item);
                             if (empty($page->taxonomy[$taxonomy]) || !in_array(htmlspecialchars_decode($item,
-                                    ENT_QUOTES), $page->taxonomy[$taxonomy])
+                                    ENT_QUOTES), array_map('strtolower', $page->taxonomy[$taxonomy]))
                             ) {
                                 $collection->remove($page->path());
                             }


campino2k:
Since Taxonomy terms are not part of i18n, this would remove the possibility to have uppercase letters in terms. Is my understanding of this correct?

For using Tags, this may be acceptible, for custom taxonomies this may be not correct in general.

In my case, it makes a huge difference for the german translation. In en locale, that's okay. "hot chocolate". But in german the category is "Trinkschokolade" and "trinkschokolade" would just look very unprofessional. Such things may occur for other people using Grav as well.

campino2k:
Examples of using categories and tags with uppercase letters are there:
* https://andy.rip/filter/visual/tag:Paper%20Cutting
* https://discourse.getgrav.org/t/multiple-taxonomy-collection-with-pagination-limit-and-order-in-twig/4752
* https://underr.space/tag:%D0%93%D1%80%D1%83%D0%B7%D0%B8%D1%8F /even cyrilic tags!
* https://underr.space/tag:%D0%9C%D0%BE%D1%80%D0%B5

rhukster:
Actually!!!!!

I tested a tag of `Franky Smith` on my local test setup and it worked fine....

I was thinking that the URL as being lowercased because of this:

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Uri.php#L294-L297

Turns out that was a user contribution, but no `case_insensitive_urls` is actually set in the default `system.yaml`.  So by default this would be null, unless you saved configuration from admin then it would be `false`. Both of these would skip the lowercasing of the URL.

If however you set this to `true` in your configuration, then your URL is being lowercased, and the  `Franky Smith` filter no longer works.  I think this new setting is not actually correct and is causing the confusion. Do you have this enabled on your site? 

I'm going to look at changing this functionality so:

1) it's more clear
2) it only lowercases the route, and not anything else.

rhukster:
Actually i've reverted the pull completely. I realized that when enabled that broke quite a few things.

campino2k:
`case_insensitive_urls: true` in my installation. That was added 25th of October without having changed this value in admin. Saving the settings occured but not this setting.

campino2k:
In fact this also breaks the latest changes to the taxonomylist-plugin. Setting this to false then doesn't need to set the `| lower` filter.

rhukster:
Yah i've actually removed that case_insensitive_url option completely.. its no longer even used.. will be in next release (setting to false will get you by for now). 

Also i've re-released Taxonomy plugin with that lowercase filter removed.  So pretty soon, things should be back to normal!

campino2k:
That's great to hear and thanks for your patience and understanding.

Issue #1742, opened 2017-11-08
GPM always connects to getgrav server, even when only want to get Local Packages


aguvillalba:
I have disabled the option enable_auto_updates_check:
enable_auto_updates_check: false

The partial template user/plugins/admin/themes/grav/templates/partials/nav.html.twig tries to get the number of installed plugins:
<span class="badge count">{{ admin.plugins|length }}</span> (line 61)

This calls to: 
Grav\\Plugin\\Admin\\Admin::plugins(true) --->
Grav\\Plugin\\Admin\\Admin::gpm() --->
Grav\\Common\\GPM::__construct()

And in that construct you have this code:
```
$this->installed = new Local\Packages();
try {
        $this->repository = new Remote\Packages($refresh, $callback);
        $this->grav = new Remote\GravCore($refresh, $callback);
} catch (\Exception $e) {
}
```
So, independently of whether I want to the local plugins (Admin::plugins(true)) or not, the constructor of GPM will always try to connect to getgrav.org server to get the plugins repository.
IMHO this does not makes sense, since I want the local plugins, there is no need to connect to the external server.
The connection to external server to get the remote packages should be removed from the constructor and place it in a explicit function that is called whenever is needed to grab the remote repository.

I will try to make a pull request to fix this.

Thank you!

rhukster:
Cool, i'll look forward to your PR.

mahagr:
I would use DI to lazy load remote packages only if they are needed.

aguvillalba:
Me too @mahagr, that's my idea as well. There is not dependency injection manager/framework in this project yet. Changing to DI is not straight forward now since it could break other things where the GPM is used (plugins for instance).
I have already asked in the slack channel about using DI manager/framework (I would not use a service locator), but no answer yet.
In the meantime, I am changing the code to do a "pseudo lazy-loading", so the Remote package is only created when it is actually needed.

mahagr:
Grav uses DI, but inits all the classes early, eg. it doesn't lazy load most things.

aguvillalba:
Well, Pimple is a service locator, not a proper dependency injection manager that's why I said that Grav is not using DI manager, and yes, it does not lazy load dependencies, which is also bad.

aguvillalba:
PR added to solve this issue

mahagr:
@aguvillalba You may want to take a look into the issue as your change was reverted because of it broke plugin/theme installation.

PR #1746, opened 2017-11-11
Fixed issue #1742 where the creation of GPM object tried to connect t…

aguvillalba:
…o Remote repositories even if only Local package was needed

Now, it only tries to connect to the outside world when really required and not when just creating a new GPM object.
This could break client classes with the following code:
```
$gpm = new GPM();
$gpm->grav->getVersion(); //or any other grav method
```
The above code must be replaced with:
```
$gpm = new GPM();
$gpm->loadRemoteGrav();
$gpm->grav->getVersion(); //or any other grav method
```

### DIFF
--- a/system/src/Grav/Common/GPM/GPM.php
+++ b/system/src/Grav/Common/GPM/GPM.php
@@ -17,6 +17,9 @@ use Symfony\Component\Yaml\Yaml;
 
 class GPM extends Iterator
 {
+    /** @var  callable */
+    private $callback;
+
     /**
      * Local installed Packages
      * @var Local\Packages
@@ -28,6 +31,9 @@ class GPM extends Iterator
      * @var Remote\Packages
      */
     private $repository;
+
+    /** @var  bool */
+    private $shouldRefresh;
 
     /**
      * @var Remote\GravCore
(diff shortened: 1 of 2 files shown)


system/src/Grav/Common/GPM/GPM.php
```diff
@@ -47,18 +53,33 @@ class GPM extends Iterator
...
-        } catch (\Exception $e) {
-        }
+        $this->shouldRefresh = $refresh;
+        $this->callback = $callback;
+    }
+
+    /**
+     * Loads Remote Grav package available
+     */
+    public function loadRemoteGrav()
```
mahagr: Cannot see this being called from anywhere...?

aguvillalba:
It's used in the admin plugin, in the classes/gpm.php to allow to load the Remote Grav repository.

mahagr:
The issue I'm seeing here is that it is a new function which wasn't there before. So it's not compatible backwards, which needs to be noted.

aguvillalba:
@rhukster I have already made the modifications and PR in the admin plugin to use this new function, so if the PR in the admin plugin is accepted, it will not cause any issue. As you suggested in the admin plugin, checking if the method exists before using it is a good solution for now.
IMHO, the problem that I see here is that the design of the application is not using real dependency injection (only service locator), so the design is not flexible and it is not straight forward to solve things like this elegantly without breaking things. For instance, the GPM class exposes to the outside the Grav instance as a public property, which breaks the encapsulation.
Anyway, as you suggested in the admin plugin, checking for the method before using it allows backwards compatibility.

Issue #1261, opened 2017-01-19
Custom header collection sorting doesn't support date format


ichik:
Hi,
I'm trying to establish custom sorting of collection of pages in a timetable like this:
```
order:
    by: header.eventdate
    dir: asc
```

Unfortunately it doesn't seem to support datetime format that is stored in this custom header (i.e. `eventdate: '03-02-2017 22:18'`).

As a result event from February 3rd comes before January 20th for example. I can resort to using just date for sorting, but date of creation an item is not an event date per se.

rhukster:
I've looked at this and really we have no way to really know what or how to order custom header type sort fields. For this reason, I suggest you change the `header.eventdata` to use ISO format dates: `yyyy-mm-dd`.  This way the sort order will be correct with regular natural sort order that is assumed for all such fields.

You can then reformat this on display via the Twig Date filter: http://twig.sensiolabs.org/doc/2.x/filters/date.html

ichik:
The thing is, though, that the format in the field is defined by this blueprint:
```
header.eventdate:
                      type: datetime
                      label: Event date
```
And the actual data is then set through Admin plugin, is there any way to change the format Admin plugin uses?

w00fz:
This is now fixed. The fix involves a new option for the `datetime` field that allows you to specify manually the format. So in your case you could write in the blueprint:

```yaml
header.eventdate:
  type: datetime
  label: Event date
  format: 'Y-m-d'
```

which will force the format to `2017-01-20` for that particular field.

PR #1763, opened 2017-11-23
Fixed CSS Pipeline failing with Google remote fonts if the file was minified (#1261)

mahagr:
Had to make regex'es ungreedy. I think the same issue is in 

CSS_SOURCEMAP_REGEX
CSS_IMPORT_REGEX
HTML_TAG_REGEX (well, this seems to be used with a single `<script>` or `<style>` tag, but...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
     * Remove support for `config.user`, it was broken and bad practise
     * Make sure that `clean cache` uses valid path [#1745](https://github.com/getgrav/grav/pull/1745)
     * Fixed token creation issue with `Uri` params like `/id:3`
+    * Fixed CSS Pipeline failing with Google remote fonts if the file was minified [#1261](https://github.com/getgrav/grav-plugin-admin/issues/1261)
     
 # v1.3.8
 ## 10/26/2017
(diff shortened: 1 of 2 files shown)


mahagr:
@rhukster Review please. :) I'm pretty sure it's all ok, but because of I pulled my hair out with this one, I'd rather to have extra pairs of eyes.

mahagr:
I'm thinking if the last case we should revert the ungreedy regex, mostly because of someone may have put unescaped tags into the JS...

mahagr:
I meant greedy again, lol.. Well, now it should work in all cases.

MattAppleton:
Hello!  Since updating to the new 1.6 Grav I have encountered this same problem again...

Previously I have been adding Google fonts to my css like this

`@import url('https://fonts.googleapis.com/css?family=EB+Garamond:400,400i,500,500i,600,600i,700,700i,800,800i');`

This @import is now omitted from the piplined css since the update.

If I put a direct link into my base.html.twig liek so

`<link href="https://fonts.googleapis.com/css?family=EB+Garamond:400,400i" rel="stylesheet">
`

my fonts reappear...

rhukster:
Hmm.. seems this fix got missed in the Assets rewrite.  I will take a look.

PR #1779, opened 2017-12-09
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertInternalType` and `assertNotInternalType` instead of `is_*` functions;
- `assertContains` instead of `in_array` function;
- `assertGreaterThan` for mathematical comparisons;
- `assertNull` and `assertNotNull` instead of strict comparison with `null` keyword;
- `assertTrue` instead of strict comparison with `true` keyword;
- `assertFalse` instead of strict comparison with `false` keyword;
- `assertArrayHasKey` instead of `isset` function;
- `assertSame` instead of comparisons `==` and strict comparisons `===`.

### DIFF
--- a/tests/unit/Grav/Common/AssetsTest.php
+++ b/tests/unit/Grav/Common/AssetsTest.php
@@ -477,9 +477,9 @@ class AssetsTest extends \Codeception\TestCase\Test
 
     public function testGetCollections()
     {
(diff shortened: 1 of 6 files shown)


rhukster:
Very cool, will review and test the tests!

PR #1801, opened 2017-12-25
add Utils::htmlGermanCharacter

macphil:
Add an small function to replace german characters to the corresponding html entities.
Maybe this will be useful for others.

(My md-files contains a lot of german chars :-))

>  Grüße aus Köln :-)

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -688,6 +688,8 @@ class Page
 
         }
 
+        // htmlGermanCharacter
+        $this->content = Utils::htmlGermanCharacter($this->content);
         return $this->content;
     }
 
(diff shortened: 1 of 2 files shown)


rhukster:
You really don't need this function as it's built into PHP with simply:

```
echo htmlentitites("Ich werde mir die größte Mühge geben zu lernen.");
```

macphil:
Of corse you can use the well-known PHP-Function, but then i can't simply replace *only* the german chars within an HTML-content.
[As you can see, i simply process the hole content](https://github.com/getgrav/grav/pull/1801/files#diff-d17e21f858763d0ffcdd6c827cc0eddc), which is not super-performant and smart, but simple.
Due to that fact, i can't use htmlentitites at that point. (e.g. '<' and '>' will lead to their html-entities)

PR #1823, opened 2018-01-07
Remove the default type attribute from script tags

mkutyba:


rhukster:
What's the specific use case for this?

mkutyba:
Nu Html Checker says:
> Warning: The type attribute is unnecessary for JavaScript resources.

PR #1825, opened 2018-01-10
Make 'Default Language' a consistent width

Flamenco:
Make 'Default Language' the same width as the other inputs.

Before:
before

After:
after


hugoaf:
In that field, you need to write the [two letter ISO language code](https://www.w3schools.com/tags/ref_language_codes.asp), you probabbly need  `en` instead of United States. That said, imho current field width should be enough.

Flamenco:
It looks so lonely next to all of the other fields :(

Flamenco:
FYI, the 'United States' was added by Chrome's autocomplete.  I did not enter it manually.  So perhaps the field also needs a bit of validation.

PR #1830, opened 2018-01-15
Clean up Utils class with minor fixes

mahagr:
* `contains()`: allow $needle to be an array just like in `startsWith()` and `endsWith()`
* Add some checks and casting to avoid possible notices
* Some minor code optimizations, use simpler code
* Use more strict comparisons
* Fix broken multi-part uploads in Windows (open file in binary mode)


rhukster:
Looks fine to me.  There are a couple of optimizations which 'might' have an effect, but I doubt. I would like to merge and test at this point, then we can roll back any issues.

PR #1834, opened 2018-01-19
Simplify returns

carusogabriel:


rhukster:
While this does simplify the code a little, it will not increase performance at all, and actually makes things harder to debug.  I find myself having to refactor code sometimes into more 'debuggable' code when fixing bugs, and this will only make it more work.  

I appreciate the PR, but I think that its best to keep the code more readable and easier to debug rather than just saving a few lines.

PR #1853, opened 2018-02-03
Enable styling of summary

raphaeljolivet:
I wrapped explicit summary (before ===) within a DIV (class='summary') in order to be able to style them differently in full blog posts.


rhukster:
Rather than hardcoding this into the core, why not simply wrap it in your twig template:

```
<div class="summary">
{{ page.summary }}
</div>
```

This gives much more flexibility and also guards against potential backward compatibility issues for people that update grav and have something in their theme that styles `.summary` unexpectedly.

PR #1880, opened 2018-02-21
Taxonomy Filtering - space characters

oneprince:
Taxonomies with spaces were not recognized because they are encoded in the uri, but the taxonomy collection was not encoded.  So, they never matched.  It seems like a simple rawurlencode of the taxonomy string should allow them to be found.


rhukster:
thanks

PR #1881, opened 2018-02-21
Feature/taxonomy filtering spaces

oneprince:
Sorry, there was one additional change that is necessary for the spaces to work in the taxonomy filtering.  I had tested, but forgot to clear the cache, so I saw an item filtered that wasn't actually valid.  Once I continued working and refreshed again, all items disappeared...  As you can see with the small change below, the rawurlencode is also needed in the following $uri->param call.

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -2487,7 +2487,7 @@ class Page
         if ($process_taxonomy) {
             foreach ((array)$config->get('site.taxonomies') as $taxonomy) {
                 if ($uri->param(rawurlencode($taxonomy))) {
-                    $items = explode(',', $uri->param($taxonomy));
+                    $items = explode(',', $uri->param(rawurlencode($taxonomy)));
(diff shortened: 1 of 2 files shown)


mahagr:
FYI: 

I just added a new class `Grav\Framework\Route` to Grav 1.4 branch in attempt to fix issue between URI and routing requirements -- in URI's you always want to have escaped paths (as URI may be outputted to the document) and for routing you never want those. There's also new `Grav\Framework\Uri` class which is PSR-7 compatible and allows URI manipulation.

Both can be fetched currently from the old `Grav\Common\Uri` class or generated by a factory class.

oneprince:
@mahagr So how does these updates help in this instance where I want to filter by a taxonomy that has a space character in it?  I want to filter like this:
/news/job%20area:Home%20Installation

gergLeft:
@mahagr the Grav\Framework\Route class that you created, how would I use that since the issue is within the Page class?

mahagr:
Looks like this has been already fixed a while ago.

PR #1884, opened 2018-02-23
Add function to dependency check

Flamenco:
This method is passive and should not affect any current code.

At the very least, plugins can this method onPluginsInitialized(). 

 Ideally the system would check deps after loading blueprints for all plugins.  TBD what to do in this case (disable, log, not call handlers, etc).

Usage
```
public function onPluginsInitialized()
    {
        if (!$this->isAdmin()) {
            return;
        }

        if (!self::checkDependencies($this)) {
            return;
        }

        $this->enable([
            'onTwigExtensions' => ['onTwigExtensions', 0],
            'onAdminMenu' => ['onAdminMenu', 0],
            'onAdminTwigTemplatePaths' => ['onAdminTwigTemplatePaths', 0],
            'onPageNotFound' => ['onPageNotFound', 1],
            'onPagesInitialized' => ['onPagesInitialized', 0],
        ]);
    }
```

To check without posting message to UI, add array param.
```
$issues = [];
if (!Plugin::checkDependencies($thePlugin, $issues)) {
            // log, disable, etc
}
```


Flamenco:
I did not know messages showed up on login screen.  Nice!

screen shot 2018-02-23 at 7 08 33 am

Flamenco:
screen shot 2018-02-22 at 6 53 53 pm

Flamenco:
@rhukster Another reason for this PR:
https://github.com/Flamenco/grav-admin-media-actions/issues/1

Now I have this function in 5 plugins.

1. There needs to be some kind of runtime dependency checking when disabling/removing plugins.
2. The system should catch exceptions generated by plugins and disable them.

I suggest we put a flag in the blueprint of plugins to disable on dependency failure.  Right next to where the dependencies are declare. Default it to false so developers can opt into the check and nothing breaks.

A couple of other issues.
* 'grav' does not show up a plugin, obviously, and needed to be special cased
* 'admin' does not report as 'enabled' in the config and needed to be special cased

Flamenco:
screen shot 2018-02-24 at 7 34 42 am

Issue #1827, opened 2018-01-11
Links not always HTTPS behind off-loaded SSL (AWS)


CSixtyFour:
I've got a site SSL offloaded on AWS. I've found some links such as in the sitemap plugin are not showing as HTTPS even though force_ssl is on in the admin panel and url rewriting is happening in htaccess. It looks like links generated using something like `$uri = $this->grav['uri'];` ( [sitemap plugin](https://github.com/getgrav/grav-plugin-sitemap/blob/c459e22e8603a5580035deb3023674fb0e0d314b/sitemap.php#L40) ) don't show https. 

This is my first time diving into Grav core but I think there are two things here:
1) force_ssl is happening to high up for direct access to `uri` so therefore in this situation this option is irrelevant in this use case.

2) In `uri.php` I think the two important environmental variables are $_SERVER['REQUEST_SCHEME'] and $_SERVER['SERVER_PORT'] both of which would be http / 80 behind an SSL offloader but probably accurate in a "normal" environment. 

I have done some messy hacking for my setup but maybe $_SERVER['HTTP_X_FORWARDED_PORT'] and $_SERVER['HTTP_X_FORWARDED_PROTO] could be an override option for SSL offloading?

rhukster:
```
$uri = $this->grav['uri'];
```

Simply gets the URI object.  `$uri->path()` will get the path, and `$uri->url()` would get the URL. If you pass `true` you should get the full path including `http://` or `https://`: https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Uri.php#L517

Maybe we should take into account `HTTP_X_FORWARDED_PORT` and `HTTP_X_FORWARDED_PROTO` here: 

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Uri.php#L745-L765

and here: 

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Uri.php#L61-L66

Similar to what we do for IP: 

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Uri.php#L745-L765

AquaL1te:
Will this fix be released soon? My sitemap also includes only http links, while my website is only accessible via https. This doesn't work with Google webmasters indexing.

drnasin:
is there a reason this isn't released yet?

rhukster:
I'm a bit confused because this was merged ages ago! https://github.com/getgrav/grav/pull/1888/files

[...]

PR #1888, opened 2018-02-23
Fix for #1827

CSixtyFour:
Modified Build scheme and Build port functions to take into account those in SSL offloading situations.

An attempt to fix: Links not always HTTPS behind off-loaded SSL (AWS) #1827


rhukster:
This got lost with the Grav 1.4 merge. Going to re-add it manually.

PR #1945, opened 2018-03-27
Add overriding of setup.php config file, via constants

rexfordkelly:
I'm working on a custom install of Grav, where being able to dynamically swap setup files was required. I've modified the $file assignment to permit overriding it's value via a defined constant, "GRAV_SETUP_PATH".

And I thought that might be useful to others, so I'm submitting a pull request.


rhukster:
Looks good... would be good to get this documented in learn.getgrav.org too.

Issue #1920, opened 2018-03-15
extract error during upgrade installation


jens-t:
following error occurs:
```
Preparing to upgrade to v1.4.1..
  |- Downloading upgrade [3.28M]...   100%
  |- Installing upgrade...    error                             
  |  '- An error occurred while extracting the package
  '- Installation failed or aborted.
```

i tried --verbose, set errors to true in system.yaml, nothing inside the logs, only "An error occurred while extracting the package".

has anyone an idea how i can get decent error information, what's failing during extraction?

Thanks!

Edit: I think it's a permission problem, but i dont know how to get the tmp path where the files are extracted to ...

rhukster:
The code where this can be set is in the `Grav\Common\GPM\Installer.php`:

```
public static function unZip($zip_file, $destination)
    {
        $zip = new \ZipArchive();
        $archive = $zip->open($zip_file);

        if ($archive === true) {
            Folder::mkdir($destination);

            $unzip = $zip->extractTo($destination);


            if (!$unzip) {
                self::$error = self::ZIP_EXTRACT_ERROR;
                Folder::delete($destination);
                $zip->close();
                return false;
            }

            $package_folder_name = preg_replace('#\./$#', '', $zip->getNameIndex(0));
            $zip->close();
            $extracted_folder = $destination . '/' . $package_folder_name;

            return $extracted_folder;
        }

        self::$error = self::ZIP_EXTRACT_ERROR;
```

Basically, all we're doing is trying to open the zip file, if that archive is false, this error gets set.

The temp folder used is `$tmp_dir = Grav::instance()['locator']->findResource('tmp://', true, true);` so this is usually `tmp/` in your Grav installation.

I have got a [Pull Requst](https://github.com/getgrav/grav/pull/1922/files) where i've added more verbose error code handling, perhaps you could test with that?

jens-t:
Thanks for the fast answer! That PR looks promising, will try that :)

I already tried to debug the `Grav\Common\GPM\Installer.php` but i didn't succeed in getting some additional debug output into the shell. Are the cli scripts cached in some way? Or, is there a simple way to dump some debug information triggered from the `Grav\Common\GPM\Installer.php` to the shell, when using `bin/gpm selfupgrade`?

rhukster:
You can dump to the log file by inserting:

```
Grav::instance()['log']->notice('My notice message');
```

in the code.  Or if you in GPM, you can simply `echo()` stuff and it will display in the console as output.

[...]

PR #1948, opened 2018-03-28
gpm selfupgrade: unzip error fix ...

jens-t:
**Error**:

The zip error handling was broken, it caused the commandline installer to show an upgrade success if the extraction fails.

**Changes:**

That the getZipError function returns strings & integers wasn't the problem. The bitwise operators in the `SelfupgradeCommand.php` where causing the condition to fail, where it should be true. So removed the bitwise operators from the condition.

And: Added a hint, what could be causing the "Not a zip archive."-Error.

**Questions:**

Why using bitwise operators at all? Isn't that bad coding style? I don't even understand what it exactly does. Can someone explain the reasoning behind this for me, would like to understand this. Thanks!

### DIFF
--- a/system/src/Grav/Common/GPM/Installer.php
+++ b/system/src/Grav/Common/GPM/Installer.php
@@ -219,7 +219,7 @@ class Installer
                 break;
 
             case \ZipArchive::ER_NOZIP:
-                $error = "Not a zip archive.";
+                $error = "Not a zip archive. (Download failed or incomplete?)";
                 break;
 
             case \ZipArchive::ER_OPEN:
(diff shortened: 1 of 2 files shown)


jens-t:
If i understand it right, the bitwise operators were used to filter for "valid" error codes. so only the two integer values of the constants would pass the condition. since the unzip errors are strings maybe something like this would be better, if you want to keep the integer validation:
`if ($errorCode & (Installer::ZIP_OPEN_ERROR | Installer::ZIP_EXTRACT_ERROR) || is_string($errorCode)) {`

The var name `$errorCode` is now also a bit misleading, maybe it is a good idea to rename it to `$error` ...

w00fz:
Ok this requires a bit of caution. I didnt realize last changes with the extra error messages where put in the errorcode. Error code is supposed to be just an int, we cant mix them.
Your pr will remove bits of logic that grants proper check for the zip access. Removing the checks is not really a solution.
The bitwise checks are used becausw the error code are bit codes in base 8.

I will look and review this

w00fz:
Closing as the it's been resolved differently as discussed in #1920

PR #1979, opened 2018-04-16
Add default configuration for images

cweiske:
This patch allows configuring image processing instructions
for all images.
A use case is generating a fixed set of sizes (derivatives) for
all images, without having to specify them in the content markdown.

Example configuration in user/config/system.yaml:

```
images:
  defaults:
    derivatives: "[300,600,1200]"
```


rhukster:
That's pretty much how i would of implemented it 👍

PR #1992, opened 2018-04-24
Add MIME type to video <source> tag

cweiske:


rhukster:
Seems if you want to use this optional `type` on `<source>` it needs to be more than just the mimetype:

https://www.html5rocks.com/en/tutorials/video/basics/

cweiske:
The MIME type itself is completely sufficient. The codecs parameter is optional.

It is also only relevant if there are multiple video/mp4 sources with different codecs, and we want to give the browser enough information to allow it decide what to fetch.
But currently Grav does only add one `<source>` tag, so it's not needed at all.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source

rhukster:
cheers

rhukster:
Ack, spoke to soon. Tested this and it broke my test cases of with a `.mov` file.  Wouldn't play until i reverted the addtion of `type`

mahagr:
Wrong issue referenced ^

PR #2000, opened 2018-04-27
Video autoposter

cweiske:
Automatically detect video thumbnail
The image file will be added as poster attribute to the video tag.

Looks for the following files in the video folder:
- $filename.jpg
- $filename.png
- $filename.$ext.thumb.jpg
- $filename.$ext.thumb.png

The code is hacky but works for me, and I don't know the Grav API good enough
to do it better.

----

It's currently based on https://github.com/getgrav/grav/pull/1992 but can be used without it.


cweiske:
The code requires PHP 7.0 because of `dirname($path, level)` use. Is this a problem?

rhukster:
Yes its a problem. Grav currently supports 5.5+

Also this also contains your mimetype PR from a previous PR request, and that needs to me modified to support codecs.

PR #2004, opened 2018-05-02
Move fix for #1114 to Truncator::truncateLetters

dliessi:
The original fix provided by #1125 in Utils::truncateHTML compared the
summary size with the full HTML string length.
Thus, the string was still being truncated (and the HTML rewritten) even
when only the HTML string length, and not the text length, exceeded the
summary size.

### DIFF
--- a/system/src/Grav/Common/Helpers/Truncator.php
+++ b/system/src/Grav/Common/Helpers/Truncator.php
@@ -47,6 +47,7 @@ class Truncator {
 
         // Iterate over words.
         $words = new DOMWordsIterator($body);
+        $truncated = false;
         foreach ($words as $word) {
 
             // If we have exceeded the limit, we delete the remainder of the content.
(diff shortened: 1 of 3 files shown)


dliessi:
The failing test is
```
$this->assertEquals('<input type="file" id="file" multiple>', Utils::truncateHtml('<input type="file" id="file" multiple />', 6));
```

I suspect that the first argument of `assertEquals` was chosen to be different from the first argument of `Utils::truncateHtml` only  because `Truncator::innerHtml` rewrites the HTML.
Is this true? Can I make them equal? Which form should be retained (with or without the closing ` /`)?

dliessi:
I chose to keep the XHTML syntax, based on other code generated by Grav, although in HTML it is unnecessary.

rhukster:
looks good.. cheers.

PR #2018, opened 2018-05-14
Do not crash when generating URLs with arrays as parameters

cweiske:
When using a default image derivatives (#1979) setting like

    images:
      defaults:
        derivatives: [300,600]

then Grav crashes when a video gets embedded.
The following error occurs:

    rawurlencode() expects parameter 1 to be string, array given
    system/src/Grav/Common/Page/Medium/Medium.php:521

This patch encodes array URL parameters correctly.


rhukster:
cheers..

PR #2040, opened 2018-05-30
Add cookie domain configuration under system.session

MaxEvan:
I have added a new config under "system.session" called "domain".

The current implementation is 

```php
$domain = $uri->host();
if ($domain === 'localhost') {
    $domain = '';
}
```

This is problematic under docker-compose because the hostname is the name of the container/service. In our case, it was forcing the cookie domain to be "domain=grav-site;".

Basically, under docker, the only way to make it work would be to use "localhost", but that would break a lot of stuff. 

This setting allows override of the domain, and it shouldn't break existing applications.


mahagr:
Implemented in above commit.

PR #2043, opened 2018-05-30
Update robots.txt

drnasin:
When using "Fetch and Render as Google" in Google Search console it will report "partial" rendering due to the blocked images in /user/images directory which is blocked because of the Disallow /user/ rule. Proposing this small change as it improves google rendering of the page.


rhukster:
Sounds good.

Issue #1818, opened 2018-01-04
Let Grav scan/load template files with different extension than *.html.twig


numito:
Hello,

Grav is awesome, I have created a plugin to write template files in HAML with Twig integration. It allows to have concise, clear and short templates. It integrates greatly with Grav and Twig. You can get the plugin here: https://github.com/numito/grav-haml 
But I figured out an issue, Grav scans only for .html.twig files in theme folders, also the type 'HTML' and '.html.twig' is not a constant and is redefined in many places in the core files. Grav should have a configurable list of template extensions to scan for and load those templates. By default all template should go through the Twig parser, but a plugin could hook for specific template extension allowing additional template languages. 
I made a few simple modifications to core files to allow that, but I am not sure they are exhaustive. I would like some recommendations. 
These are the modifications I did to make my plugin work:

Line 60 of Grav\Common\Twig\Twig
Add
```php
    public $templateExt = TEMPLATE_EXT;
```

Ligne 240 modify
```php
 if ($item->modularTwig()) {
                $twig_vars['content'] = $content;
                $template = $item->template() . $this->templateExt;
                $output = $content = $local_twig->render($template, $twig_vars);
 }
```

Line 352 modify
````php
 if ($ext != $this->templateExt) {
                try {
                    $page->templateFormat('html');
                    $output = $this->twig->render($page->template() . $this->templateExt, $vars + $twig_vars);

                } catch (\Twig_Error_Loader $e) {
                    throw new \RuntimeException($error_msg, 400, $e);
                }
            } else {
                throw new \RuntimeException($error_msg, 400, $e);
 }
````

Line 72 et 74 de Grav\Common\Page\Types
```php 
public function scanTemplates($uri) {
       $options = [
            'compare' => 'Filename',
            'pattern' => '/\.(haml|html)\.twig$/',
            'filters' => [
                'value' => '/\.(haml|html)\.twig$/'
            ],
            'value' => 'Filename',
            'recursive' => false
        ];
}

rhukster:
i think its not this simple.  The admin needs to be looked at also.  I would prefer to improve this with something more flexible, not just `haml|html` but something that lets you set the markup output if its not inline with the request.  By default grav will look for `.json.twig` if you request a URL with `.json`, and `.html.twig` if you request `.html` (or leave it blank as it's default).  

This needs to be tackled at a higher level i think.  I will think on this a bit more.

rhukster:
Does your HAML plugin require these code changes to work???

[...]

PR #2044, opened 2018-05-31
Modular template extension follows the master page extension

jbanety:
This PR is related to the issue #1818.
This builds the modular template extension from the URI's extension (i.e. `http://examp.le/my-page.amp` or `http://examp.le/other-page.docx`).
It fallbacks to `TEMPLATE_EXT` const if URI extension is `null` or empty.

---
 
For my project : 

I'm creating an [AMP](https://www.ampproject.org) version of my website in the same theme.

my custom `system.yaml`
```yaml
...
pages:
  ...
  types:
    - html
    - amp
  ...
```

So I have two template extensions : `.html.twig` for the regular pages and `.amp.twig` for the AMP pages.

I'm using the same content for the two versions.

If I create a modular page with the `.amp.twig`, Twig will load modular templates with the extension `.html.twig` (which will be wrong formated for AMP framework and not validated).

### DIFF
--- a/system/src/Grav/Common/Twig/Twig.php
+++ b/system/src/Grav/Common/Twig/Twig.php
@@ -238,7 +238,9 @@ class Twig
             // Process Modular Twig
             if ($item->modularTwig()) {
                 $twig_vars['content'] = $content;
-                $template = $item->template() . TEMPLATE_EXT;
+                $extension = $this->grav['uri']->extension();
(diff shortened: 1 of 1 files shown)


rhukster:
Should this be limited only to modular pages? Does this make sense for modular and non-modular both??

jbanety:
Modular template name is hard-coded with the coconstant.s
So yes It's fix for modular only.

rhukster:
Ah yes, I see now.. Cheers.

PR #2045, opened 2018-05-31
Fix image memory use

tsnorri:
Creating image derivatives currently keeps all the images in memory until the PHP interpreter is finished. With a substantial number of files or large images this can cause out of memory errors. I attempted to fix the issue by saving the image on disk immediately after the derivative has been created and then freeing it.

Processing large images still takes time. Ideally, multiple threads could be used to create the derivatives, if there is enough memory.

Please note that currently Gregwar\Image\Adapter\GD does not free the associated image resource on deallocation. I've created a [pull request](https://github.com/Gregwar/Image/pull/154) to fix this, too. (The checks don't pass because of a configuration error.)

### DIFF
--- a/system/src/Grav/Common/Page/Medium/ImageMedium.php
+++ b/system/src/Grav/Common/Page/Medium/ImageMedium.php
@@ -301,6 +301,7 @@ class ImageMedium extends Medium
                 $derivative->set('width', $width);
                 $derivative->set('height', $height);
 
+                $derivative->resetImage();
                 $this->addAlternative($ratio, $derivative);
             }
         }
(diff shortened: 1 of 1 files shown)


rhukster:
Looks good!

system/src/Grav/Common/Page/Medium/ImageMedium.php
```diff
@@ -587,7 +597,9 @@ protected function saveImage()
             $this->image->merge(ImageFile::open($overlay));
         }
 
-        return $this->image->cacheFile($this->format, $this->quality);
+        $cachedPath = $this->image->cacheFile($this->format, $this->quality);
+        $this->set('filepath', $cachedPath);
```
mahagr: This line breaks image handling as the image quality gets worse and worse on every resize.

mahagr:
@tsnorri Can you please try my approach in your use case?

tsnorri:
@mahagr Unfortunately I had a rather tight schedule and then forgot about this. I just tested with Grav 1.5.8 and did not get any out-of-memory errors.

PR #2064, opened 2018-06-19
append_url_extension unignored

scaccogatto:
I found that this header variable:

`append_url_extension: json`

is not correclty used within `Page.php` rendering section.

Here is the fix!


rhukster:
I need to confirm that this won't have the unintended result of appending the default extension when it's not wanted.

scaccogatto:
I would like to bring focus on this again: after the last update the problem is still here.
I use this patch in a production website and there is no problem at all.

PR #2083, opened 2018-07-02
Update Media.php

tcyrus:
Fixes getgrav/grav-plugin-admin#1330. Media with Unicode filenames are not detected because `SplFileInfo::getBasename` ignores names which don't match the locale. `SplFileInfo::getFilename` does the same thing as `SplFileInfo::getBasename` without this issue.


rhukster:
Cheers!

tcyrus:
Thanks for the quick merge, I would suggest replacing `getBasename` in other places as well to avoid similar issues.

mahagr:
Please add a new issue about it so it won't get forgotten. :)

PR #2087, opened 2018-07-04
Change getBasename to getFilename where possible

tcyrus:
`SplFileInfo::getFilename` returns the same value as `SplFileInfo::getBasename` (no arguments) while being locale safe (See #2083). For situations where `SplFileInfo::getBasename` needs to accept an argument, I would suggest implementing a `basename` helper function ([Good Example](https://github.com/cakephp/cakephp/blob/659a25db6d8013d5e187aec5bcfea47cdafa418f/src/Filesystem/File.php#L355)).

### DIFF
--- a/system/src/Grav/Common/Config/ConfigFileFinder.php
+++ b/system/src/Grav/Common/Config/ConfigFileFinder.php
@@ -207,7 +207,7 @@ class ConfigFileFinder
                     continue;
                 }
 
-                $name = $directory->getBasename();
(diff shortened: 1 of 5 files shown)


rhukster:
Thanks!

mahagr:
@tcyrus Found few in admin plugin in case if you want to take care of them. :)

tcyrus:
@mahagr, Made a pull request

PR #2103, opened 2018-07-18
Add Grav version to PHPdebugbar

hughbris:
Adds "Grav v*x.y.z*" as first line info to Messages tab on debug bar

PR #2106, opened 2018-07-19
Add Grav version to PHPdebugbar

hughbris:
Adds "Grav vx.y.z" as first line info to Messages tab on debug bar.

Second crack at passing PR, first was https://github.com/getgrav/grav/pull/2103


rhukster:
Cheers...

PR #2121, opened 2018-07-27
Fix broken nounce handling

gn-jgeerds:
It seems to me that the current implementation of nounces are broken. Please read my commit messages as they should explain everything:

---

Remove deprecated "getNonceOldStyle" function

This commit removes the following functions:

- getNonceOldStyle
- generateNonceStringOldStyle

The functions have been replaced in newer versions of
grav. It seems to me that they only existed in order to make a
upgrade to a newer version of grav painless (i.e. accept both types of
nonce tokens). Nowadays, existing old style nonces are expired long
time ago so it should be save to delete the deprecated funtions.

---

Don't save nonces in static class variable

Currently, the behavior of `getNonce` is broken because it saves the
generated nonce in an array and only use the $action as the
key. However, the generated nonce does not only depend on the $action,
but also on $plusOneTick.

I can't imagine that this is a performance critical part of grav. So
the easiest way is to remove the whole "caching" mechanism. If we
really need some kind of caching, we should implement it in a proper
way and also think about how (or when) the cache should be
invalidated.

---

Fix broken "plusOneTick" for nonces

It looks to me that there is a bug in the current implemention of
verifyNonce. Here is an example:

- 2018-08-01 10:00: We respond to a request and generate a nonce. The
  current tick is at 35489

- 2018-08-01 10:05: We use the previously generated nonce to make
  another request. We compare the given nounce with a new generated
  one (based on the same tick). The result is exactly the same and the
  request succeeds.

- 2018-08-01 14:00: We're now one tick ahead. Remember: A day (24
  hours) is separated into two ticks (each 12 hours). A request comes
  in, we compare the given nounce with a newly generated one based on
  the current tick (now at 35490). They don't match (which is totally
  okay).

  If the comparison fails, we then compare the given nounce with a
  another, newly generated one. This time, we pass "plusOneTick", to
  the function, which increases the current tick by one. Our tick is
  now at 35491. We generate a nonce based on that tick and of course,
  it still does not match the given nonce.

  Instead of increasing the tick, we should rather decreasing it by
  one (i.e. use the previous tick). If the first comparison fails, we
  use the current tick (35490), decrease it by one (35489) and then
  compare it again. 35489 is the same tick as in the very first
  request.

This bug leads to a maximum life time of 12 hours for a nonce and in
worst case only a few seconds (!)

I would like to prove the bug with an unit test but I'm too unexperienced
in PHP. Furthermore it seems that we need some kind of library which
is able to mock builtin functions (like "time"). Maybe
<https://github.com/Codeception/AspectMock> would be a good canditate?

---

I did some tests inside a docker container (with `libfaketime` to modify the time inside the container). I can reproduce the issues and also that my changes fix them.

You can more ore less easily reproduce the issue yourself:

1) Install `grav-plugin-admin` and open the admin interface in your brower
2) Make sure it is before 12:00 (UTC)
3) Click on "Pages"
4) Click on "Add" or open an existing page
5) ...wait for the next tick
6) Click on "Save"

What happens: Now you will see the `PLUGIN_ADMIN.INVALID_SECURITY_TOKEN` error message.

What I expected: Successful request. The nonce (token) is not older than 24 hours so it should just work.

### DIFF
--- a/system/src/Grav/Common/Utils.php
+++ b/system/src/Grav/Common/Utils.php
@@ -727,27 +727,6 @@ abstract class Utils
         return ($i . '|' . $action . '|' . $username . '|' . $token . '|' . Grav::instance()['config']->get('security.salt'));
     }
 
-    //Added in version 1.0.8 to ensure that existing nonces are not broken.
-    private static function generateNonceStringOldStyle($action, $plusOneTick = false)
-    {
-        if (isset(Grav::instance()['user'])) {
-            $user = Grav::instance()['user'];
-            $username = $user->username;
-            if (isset($_SERVER['REMOTE_ADDR'])) {
-                $username .= $_SERVER['REMOTE_ADDR'];
-            }
-        } else {
-            $username = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : '';
-        }
-        $token = session_id();
-        $i = self::nonceTick();
-        if ($plusOneTick) {
-            $i++;
-        }
-
-        return ($i . '|' . $action . '|' . $username . '|' . $token . '|' . Grav::instance()['config']->get('security.salt'));
-    }
-
     /**
      * Get the time-dependent variable for nonce creation.
      *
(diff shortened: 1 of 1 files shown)


mahagr:
I believe I've run into the issue a few times (unexpected error messages), so I do think it is important to find a fix for this, but I do agree that the nonce function should still have caching from the reasons given by @rhukster .

gn-jgeerds:
@rhukster Now I see your point, thanks for the clarification.

I have updated the PR:
- reverted the removal of nonces caching
- fix caching so it works with `$action` and `$previousTick`
- squashed the new commits together
- rebased against current development branch

I'll test the changes in the next few days but I think they should work as expected. What do you think?

gn-jgeerds:
Can I do anything to speed things up?

rhukster:
I just need to do some testing..

rhukster:
OK, tested and looks good.  Thank!

PR #2138, opened 2018-08-08
Allow capitalized usernames if the regex agrees

newbthenewbd:
I don't know the exact motivation behind these uses of `strtolower`, but I believe that the existence of the username regular expression in Grav's config clearly implies that the usernames don't just _have_ to be lowercase, hence this pull request.

### DIFF
--- a/system/src/Grav/Common/User/User.php
+++ b/system/src/Grav/Common/User/User.php
@@ -33,6 +33,9 @@ class User extends Data
         /** @var UniformResourceLocator $locator */
         $locator = $grav['locator'];
 
+        // force lowercase of username
+        $username = mb_strtolower($username);
+
         $blueprints = new Blueprints;
         $blueprint = $blueprints->get('user/account');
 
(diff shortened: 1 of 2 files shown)


newbthenewbd:
Right, completely forgot about this limitation, I apologize.

Still, the way this currently is implemented, Grav noticeably lowers its chances of being used in any sort of social platform, except possibly one about retro computing, limiting its usage to great-looking, high-performance blogs. Due to that, I believe this should still be somehow fixed.

The best implementation I can spontaneously think about is to prepend the usernames with the character case masks, in an optimal number system, i.e. `2002_xx13375h0073rxx` for `xX13375h0073rXx` because hexadecimal `2002` is binary `0010000000000010`. That shouldn't matter at all for websites which only ever have a few admins registered, while the slightly increased disk usage is certainly a better solution for those who want more. Note also that this seems like a naive implementation, as the numbers could visibly be omitted here, however the intuitively better one might use too much CPU time in the lovely language PHP is, hence needing some tests.

Due to that, I request that this pull request stays open for now, with higher quality commits hopefully appearing soon.

mahagr:
This may help to get started:

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/User/User.php#L206

newbthenewbd:
Please review it again. Note that the bug about usernames not getting saved lowercase has been patched by removing the `!$file->filename()` conditional, which seemed to erroneously return false at all times, however I'm admittedly not exactly sure what it was supposed to do in there.

Thanks!

newbthenewbd:
Whoops, forgot about that one (and totally haven't woken up suddenly realizing this), here we go, hopefully.

mahagr:
Looks much better now. The only thing which I don't like is to remove the filename check as it causes filename to potentially change after every save thus causing the user to duplicate.

newbthenewbd:
Closed due to ~the general mess in this pull request's commits and~ the release of Grav 1.5.0. Should totally make a new pull request for the `develop` branch Soon™.

Issue #2140, opened 2018-08-09
Theme inheritance and config files


Perlkonig:
Let's start with the ask: **can Grav be altered to actually *process* an inherited YAML file instead of just obscuring it?**

I'm dealing with [an inheritance issue](https://github.com/Perlkonig/grav-theme-knowledge-base/issues/24) with my [Knowledge Base theme](https://github.com/Perlkonig/grav-theme-knowledge-base).

The fundamental problem is that Grav inheritance is all or nothing. New files simply obscure inherited ones. This makes sense for Twig files, but this is problematic when your theme uses the theme YAML file. When you inherit a theme, the docs instruct you to create a new theme config file with the `streams` settings, which clobbers any settings in the inherited config file.

So first of all, I'll be submitting a PR against the docs adding some additional instruction to work around this issue.

But ideally, can Grav be altered to actually *process* the inherited YAML file instead of just obscuring it? That way the person inheriting the theme can still create a new file, drop in the `streams` stuff, and then just overwrite individual settings, but all the theme defaults will still be set.

I'd go straight to a PR, but I'm not familiar enough with the internals to know where to start. With some direction, I'm certainly willing to try my hand at it. 

Thanks!

OleVik:
I'm in favor of more programmatic treatment of blueprints, to generate them rather than state them, so we can get more granular control and less reiterating existing fields with the same settings. I think you'd have to write an interface for generating and iterating through the hierarchy of blueprints to accomplish this, but I'm not overly familiar with the internals either.

Perlkonig:
Found the code. Possible fix suggested.

Mimoza:
I just spent some time debugging this exact behavior, so let me add a fresh datapoint:

**Setup:** Grav 2.0, theme `editorial` 4.3.1 (multilingual site, fr/es/en/eo), plus a site-specific child theme that only overrides the homepage template (`streams` block + PHP class extending the parent — the documented setup).

**What happened:** All theme settings are stored in `user/config/themes/editorial.yaml` (langswitcher menu icon, search icon, custom logos, copyright, contact block, custom menu, sidebar widgets...). The moment I activated the child theme, `theme_var()` could only see `themes.anargile-leanbertaire-editorial`, which is empty — the parent user config is never consulted. The sidebar silently lost the language selector, the search box, the custom logos, the contact block... Every `theme_var()` fell back to its `|defined(false)` default.

[...]

PR #2143, opened 2018-08-11
Merge all theme config files together when inherited

Perlkonig:
Naive solution to PR #2140. I say "naive" because (a) I can only do so much testing and (b) it assumes Grav never uses prefixes in theme inheritance. The docs never mention it, so I assume it's true.

The code still merges `user/config/themes` files correctly. The change should be mostly backwards compatible. The only people affected will be those who both 
  * have inherited from a theme that uses the config file to affect behaviour AND
  * have relied on the nonexistence of a clobbered setting rather then copying the settings over and changing them.

Such scenarios are easily fixed by overriding the desired settings.

If accepted, I'd then submit a separate PR against `grav-learn` explaining how to inherit themes given this new behaviour.

I tested it as follows:
  * I added the setting `inherited: "No!"` to my root theme.
  * I then created a new child theme. In that theme's YAML I did the following:
    * Added the correct `streams` settings.
    * Added the setting `inherited: "Yes!"`.
    * Added the setting `uniqueRoot: true`.
  * I then created a new YAML file for the *child* theme in `user/config/themes` and did the following:
    * Added the setting `inherited: "Double!"`.
    * Added the setting `uniqueCustom: true`.
  * Then I went to my dev site, picked a demo page, and added the following: `{{ dump(grav.theme.config) }}`

Here's the result ([visible here](https://www.perlkonig.com/dev/demos/twig-extensions)):

```
array:6 [
  "enabled" => true
  "dropdown" => array:1 [
    "enabled" => true
  ]
  "inherited" => "Double!"
  "streams" => array:1 [
    "schemes" => array:1 [
      "theme" => array:2 [
        "type" => "ReadOnlyStream"
        "prefixes" => array:1 [
          "" => array:2 [
            0 => "user/themes/newtheme"
            1 => "user/themes/minimalist-blog"
          ]
        ]
      ]
    ]
  ]
  "uniqueRoot" => true
  "uniqueCustom" => true
]
```

### DIFF
--- a/system/src/Grav/Common/Themes.php
+++ b/system/src/Grav/Common/Themes.php
@@ -262,6 +262,12 @@ class Themes extends Iterator
             }
         }
 
+        // If theme is inherited, merge the theme config files
+        $paths = $config['prefixes'][''];
+        if (count($paths) > 1) {
+            $this->mergeConfiguration($name, $paths, $this->config);    
+        }
+
         // Load languages after streams has been properly initialized
         $this->loadLanguages($this->config);
     }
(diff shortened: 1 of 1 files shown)


Perlkonig:
Just checking in on old PRs.

hughbris:
Could one of the core team please comment on this or its associated issue? It addresses a major pain point when using inherited themes. No comments or assignment from any core devs after a few years, I'm wondering what's up with this.

mahagr:
You can do that from the blueprint; just extend the theme just like you would extend a page.

The way that this PR tries to do it just messes up the blueprints by blindly merging all the arrays. The method here does not merge blueprints correctly and can cause undesired side effects.

hughbris:
@mahagr I got your response by email notification but can't see it here. Strange, maybe you accidentally deleted it. Do you mind me responding here?

Basically, it would be great if you could outline your preferred approach using blueprints. I will add it to Grav Learn because I think it's important. Cheers.

PR #2146, opened 2018-08-14
Add a database select option for Redis

sandermangel:
Redis support 16 (0 through 15) separate databases by default, this adds a way to select the database via the system.yaml


mahagr:
Already implemented: https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Cache.php#L316

PR #2165, opened 2018-08-31
Allow users to override plugin handler priorities

Perlkonig:
Not sure how pressing this is, but I've seen it once or twice on the forums, and it seems easy enough to fix.

Most plugin authors leave the priority of their listeners at 0 unless there is some internal reason to tweak them. But no plugin author can predict what sorts of interactions may exist with all other plugins. If a user discovers that they need certain plugins to run before others, the only current way to change that is by modifying the priorities in the plugin's `.php` file. Which will get clobbered every time the plugin is updated.

This PR allows users to create a `user/config/priorities.yaml` file that may contain the following data:

```yaml
pluginName:
  eventName:    #for example, "onPagesInitialized"
    listenerName: [integer]
```

The PR replaces the ternary if/else statement in `Grav\Common\Plugin::enable` with a new function `getPriority` that now has three possibilities instead of two:

1. If a value is provided in `priorities.yaml`, it is returned.
2. Otherwise it uses the value from the plugin's `.php` file, if provided.
3. Otherwise it returns 0.

This is advanced usage that still requires users to do some digging to find the listeners that need adjusting. This PR is backwards compatible unless there's some other use for `user/config/priorities.yaml` that I'm not aware of. It does appear to work with plugins with hyphens in their names.

I will draft up a PR to `grav-learn` to match this approach. If implemented, I can see someone developing an admin plugin to make it really easy to make these adjustments.

This is only one approach. There may be very good reasons why this approach doesn't work. I'm open to any feedback. I don't know how to test the effect my changes have on performance. I'm open to instruction.


mahagr:
Interesting thought; I've been thinking of ways of improving plugin events and this is certainly one way to improve it.

Perlkonig:
Just checking in on old PRs.

rhukster:
Thanks! Definitely, need to get this documented...

Perlkonig:
Thanks! PR to learn already submitted.

rhukster:
I didn't see the Learn PR, do you have a URL?

Perlkonig:
I referenced up top. Here it is again.

https://github.com/getgrav/grav-learn/pull/643

PR #2175, opened 2018-09-09
FileStorage Adapter

OleVik:
PSR-16 compliant adapter for abstract handling of transient and persistent files. The difference from [FileCache.php](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Framework/Cache/Adapter/FileCache.php) is that the implementation is not outright set up for caching with Grav's normal setup, but rather a destination must be explicitly given. Otherwise they largely overlap.

For an example use, see this [SchemaBlueprintCommand.php](https://github.com/OleVik/grav-plugin-timeline/blob/f886912e579d937fd317ec7220d2f74fea5848d3/cli/SchemaBlueprintCommand.php), where plugin option determine the permanence of the file. Ie., `transient` for normal /cache, `native` for within-plugin (busted on new version), or `persist` for /user/data. This command intentionally deletes the file regardless, elsewhere it is used to check for existence or recreate (planned features in the plugin).

@mahagr Thoughts? I dropped the Symfony-components.

### DIFF
--- a/system/src/Grav/Framework/Cache/Adapter/FileStorage.php
+++ b/system/src/Grav/Framework/Cache/Adapter/FileStorage.php
@@ -117,7 +117,7 @@ class FileStorage extends AbstractCache
 
                 return rename($this->tmp, $file);
             } catch (\CacheException $e) {
-                throw new \CacheException($e);
+                throw $e;
             } finally {
                 restore_error_handler();
             }
--- a/system/src/Grav/Framework/Cache/Adapter/FileStorage.php
+++ b/system/src/Grav/Framework/Cache/Adapter/FileStorage.php
@@ -1,0 +1,179 @@
+<?php
+/**
+ * @package Grav\Framework\Cache
+ *
+ * @license MIT License; see LICENSE file for details.
+ */
+
+namespace Grav\Framework\Cache\Adapter;
+
+use Grav\Framework\Cache\AbstractCache;
+use Grav\Framework\Cache\Exception\CacheException;
+use Grav\Framework\Cache\Exception\InvalidArgumentException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Exception\IOExceptionInterface;
+
+/**
+ * Storage class for PSR-16 compatible "Simple Cache" implementation using file 
+ * backend. Defaults to 1 year TTL.
+ *
+ * @package Grav\Framework\Cache
+ */
+class FileStorage extends AbstractCache
+{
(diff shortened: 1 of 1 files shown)


mahagr:
Is there a way to extend the `FileCache` and just override constructor and `getFile()` methods? It may need `AbstractFileCache` base class without the constructor, though :) Think about it.

system/src/Grav/Framework/Cache/Adapter/FileStorage.php
```diff
@@ -0,0 +1,179 @@
...
+
+                file_put_contents($this->tmp, $value, LOCK_EX);
+
+                if ($expiresAt !== null) {
+                    touch($this->tmp, $expiresAt);
+                }
+
+                return rename($this->tmp, $file);
+            } catch (\CacheException $e) {
+                throw new \CacheException($e);
```
mahagr: This won't work. use `throw $e;` instead.

mahagr (changes requested):

system/src/Grav/Framework/Cache/Adapter/FileStorage.php
OleVik: Changed to just `throw $e;`.

OleVik:
There is, largely, but not without switching from `private` to `protected` on FileCache's `init()` and `write()`. Also, is it preferable to keep the `ttl` and `key` in the output when the file is stored - and not cached - like this? Because of [`doSet()`](https://github.com/getgrav/grav/blob/1.6.0-beta.4/system/src/Grav/Framework/Cache/Adapter/FileCache.php#L74), both will occupy a line each.

Eg., from the reference-command in the initial proposal, the resulting files ends up:

```
1540633949
Event.schema.json
s:389:"{"Schema":{"Thing":{"Event":{"BusinessEvent":null,"ChildrensEvent":null,"ComedyEvent":null,"DanceEvent":null,"DeliveryEvent":null,"EducationEvent":null,"Festival":null,"FoodEvent":null,"LiteraryEvent":null,"MusicEvent":null,"PublicationEvent":{"BroadcastEvent":null,"OnDemandEvent":null},"SaleEvent":null,"SocialEvent":null,"SportsEvent":null,"TheaterEvent":null,"VisualArtsEvent":null}}}}";
```

Wouldn't I also have to override [`write()`](https://github.com/getgrav/grav/blob/1.6.0-beta.4/system/src/Grav/Framework/Cache/Adapter/FileCache.php#L187) to preempt the `ttl`-check when extending FileCache?

PR #2179, opened 2018-09-12
Redirect regex pattern to match entire $source_url

Ralla:
**Problem**
The regex checking for redirects defined in site.yaml, allows and passes on just a part of the path. 

**Example site.yaml**
```yaml
redirects:
  /this-is-my-url: '/somepath'
  /this-is-sparta: '/some-other-path'
  /this: '/some-different-path'
```

With this configuration, when hitting /this, I would get redirected to /somepath because it matches on the first redirects entry. 

This happens because the current regex only checks start of string until found:

`#^\/this#`

Adding an end of string check in the regex, will not match on the first two redirects, but continue on until it matches on the last entry:

`#^\/this$#`

### DIFF
--- a/system/src/Grav/Common/Page/Pages.php
+++ b/system/src/Grav/Common/Page/Pages.php
@@ -501,7 +501,7 @@ class Pages
                     $site_redirects = $config->get("site.redirects");
                     if (is_array($site_redirects)) {
                         foreach ((array)$site_redirects as $pattern => $replace) {
-                            $pattern = '#^' . str_replace('/', '\/', ltrim($pattern, '^')) . '#';
+                            $pattern = '#^' . str_replace('/', '\/', ltrim($pattern, '^')) . '$#';
                             try {
                                 $found = preg_replace($pattern, $replace, $source_url);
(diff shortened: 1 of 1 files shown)


mahagr:
What about `/this/is/my/new/path`, it should redirect to `/some-different-path/is/my/new/path`

hughbris:
Could we have this as a config toggle though? Developers inherit sites to migrate to Grav that have all kinds of crazy URL systems. I am doing one now where having such a toggle would allow me to put in a full set of redirects. Without this, I can't see a way to make them all work :/

mahagr:
This will work:

```yaml
redirects:
  '/this-is-my-url(/.*)?$': '/somepath'
  '/this-is-sparta(/.*)?$': '/some-other-path'
  '/this(/.*)?$': '/some-different-path'
```

mahagr:
Instead of implementing the patch and breaking all redirects, you can do this instead:

```yaml
redirects:
  '/this-is-my-url$': '/somepath'
  '/this-is-sparta$': '/some-other-path'
  '/this$': '/some-different-path/?'
```

Issue #2180, opened 2018-09-12
Forcing JSON withouth *.json


mgollmer:
Hello there, 

even if I run the risk that it has been asked in the past, but how do i force JSON as content type without adding ".json" to the url?

I'm trying to develop a plugin, where I need to set up an API. Is it possible to use some method/value in the Page object?

Regards, 
Michael

Perlkonig:
[I had a PR related to this that ultimately died.](https://github.com/getgrav/grav/pull/1088) I'm now [working on fixing the arbitrary error code issue](https://github.com/getgrav/grav/issues/2164) and then I'm going to try issuing my PR again. I agree that Grav should be fully capable of understanding and responding to the `Accept` header, but currently I can't get the core to issue the appropriate `406 Not Acceptable` code.

mgollmer:
Hi @Perlkonig, 

uuhhhh seems like I asked for a complex topic.  Is there a way to force the content type inside the coding? 

I mean yeah it would be nice if Grav would react on the `Accept` attribute of the header, but isn't there any other way?

Perlkonig:
I had forgotten about that. We need to add that to the [Headers / Frontmatter page](https://learn.getgrav.org/content/headers), methinks.

rhukster:
Definitely should add it... if someone wants to add it as a PR on learn, that would help! :)

Perlkonig:
I will do my best.

Perlkonig:
Sorry. [It's already there.](https://learn.getgrav.org/content/headers#template-format) Somehow I missed it.

So, @mgollmer, if indeed you want your page to *always* return JSON, just add `template_format: json` in your frontmatter and Grav will automatically apply the appropriate `*.json.twig` template.

mgollmer:
UHHHH!!! Cool!

Thanks @rhukster! I must have scrolled over it. Just saw template, but not templateFormate. Thank you very much!

PR #2186, opened 2018-09-14
Honour ACCEPT header in the absence of other signals

Perlkonig:
**Relies on PR #2181! Don't merge unless that functionality is added.**

Related to issue #2180 and PR #1088.

# Background

There are two standard ways for an agent to tell the server what format they want in response. The first is by URL. This is currently the only signal Grav pays attention to (e.g., by adding a `.json` extension to a slug). The other standard method is via the HTTP header `Accept`, which Grav currently ignores. This PR fixes that.

# Approach

If no specific format is provided either in the page header or in the URI, then Grav looks at the `Accept` header, sorts the listed MIME types by priority, and then checks each one in order. If at any point it finds a matching template, it returns it. Otherwise it returns a `406 Not Acceptable`. (This is why PR #2181 or similar functionality is required. Otherwise all you get is a `500 Internal Server Error`.)

# Components

This PR makes three sets of changes.

1) The first is to remove the default `html` assignment in `Grav\Common\Page\Page::templateFormat`. 

   This was necessary because otherwise there was no way to know if a signal was explicitly provided by the author (in the page header) or by the agent (in the URI). I've checked where `templateFormat` gets called, and run some very basic tests, and I can't see that this change breaks anything. **I believe that this change requires the closest scrutiny.**

2) `Grav\Common\Utils` was extended to include the public function `parseAccept()` and the private support function `sortAccept()`.

   These only look at the `q` parameter and ignores all others. This is not strictly spec compliant, but it is sufficient for Grav's needs. This is a simplified version offered by @w00fz in my original PR #1088.

3) The function `Grav\Common\Twig\Twig::processSite` was altered to now pay attention to the `Accept` header.

   If a format was not explicitly stated, it goes through each MIME type in the `Accept` header in priority order. If it finds a matching template, it returns the rendered content. Otherwise it returns a `406 Not Acceptable`.

# Testing

As I've said in other PRs, I'm well aware that I do not have enough knowledge to thoroughly test all aspects of the code and any impacts to performance. I welcome any and all feedback. Thanks for your time.


mahagr:
I agree that this is a problem in Grav.

JFYI: In the long term, I have some plans to move Grav to use PSR-7 ServerRequest and PSR-15 handlers and move all the logic for this into the handlers. I think it will be the best way to go as it allows customizable behaviour based on your own needs.

rhukster:
@Perlkonig I implemented support for `accept` headers as part of another bug fix: https://github.com/getgrav/grav/commit/74cd3ac1e07ae2a0a7896c8089c815988184fc49

PR #2210, opened 2018-10-04
Register theme prefixes as namespaces in twig

WizKid:
When creating a theme that inherit from another theme one common thing is that you want to change a tiny thing in an existing twig file.

I couldn't find a good way to do that without copy the whole twig-file over to my theme. Which means that if the base theme update that twig-file I will never get the changes. And I couldn't figure out a way to reference the base theme twig file and override it at the same time.

So with this change I can change my theme config to something like this:
```
streams:
  schemes:
    theme:
      type: ReadOnlyStream
      prefixes:
        '':
          - user/themes/mytheme
          - user/themes/quark
        'quark/':
          - user/themes/quark
```

And then if I want to update the footer in base.html.twig I can just create a base.html.twig file in mytheme that looks something like this:
```
{% extends '@quark/partials/base.html.twig' %}

{% block footer %}
    My footer text
{% endblock %}
```

When I first implemented it I just had one foreach that handled the empty prefix also. But because when `$this->grav->fireEvent('onTwigTemplatePaths'); ` some stuff goes in and touches `$this->twig_paths` that didn't work.

### DIFF
--- a/system/src/Grav/Common/Twig/Twig.php
+++ b/system/src/Grav/Common/Twig/Twig.php
@@ -102,6 +102,28 @@ class Twig
 
             $this->loader = new \Twig_Loader_Filesystem($this->twig_paths);
 
+            // Register all other prefixes as namespaces in twig
+            foreach ($locator->getPaths('theme') as $prefix => $_) {
+                if ($prefix === '') {
+                    continue;
+                }
+
+                $twig_paths = [];
+
+                // handle language templates if available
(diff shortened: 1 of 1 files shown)


rhukster:
This look good to me, let me discuss with @mahagr though...

WizKid:
I mostly want some way to access the original template. Because that would make things so much easier. This was the easiest way I could figure out. So there may absolutely be some better way to do this.

PR #2235, opened 2018-10-23
[Bug Fix] - Markdown Processing for Telephone Links

ScottHamper:
Hey All,

Telephone link URLs use the `+` character to specify a [globally unique number](https://tools.ietf.org/html/rfc3966#section-5.1.4). However, when Grav parses a markdown telephone link, it replaces any `+` characters with spaces.

For example, `[Call us](tel:+15551231234)` becomes `<a href="tel: 15551231234">Call us</a>`

This replacement is happening because [`inlineLink` in `ParsedownGravTrait` calls `processLinkExcerpt`](https://github.com/getgrav/grav/blob/1.5.3/system/src/Grav/Common/Markdown/ParsedownGravTrait.php#L240), which then [uses `urldecode` on the `href` attribute](https://github.com/getgrav/grav/blob/1.5.3/system/src/Grav/Common/Helpers/Excerpts.php#L120) - `urldecode` replaces all `+` characters with spaces.

This pull request attempts to solve the issue by simply changing the call to `urldecode` to [`rawurldecode`](http://php.net/manual/en/function.rawurldecode.php). `rawurldecode` works the same as `urldecode`, except that it does not replace `+` characters.


rhukster:
thanks

PR #2357, opened 2019-01-29
Remove unused parameter

hamiltont:
UNTESTED!!

Happened to notice what looks like an unused parameter. Thought it would be nice to clean up the code a bit


mahagr:
Looks like it is unused (the variable is used directly in the method).

hamiltont:
Yup, that's what I thought as well. Granted, this is basically a no-op pull request in terms of functionality. I figured sending it in could still help the project keep the codebase clean

PR #2374, opened 2019-02-12
Theme language.yaml unable to override fixes #2372 solution 2

### LINKED ISSUE #2372: Theme language.yaml unable to override
Using the twentyfifteen skeleton I found that I am unable to override the themes languages.yaml content. I limit the site to 'en' only, so I put my overridden key in /user/languages/en.yaml (as per the docs). But my overridden key is not shown, unless I remove the original languages.yaml file (which is not the goal).

I found that this is because of the mergeRecursive function. The function is working correctly for plugin translation overrides, which are loaded early using the ConfigServiceProvider class. 
However, the Theme translations are only loaded later via the ThemesProcessor class. At that point the overriding file /user/languages/en.yaml is already loaded, and then the ThemesProcessor adds it's default translations (putting the original translations for the theme as the end translation to use).

I can think of two solutions to solve this problem quickly (I will provide pull requests):

1. Store the theme override translations in user://languages/theme.yaml and load them after the initial theme://languages.yaml file
2. Add unexisting keys from theme://languages.yaml to the language cache by adding an "inverse" parameter to the mergeRecursive function.

I am not sure which one is the best, the first one adds a new approach to overloading of translation keys for themes. The second one might have a performance impact when loading translation data from the theme

dameert:
Add unexisting keys from theme://languages.yaml to the language cache by adding an "inverse" parameter to the mergeRecursive function.

### DIFF
--- a/system/src/Grav/Common/Config/Languages.php
+++ b/system/src/Grav/Common/Config/Languages.php
@@ -13,6 +13,8 @@ use Grav\Common\Utils;
 
 class Languages extends Data
 {
+    const INVERSE = true;
+
     public function checksum($checksum = null)
     {
         if ($checksum !== null) {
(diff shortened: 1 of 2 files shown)


dameert:
Closing due to age of the PR (2year - cleanup in my PRs to follow up.

Issue #2383, opened 2019-02-19
GPM's getDependencies skips installed but disabled plugins


jgonyea:

rhukster:
no issue.. deleting

rhukster:
i guess your editing now? lol

jgonyea:
I came across an unexpected use case where I had a plugin installed, but disabled.  GPM's getDependencies only looks for missing plugins, not if any of the installed plugins are required to be enabled.

Would adding a check in this (and related functions) that a dependency is both installed and enabled make sense?

jgonyea:
Right now, the return actions are 
* ignore
* install
* update

I propose there be another option "enable", meaning that the plugin is installed, but not currently enabled.

PR #2386, opened 2019-02-21
Dependency fix

jgonyea:
An initial stab at ensuring that already installed (but disabled) dependencies are enabled.

If you feel like I should split out the enabling to its own command file, I can refactor and resubmit.

This PR also contains some PSR2 fixes.

### DIFF
--- a/system/src/Grav/Console/Gpm/InstallCommand.php
+++ b/system/src/Grav/Console/Gpm/InstallCommand.php
@@ -60,6 +60,7 @@ class InstallCommand extends ConsoleCommand
     {
         $this
             ->setName("install")
+            //->alias("enable")
             ->addOption(
                 'force',
                 'f',
@@ -311,13 +312,10 @@ class InstallCommand extends ConsoleCommand
 
             if ($answer) {
                 foreach ($packages as $dependencyName => $dependencyVersion) {
-                    
                     $package = $gpm->findPackage($dependencyName);
                     if ($package) {
-                        $package_array = $package->toArray();
-                    
-                        // Sanity check. Only plugins should be allowed to be enabled.
-                        if ($package_array['package_type'] != 'plugins'){
+                       // Sanity check. Only plugins should be allowed to be enabled.
+                        if ($package->package_type != 'plugins') {
                             $message = $dependencyName . " was not enabled.  It is not a plugin.";
                             $this->output->writeln($message);
                             continue;
@@ -326,8 +324,7 @@ class InstallCommand extends ConsoleCommand
                     if ($dependencyVersion == 'enable') {
                         if (InstallCommand::enablePackage($dependencyName)) {
                             $this->output->writeln('  |- Enabled <cyan>' . $dependencyName . '</cyan>');
-                        }
-                        else {
+                        } else {
                             $this->output->writeln('  |- Not Enabled <red>' . $dependencyName . '</red>. See the log file for more information');
                             throw new \Exception();
                         }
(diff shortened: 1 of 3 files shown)


jgonyea:
Forgot to mention that this is related to Issue #2383 https://github.com/getgrav/grav/issues/2383

OleVik:
Just to clarify; the change enforces an enabled-state for all plugins when upgrading/updating?

If so, I wholeheartedly agree that it is a necessary constraint. Otherwise, upgrading/updating can potentially push versions beyond the point where a disabled plugin can operate. The disabled-state may only be temporary, and so upgrading can break functionality without the possibility of a downgrade to the previous ecosystem of working versions.

system/src/Grav/Console/Gpm/InstallCommand.php
```diff
@@ -322,6 +345,42 @@ public function installDependencies($dependencies, $type, $message, $required =
...
+     *
+     * @param array  $package_slug   The package to enable.
+     *
+     * @return bool Success of enabling a package.
+     * 
+     * @throws \Exception
+     */
+    public static function enablePackage($package_slug)
+    {
+        $grav = new Grav;
```
w00fz:
```suggestion
        $grav = new Grav;
```

system/src/Grav/Console/Gpm/InstallCommand.php
```diff
@@ -322,6 +345,42 @@ public function installDependencies($dependencies, $type, $message, $required =
...
+
+        $config_file_path = USER_DIR . "config" . DS . "plugins" . DS;
+        $config_filename = $package_slug . '.yaml';
+        $config_file = File::instance($config_file_path . $config_filename);
+        
+        if (!is_writable($config_file->filename())) {
+            $message = "Could not save config for " . $package_slug . ". Is the config folder/ file writable?";
+            $grav::instance()['log']->error($message);
```
w00fz:
```suggestion
            $grav['log]->error($message);
```

system/src/Grav/Console/Gpm/InstallCommand.php
```diff
@@ -322,6 +345,42 @@ public function installDependencies($dependencies, $type, $message, $required =
...
+        if ($config_file->exists()) {
+            // Load existing config file.
+            $config_data = Yaml::parse($config_file->content());
+        }
+
+        // Set 'enable' plugin option.
+        $config_data['enabled'] = true;
+
+        $config_file->save(Yaml::dump($config_data));
+        $grav::instance()['log']->info($package_slug . " was enabled.");
```
w00fz:
```suggestion
        $grav['log']->info($package_slug . " was enabled.");
```

jgonyea:
This only forces an enable of Plugin A if: 

1. Plugin A is installed, but in a disabled state
2. Another plugin/ theme (Package B) is installed/ updated via GPM that depends on that disabled Plugin A


I didn't go as far as forcing a enabled yaml file for even the installed Package B.

system/src/Grav/Console/Gpm/InstallCommand.php
```diff
@@ -322,6 +345,42 @@ public function installDependencies($dependencies, $type, $message, $required =
...
+            return false;
+        }
+
+        if ($config_file->exists()) {
+            // Load existing config file.
+            $config_data = Yaml::parse($config_file->content());
+        }
+
+        // Set 'enable' plugin option.
+        $config_data['enabled'] = true;
```
w00fz: This could fail if `$config_file` doesn't exist?

jgonyea: No, if I change L366 to:
`if (!$config_file->writable($config_file->filename())) {`

jgonyea: To step through the steps just prior:
if the config file is present, then the `Yaml::parse` loads its contents into the $config_data array.
Whether or not the config file isn't present, then this L377 will just create a new array (or edit the existing array) at $config_data with the enable key setting it to `true`

jgonyea:
Newest commit now includes removal of a package's config yaml on uninstall.

jgonyea:
I'm closing this pull request and will resubmit a better, updated version of it.

PR #2404, opened 2019-03-13
Preliminary tests for CsvFormatter::encode + minor fix for empty data

belgattitude:
Just adding a test for CsvFormatter and allowing empty data


belgattitude:
Haven't changed the original behaviour...

One thing that can be discussed, do we display header if they actually are numeric ? (I think for now yes, this is enforced by the test, but can be changed later)

PR #2405, opened 2019-03-13
Preliminary tests for CsvFormatter::encode + minor fix for empty data

belgattitude:
Haven't changed the original behaviour...

One thing that can be discussed, do we display header if they actually are numeric ? (I think for now yes, this is enforced by the test, but can be changed later)


mahagr:
Good catch that empty test. 

BTW: If you look at the decode(), it assumes that the first line has the headers. :)

PR #2465, opened 2019-04-19
Add `getSubnet` util to Grav\Common\Utils

Vivalldi:
Added `getSubnet` util. To be used by Login Plugin for getgrav/grav-plugin-login#204


rhukster:
Thanks merged both and updated blueprints/changelog etc.

Vivalldi:
I didn't mention it in the inline docs but the function supports both IPv4 and IPv6

PR #2466, opened 2019-04-20
Use Null Coalesce Operator

carusogabriel:
Just a small refactoring 💅


mahagr:
Nope, it was not refactoring, it actually broke the code. You cannot replace `array_key_exists()` call with `??`, the meaning of the code changes.

mahagr:
Reverted the code.

carusogabriel:
@managr Maybe we should add a test than?

PR #2496, opened 2019-05-08
Update pages/01.home/default.md

Wintertown:
Correct an error that ignored the "Typography" menu item in the default Grav installation and caused a discrepancy between the install page and the corresponding tutorial on getgrav/grav-learn. Related to PR https://github.com/getgrav/grav-learn/pull/681 and solves issue https://github.com/getgrav/grav-learn/issues/650.


rhukster:
so many places to update! thanks.

PR #2512, opened 2019-05-22
Fix session with multilanguage

### LINKED ISSUE #2510: Utils::pathPrefixedByLanguageCode doesn't detect homepage with prefix without trailing slash
A URL like `/en`, without trailing slash (which you get automatically from the initial language redirect and also from the language switch plugin) will not be detected by `Utils::pathPrefixedByLanguageCode`. `/en/` works as expected.

### LINKED ISSUE #2511: Presence of language prefix in URL incorrectly assigns admin session
I discovered the following strange behavior on a multi-language page with admin plugin:

* After logging into the admin, my username would appear in the navbar on the page, but only sometimes. Further investigation showed that it would be hidden on the homepage with language prefix without trailing slash (e.g. `/en`) but visible with the trailing slash (e.g. `/en/`) or on sub-pages (e.g. `/en/blog`). This particular symptom is actually because of #2510 in combination with this bug.
* The "save language in session" feature would not work for me (after I "fixed" the previous issue using a htaccess redirect from `/en` to `/en/` - I hadn't fully understood the problem yet). Further investigation showed that on `/` a different session would be used than on `/en/`/`/de/`, and hence the new language would be written to session A but `/` would read from session B and redirect to the wrong language.

Turns out that this is due to a bug in `SessionServiceProvider.php`: It sets `$is_admin = true` if the route starts with `/admin` _or with a language prefix_. It should instead check if it starts with `/<prefix>/admin` in that case!

CherryDT:
Closes #2510
Closes #2511

### DIFF
--- a/system/src/Grav/Common/Utils.php
+++ b/system/src/Grav/Common/Utils.php
@@ -980,7 +980,7 @@ abstract class Utils
      */
     public static function pathPrefixedByLangCode($string)
     {
-        if (strlen($string) <= 3) {
+        if (strlen($string) < 3) {
             return false;
         }
 
--- a/system/src/Grav/Common/Service/SessionServiceProvider.php
+++ b/system/src/Grav/Common/Service/SessionServiceProvider.php
@@ -56,7 +56,9 @@ class SessionServiceProvider implements ServiceProviderInterface
                 $current_route = str_replace(Uri::filterPath($uri->rootUrl(false)), '', parse_url($uri->url(true), PHP_URL_PATH));
 
                 // Check no language, simple language prefix (en) and region specific language prefix (en-US).
-                if (Utils::startsWith($current_route, $base) || Utils::pathPrefixedByLangCode($current_route)) {
+                if (Utils::pathPrefixedByLangCode($current_route)) $current_route = preg_replace('/^\/[^\/]+\b/', '', $current_route);
+
+                if (Utils::startsWith($current_route, $base)) {
                     $cookie_lifetime = $config->get('plugins.admin.session.timeout', 1800);
                     $enabled = $is_admin = true;
                 }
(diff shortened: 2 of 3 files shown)


system/src/Grav/Common/Utils.php
```diff
@@ -980,7 +980,7 @@ public static function arrayUnflattenDotNotation($array, $separator = '.')
      */
     public static function pathPrefixedByLangCode($string)
     {
-        if (strlen($string) <= 3) {
+        if (strlen($string) < 3) {
```
mahagr: Language prefix can have more than 2 letters in it.

CherryDT: Yes, that didn't change.

CherryDT: What change are you suggesting? The original logic was that the URL must have at least 4 characters to be even considered, and I changed it to 3 to match `/en` too instead of only `/en/`... None of this changes anything for prefixes that are longer, since those were not short-circuit-rejected in the first place... (They would lead an URL longer than 3 characters)

CherryDT: It occurs to me now that you may have looked at this change out of context. The full code is:

```php
if (strlen($string) < 3) {
  return false;
}
```

rhukster:
I've fixed this in a more readable way.  See the issues, as they are linked.

CherryDT:
Thanks but I'm not sure I understand how it is supposed to work, please check my comments in the other two issues.

In short:
* For #2510 I don't see any fix in your commit at all (and no added tests for the fix)
* For #2511 I am scratching my head how `$lang_admin_base = '/' . $lang . $admin_base;` will work if `$lang` is simply `true`/`false` as it is returned from `Utils::pathPrefixedByLangCode` - wouldn't it then check for `/admin` in the `false` case (that may be intended) but `/1admin` in the `true` case?

system/src/Grav/Common/Utils.php
rhukster: i actually removed this completely, as languages are not restricted in their length now, can be anything really.

PR #2541, opened 2019-06-06
run route.php when grav is not at the root of the server

aoloe:
Currently, if I want to use the `router.php` provided by Grav I need to start `php -S` in the Grav's root folder.

If on the production server, Grav is not installed at the documents root, this will make it slightly more complex to manage the assets.

This patch adds the option to start Grav as:

```sh
GRAV_BASEDIR="/grav" php -S localhost:8000 grav/system/router.php
```

(in this case it will also accept "grav" and "grav/" as the basedir values)

### DIFF
--- a/system/router.php
+++ b/system/router.php
@@ -17,11 +17,25 @@ if (is_file($_SERVER['DOCUMENT_ROOT'] . DIRECTORY_SEPARATOR . $_SERVER['SCRIPT_N
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


aoloe:
coucou...

rhukster:
I'm going to optimize this a bit for readibility, but functionality will remain unchanged.

aoloe:
ok, thanks for merging!

PR #2551, opened 2019-06-14
Fix nginx config for ddev

pquerner:
In ddev, it is required to have a healthcheck available. 
This commit adds that functionality according to the latest ddev docs.

(ref https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/)


jgonyea:
I'm sorry, I didnlt notice your pull request and I ended up making my own (which was just merged). Can you confirm the latest code does the health check properly? I think it's done via the inclusion of the monitoring.conf file.

PR #2621, opened 2019-08-13
Check if GRAV_USER_INSTANCE constants are already defined

### LINKED ISSUE #2620: Constant GRAV_USER_INSTANCE already defined
When trying to write Unit Tests for my plugin, following the Codeception/`_bootstrap.php` approach like is done in the Grav repo, if I try and get a user before every test:

```php
protected function _before()
{
    $grav = Fixtures::get('grav');
    $this->grav = $grav();
    $this->user = $this->grav['accounts']->load('testuser');
}
```

Then it throws an error: 
```
Constant GRAV_USER_INSTANCE already defined
```

I did a little digging, and it seems that the [AccountsServiceProvider.php](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Service/AccountsServiceProvider.php#L49-L57) is blindly defining a constant and not checking if it first exists. 

This is fine when your PHP process instantiates Grav once, but when trying to write unit tests and reinstantiating Grav multiple times, it means _only the first_ test will pass and all the rest will throw an error due to `$this->grav['accounts']->load` in the `_before()` function trying to redefine a constant.

I believe a simple fix would be wrapping them in a check within `AccountsServiceProvider.php` like so:

```php
if (!defined('GRAV_USER_INSTANCE')) {
    define('GRAV_USER_INSTANCE', 'DATA');
}
```

Regaez:
Closes #2620 

Small check to allow accounts to prevent a constant being defined again. This allows `$grav['accounts']`  to be accessed multiple times when new `Grav::instance()` are created, for example when unit testing.

### DIFF
--- a/system/src/Grav/Common/Service/AccountsServiceProvider.php
+++ b/system/src/Grav/Common/Service/AccountsServiceProvider.php
@@ -46,7 +46,9 @@ class AccountsServiceProvider implements ServiceProviderInterface
 
     protected function dataAccounts(Container $container)
     {
-        define('GRAV_USER_INSTANCE', 'DATA');
+        if (!defined('GRAV_USER_INSTANCE')) {
+            define('GRAV_USER_INSTANCE', 'DATA');
+        }
 
         // Use User class for backwards compatibility.
         return new DataUser\UserCollection(User::class);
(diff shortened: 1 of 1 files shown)


Regaez:
Hmm. The CI build failed, but it seems unrelated to my change?

mahagr:
Just a note on this: you still cannot easily switch between those two; you are stuck on the account type you were using the first time.

The issue is that I had to make `Grav\Common\User` class to behave just like the old class was.

PR #2638, opened 2019-08-19
[RFR] Fix parsedown vulnerabilities

jnadaud:
Fix Parsedown current vulnerabilities:

 * [CVE-2018-1000162][]: Cross-Site Scripting
 * [CVE-2019-10905][]: Class-Name Injection


rhukster:
Parsedown 1.7 is not backwards compatible and as such plugins that provide markdown functionality will break.  This is not an option for us in the current version of Grav.  We're looking to move to latest version of Parsedown in Grav 2.0 when we can break compatibility somewhat.

PR #2679, opened 2019-10-02
Adjusted method call for firing an event, Symfony 4.3 Support

jackthomasatl:
This PR is dependent on an additional pull request against rockettheme/toolbox being merged.

https://github.com/rockettheme/toolbox/pull/27

Additionally, once rockettheme cuts a new version we'll need to update the composer.json here.

### DIFF
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -362,7 +362,7 @@ class Grav extends Container
         /** @var EventDispatcher $events */
         $events = $this['events'];
 
-        return $events->dispatch($eventName, $event);
+        return $events->dispatch($event, $eventName);
     }
 
     /**


rhukster:
Seems your change fails in our automated tests. Please review the Travis output or run the built in tests on your fork `bin/grav test`

jackthomasatl:
hey @rhukster, 

Please see rockettheme/toolbox#27

I'm using Symfony 4.3 with Grav to have controlllers /routing and the rollback to Symfony 4.2 breaks our code base.

I migrated the method call for FireEvent to the newer Symfony 4.3 header, But this also relies on the above PR.

mahagr:
Symfony 4.3 supports the old ordering of the fields, which makes upgrading both Toolbox and Grav totally unnecessary at this point. We just cannot break all the code because of Symfony now supports  [PSR-14](https://www.php-fig.org/psr/psr-14/) for the events and are deprecating the old behaviour.

It does not mean that I've not looked into this already and had plans to upgrade events, but for now because of all the other tasks I've decided to delay this new way until we get a bit closer to 2.0, which is not that far away.

In any case, if you read PSR-14, you will see that events will need to be fully redesigned to make the move, which is why I have a hard time accepting your changes. They are trying to use the new event system in the old way.

For PSR-compatible events `dispatch()` needs to take Event object and the object only as of the parameter -- event object itself acts as the event name. Also, there is no interface for the event class, you can use whichever class you want to. This makes Toolbox Event classes deprecated (read not needed anymore) and basically forces us to change how events work.

jackthomasatl:
Understood. Would we possibly move forward with using Symfony EventDispatcher or is the idea to stick with rockettheme/toolbox ?

mahagr:
Well, we are actually already using Symfony, though we have extended it slightly to add some extra features. But basically what I was trying to say is that in the future any class can be used as an Event, you don't need to implement any interface or extend any class to do it. Then you just hook to listen for events using the class you're passing.

PR #2689, opened 2019-10-11
Update Core to follow PSR-2 standard

PatricNox:
A mass change to fix indentations, unnecessary spaces and inconsistency.


mahagr:
This change needs to be against 1.7 branch as it has the most recent code with a lot of small fixes in almost every file, likely some of yours as well.

Otherwise I would be happy to see these code style fixes in.

PR #2699, opened 2019-10-16
Add a .phan config file and stubs for Phan static analysis

rlerdorf:
I fixed some small inconsistencies here and there. There are still
a little over 200 to go. This time against the 1.7 branch.

To try it, checkout this PR and run:
composer require phan/phan
It doesn't have to be in the Grav tree. You can install it anywhere.
Then from the top-level Grav dir:
/vendor/bin/phan -p

You can see the ones that are left to address here:
https://gist.github.com/69eac9b37ced1cadc08ed4be0ee84f40

Also, and the main reason I added it, was to get some nice
dependency graphs from Phan to help me understand the Grav code. Some
examples:
http://pdep.lerdorf.com/?mode=class&node=\Grav\Common\File\CompiledFile&d=3
http://pdep.lerdorf.com/?mode=class&node=\Grav\Common\Filesystem\Folder&d=1
http://pdep.lerdorf.com/?mode=class&node=\Doctrine\Common\Cache\Cache&d=2
http://pdep.lerdorf.com/?mode=class&node=\RocketTheme\Toolbox\ArrayTraits\Export&d=1

Move the slider to change the number of dependency levels shown.

### DIFF
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -47,13 +47,15 @@ class Grav extends Container
     public $output;
 
     /**
-     * @var static The singleton instance
+     * @var ?static The singleton instance
      */
     protected static $instance;
 
     /**
      * @var array Contains all Services and ServicesProviders that are mapped
      *            to the dependency injection container.
+     *
+     * @suppress PhanPluginMixedKeyNoKey
      */
     protected static $diMap = [
         'Grav\Common\Service\AccountsServiceProvider',
@@ -121,7 +123,7 @@ class Grav extends Container
      */
     public static function instance(array $values = [])
     {
-        if (!self::$instance) {
+        if (empty(self::$instance)) {
             self::$instance = static::load($values);
         } elseif ($values) {
             $instance = self::$instance;
@@ -153,9 +155,6 @@ class Grav extends Container
         if ($environment) {
             Setup::$environment = $environment;
         }
-
-        $this['setup'];
-        $this['streams'];
 
         return $this;
     }
@@ -376,7 +375,7 @@ class Grav extends Container
      * Fires an event with optional parameters.
      *
      * @param  string $eventName
-     * @param  Event  $event
+     * @param  ?Event $event
      *
      * @return Event
      */
--- a/system/src/Grav/Framework/Form/Interfaces/FormInterface.php
+++ b/system/src/Grav/Framework/Form/Interfaces/FormInterface.php
@@ -131,14 +131,14 @@ interface FormInterface extends RenderInterface, \Serializable
 
     /**
      * @param ServerRequestInterface $request
-     * @return $this
+     * @return FormInterface
      */
     public function handleRequest(ServerRequestInterface $request): FormInterface;
 
     /**
      * @param array $data
-     * @param UploadedFileInterface[] $files
-     * @return $this
+     * @param ?UploadedFileInterface[] $files
+     * @return FormInterface
      */
     public function submit(array $data, array $files = null): FormInterface;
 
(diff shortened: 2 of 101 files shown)


system/src/Grav/Framework/Form/Interfaces/FormInterface.php
```diff
@@ -131,14 +131,14 @@ public function getFlash();
 
     /**
      * @param ServerRequestInterface $request
-     * @return $this
```
mahagr: $this was actually the correct return value, but current PHP versions do not support covariant return values.

mahagr: Same applies to all the later ones, too. The covariance will be supported in PHP 7.3 and we should add those type hints where ever we will need to change those in the future.

system/src/Grav/Framework/Form/Traits/FormTrait.php
```diff
@@ -408,6 +409,7 @@ public function getAllFlashes(): array
     /**
      * {@inheritdoc}
      * @see FormInterface::render()
+     * @suppress PhanAccessMethodInternal - Twig render() is marked as internal
```
mahagr: This is not internal but our own implementation which has nothing to do with Twig.

system/src/Grav/Framework/Object/Property/ArrayPropertyTrait.php
```diff
@@ -50,7 +50,7 @@ protected function doHasProperty($property)
      * @param bool $doCreate        Set true to create variable.
      * @return mixed                Property value.
      */
-    protected function &doGetProperty($property, $default = null, $doCreate = false)
```
mahagr: Should be never public.

mahagr:
We are already using phpstan for code analysis and I still have over 700 issues to go (to level 7). It is a huge job to figure out how everything should be fixed and this work started on Grav 1.5 release.

Some of the errors are not bugs and not all the real bugs are easy to fix either, especially with the older code.

rlerdorf:
True, some of these came from the switch from the development branch to the 1.7 branch and are no longer applicable in 1.7. The majority of them should still be fixed though. There are a lot of missing nullables in the phpdoc types and some outright bugs.

system/src/Grav/Framework/Form/Interfaces/FormInterface.php
rlerdorf: Covariant returns are not until PHP 7.4 so at this point you can't return a more specific type from an interface method like this.

mahagr: I meant PHP 7.4... I must have misspelt there or perhaps I had a brain-fart.

Nevertheless, I can for sure return more specific type as it is very useful to use covariance there even if PHP does not yet support it as a return type. This allows me to access all the methods from my more specific class and it does work with phpstorm and phpstan. Meaning that I get a full list of the methods available as well as checks against bad/deprecated method calls.

mahagr:
I know every issue should be fixed and I've been finding bugs myself as well. Unfortunately, I have also introduced a few bugs by fixing these reported issues.

I am going to be away the rest of the week, so I will take another look at this later.

system/src/Grav/Common/Grav.php
```diff
@@ -121,7 +123,7 @@ public static function resetInstance()
      */
     public static function instance(array $values = [])
     {
-        if (!self::$instance) {
+        if (empty(self::$instance)) {
```
mahagr: should be `null !== self::$instance`

mahagr:
I reverted many of the changes as
- there was a fatal error in PHP 7.1 (incompatible function definition)
- many of those changed the logic and badly broke Grav
- some unit tests were failing because of the changes
- some changes to the interfaces which may be used in plugins
- I don't want to silence the errors in static tools, at least not from the code which may hide bugs

In short: it is really dangerous to make these kinds of changes without knowing the code and doing a throughout testing afterwards. We are fixing those ourselves, but it is much harder than it may look like because of the unintended side effects that the code changes may have.

After making these changes, it looks like we're down to 650 errors in phpstan tool, too.

PR #2709, opened 2019-10-27
Instantiate parent in Page->translatedLanguages()

### LINKED ISSUE #2163: Page::translatedLanguages(): contains slugs or routes
The documentation for Page::translatedLanguages() states:
> Return an array with the routes of other translated languages

However, the function returns an array with a **route** when the page has a default route defined, else it returns the **slug**. Which are quite different imho...

For example, let's assume we have the following folder structure:
```
.../pages/
     01.my/
        01.example/
           01.page/
```
- When the page has a default route defined, it returns the route.
With the following definition in frontmatter...
  ```
  routes:
    default: '/my/example/other-page'
  ```
  ...it returns the array: 
  ```
  [ "en" => "/my/example/other-page"]
- When no default route has been defined, it returns the slug for the page:
  ```
  [ "en" => "page"]
  ```

**Consequence**:
When looping through the results of $page->translatedPages() the different "route" types lead to unexpected results.
```
$translatedLanguages = $page->translatedLanguages();
foreach ($translatedLanguages as $language => $route) {
    // process $route ;
}
```

OleVik:
Fixes #2163.

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -214,6 +214,7 @@ class Page implements PageInterface
             if ($exists) {
                 $aPage = new Page();
                 $aPage->init(new \SplFileInfo($path), $languageExtension);
+                $aPage->parent($grav['page']->parent());
 
                 $route = $aPage->header()->routes['default'] ?? $aPage->rawRoute();
                 if (!$route) {
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -215,6 +215,11 @@ class Page implements PageInterface
                 $aPage = new Page();
                 $aPage->init(new \SplFileInfo($path), $languageExtension);
                 $aPage->parent($grav['page']->parent());
+                if (file_exists($aPage->parent()->path() . DS . $aPage->parent()->template() . $languageExtension)) {
+                    $aPage->parent()->route(
+                        $language->getLanguageURLPrefix($languageCode) . $aPage->parent()->route()
+                    );
+                }
 
                 $route = $aPage->header()->routes['default'] ?? $aPage->rawRoute();
                 if (!$route) {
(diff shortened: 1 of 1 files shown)


mahagr:
A similar fix is probably needed for Flex Pages.

mahagr:
@OleVik There's a small issue there; parent isn't the correct one as it comes from the current language, not from the same language as the page.

OleVik:
@mahagr I added a check for a translated version of the parent, in the same language as the current Page, and set the parent's route to reflect that. I also updated the returned route to include the language-prefix, otherwise all languages return the same route.

mahagr:
It still won't work as you can change slug from the parent based on the language. So the URL is basically going to be wrong: `/en/dogs/intro` -> `/fi/koirat/esittely`. With your change, the resulting URL would be: `/fi/dogs/esittely` which is still wrong.

The only way to get it right, is to go up on a tree and initialize all the translated parent pages and get route from them.

OleVik:
That's derived from aliases, right? Wouldn't the router then apply `/fi/koirat/esittely` if `/fi/dogs/intro` was requested? I'm trying to avoid recursion upwards, assigning new routes along the way, if possible - to not impact performance as much.

mahagr:
I think the best thing though, would be to get the `rawRoute()` instead of trying trying to build translated parent route (it won't work with your code). If you use rawRoute(), it will never be translated (it uses folders in the filesystem to name the route) and that route is always added to the router and known to work.

The issue with your approach is that it'll not build the nested route right way causing 404 in many cases as the route cannot be found.

mahagr:
No, it won't work like that as Grav currently keeps record only from the current translation. This means that all the other slugs are coming from the wrong language.

OleVik:
What associates `/en/dogs/intro` with `/fi/koirat/esittely` though? Seemingly the slug, per [Multi-Language Routing](https://learn.getgrav.org/16/content/multi-language#multi-language-routing). I probably need to expand my test-case, I initially just needed a fix for `$this->grav['page']->translatedLanguages(true)` - which just returns empty values right now.

rhukster:
Each language, no matter the language specific slugs, are associated by Lang and raw route.

OleVik:
But do they then belong to different folders, or does the Page FrontMatter just declare the translation? I haven't really had need to use multi-language that much in Grav before, so my test-cases were always rather simple.

PR #2721, opened 2019-11-07
Fix rewrite rule for lighttpd config

gatesphere:
Just plain doesn't work without this!  Was probably omitted in error.


JopoJP:
Important. Even if lighttpd isn't used a lot, it should rewrite it correctly.

PR #2731, opened 2019-11-14
fix.(cache): issue cache key is null

darkiron:
Prevent Error when cache key doesn't exist


mahagr:
Cannot really do that. The pages will conflict with each other causing unexpected results.

darkiron:
how to resolve 500 internal when function return null ?

PR #2770, opened 2019-12-20
Fix for Saving Flex Pages on Windows

SeriousKen:
This fixes an issue with Flex Pages on Windows where dirname returns backslashes and prevents the page from being saved.


mahagr:
I made a better fix in https://github.com/getgrav/grav/commit/608457bd010e6df853451f7ecf48da72bb2868b8, thank you for pointing me out the weird behaviour of dirname() method.

PR #2780, opened 2020-01-02
Update $date_formats

petira:
Added seconds and new default formats (used in the Middle Europe) into $date_formats. However, I still prefer the ability to insert (and save) my own default date format via the Admin Panel. More information here: https://github.com/getgrav/grav-plugin-auto-date/pull/7#issue-358484215


rhukster:
You have changed the default keys. For example: `d-m-Y H:i` is now `d-m-Y H:i:s`.  This means people who have `d-m-Y H:i` set will not result in a valid format being found, and will use default. This will break existing behavior.

I think a plugin, or another approach to allow adding custom date formats would be better than changing this hardcoded list.

petira:
Hi Andy, I don't think it would affect existing behavior. If I am not mistaken, these are only non-indexed lists and the current item is stored in `system.yaml`.

For now, I am making the change directly in `system.yaml`, but I am the only administrator. If the site is managed by multiple users, there is a risk that someone will change it through the `Admin Panel`, and can only enter some of the default values.

I conclude that `seconds` are an important part of the `date` and should be in the default format.

PR #2788, opened 2020-01-08
Added twig extension for preg_match and preg_split

thexmanxyz:
I added a twig extension for the PHP methods `preg_split` and `preg_match` (including code documentation) to cover those aspects not only within Gantry but also in Grav as well. For more information please also check this FR (gantry/gantry5#2586) and this PR (gantry/gantry5#2590).

### DIFF
--- a/system/src/Grav/Common/Twig/TwigExtension.php
+++ b/system/src/Grav/Common/Twig/TwigExtension.php
@@ -157,6 +157,8 @@ class TwigExtension extends \Twig_Extension implements \Twig_Extension_GlobalsIn
             new \Twig_SimpleFunction('repeat', [$this, 'repeatFunc']),
             new \Twig_SimpleFunction('regex_replace', [$this, 'regexReplace']),
             new \Twig_SimpleFunction('regex_filter', [$this, 'regexFilter']),
+            new \Twig_SimpleFunction('preg_match', [$this, 'pregMatch']),
+            new \Twig_SimpleFunction('preg_split', [$this, 'pregSplit']),
             new \Twig_SimpleFunction('string', [$this, 'stringFunc']),
             new \Twig_SimpleFunction('url', [$this, 'urlFunc']),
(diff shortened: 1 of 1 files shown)


system/src/Grav/Common/Twig/TwigExtension.php
```diff
@@ -157,6 +157,8 @@ public function getFunctions()
             new \Twig_SimpleFunction('repeat', [$this, 'repeatFunc']),
             new \Twig_SimpleFunction('regex_replace', [$this, 'regexReplace']),
             new \Twig_SimpleFunction('regex_filter', [$this, 'regexFilter']),
+            new \Twig_SimpleFunction('preg_match', [$this, 'pregMatch']),
```
mahagr: Cannot we just call the method directly:

```php
new \Twig_SimpleFunction('preg_match', 'preg_match')
```

thexmanxyz: @mahagr The `preg_match` is taken directly the same from Gantry5 in the form it is currently in this PR request (without `offset` and `flags` parameter). TBH IDK if incompatibilities might occur if we have different method signatures. Same for the the `preg_split` but only with the `flags` parameter. I decided to also omit the `flags` parameter as it was skipped for `preg_match` within Gantry5 as well.

rhukster: @mahagr you probably know best about the code in G5, but it seems that `false` return value is there for a reason in G5.

mahagr: Oh, right. The `preg_match()` PHP method returns true/false by default and puts the matching strings into the third parameter. It looks like @newkind added this method back in 2015, so I guess it's being used in a few themes.

If this was a filter, I would agree on the returned value, but in the function, not sure. I guess that you cannot update values inside parameters in twig, so it would make sense to remove the third parameter altogether -- it's being returned anyway.

mahagr:
Merged but renamed the methods to match the other two.

PR #2791, opened 2020-01-16
Replaced Whoops deprecated method

Volga:
pushHandler -> prependHandler


mahagr:
This has already been fixed in Grav 1.7.

PR #2792, opened 2020-01-16
Fixed phpDoc missing return statement

Volga:


mahagr:
This has already been fixed in Grav 1.7.

PR #2820, opened 2020-02-09
Temporarily fix for using multisites  advanced

AkkuFresh:
There are 2 way to setup multisites in Grav CMS. 
The first way: with different plugins set for every sites - it was supported already with earlier temporarily fix
The second way: COMMON plugins set for all domains in multisite was unsupported.
With this modification you can use a stream definition in your setup.php like this
            'pluginupdates' => [
                'type' => 'ReadOnlyStream',
                'paths' => [
                    'user/plugins',
                ],
                'prefixes' => [
                    '' => ['user/plugins'],
                ]
            ],
And if this stream definition exists, the plugin install path will not be modified and will be suitable for the common use in 'user/plugins' folder.

### DIFF
--- a/system/src/Grav/Common/GPM/Remote/AbstractPackageCollection.php
+++ b/system/src/Grav/Common/GPM/Remote/AbstractPackageCollection.php
@@ -54,11 +54,15 @@ class AbstractPackageCollection extends BaseCollection
         $this->raw        = $this->cache->fetch(md5($this->repository));
 
         $this->fetch($refresh, $callback);
+		//Get schemes defined in setup.php -> To make possiblity to use different multisite setups
+		$schemes = Grav::instance()['locator']->getSchemes();
         foreach (json_decode($this->raw, true) as $slug => $data) {
             // Temporarily fix for using multisites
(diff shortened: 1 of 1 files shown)


mahagr:
You can already do this by defining `plugins://` and `plugin://` stream not to use `user://` stream.

AkkuFresh:
@mahagr I am sorry, you completely misunderstood. The point of this modification was not to create a new stream, but to make the hardcoded plugin install folder variable depending on the multisite settings. Do you see the suggested code?

AkkuFresh:
If this part is not handled, from line 57, the line 61 will overwrite the plugin install folder and you cannot set it up form the setup.php
```
foreach (json_decode($this->raw, true) as $slug => $data) {
            // Temporarily fix for using multisites
            if (isset($data['install_path'])) {
                $path = preg_replace('~^user/~i', 'user://', $data['install_path']);
                $data['install_path'] = Grav::instance()['locator']->findResource($path, false, true);
            }
            $this->items[$slug] = new Package($data, $this->type);
        }
```

PR #2891, opened 2020-04-24
Support for the X-Forwarded-Host

### LINKED ISSUE #2877: Handle HTTP_X_FORWARDED_HOST variable
I wonder why `HTTP_X_FORWARDED_HOST` (from the X-Forwarded-Host header) is not handled. This (unofficial) header is sent by reverse proxies to tell the "real" address for the website.

Is this because it might be used to poison one's cache? (In case it's not vetted by the reverse proxy).

As a matter of fact, other variables of this kind are used ( `HTTP_X_FORWARDED_PORT` ).

Fixing this is pretty easy it seems. In `system\src\Grav\Common\Uri.php`, line 1172:

```php
        if (isset($env['HTTP_X_FORWARDED_HOST'])) {
            $hostname = $env['HTTP_X_FORWARDED_HOST'];
        } else if (isset($env['HTTP_HOST'])) {
 ```

I can do a PR if needed.

In the meantime, I can use the `custom_base_url` options, which works.

### PR DESCRIPTION
**Pull request by A----**
Fixes #2877

### DIFF
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -1169,7 +1169,9 @@ class Uri
 
         // Build host.
         $hostname = 'localhost';
-        if (isset($env['HTTP_HOST'])) {
+        if (isset($env['HTTP_X_FORWARDED_HOST'])) {
+            $hostname = $env['HTTP_X_FORWARDED_HOST'];
+        } else if (isset($env['HTTP_HOST'])) {
             $hostname = $env['HTTP_HOST'];
         } elseif (isset($env['SERVER_NAME'])) {
             $hostname = $env['SERVER_NAME'];

### CODE REVIEW

**Comment by drzraf** [comment]
- Contrary to `SERVER_NAME` which is server provided,
- ... and `HTTP_HOST` which is user-provided but often defacto filtered-down by httpd configuration,
- `X_FORWARDED_HOST` is under full control of the visitors (except in the rare situations where httpd is configured to deal with this header)

Is user-control of the `$hostname` variable (and the implication for callers) safe?

**Comment by mahagr** [comment]
We are filtering hostname for illegal letters. We are using host internally to load the correct site configuration and checking against referrer if it's coming from the same site. Maybe something else as well?

**Comment by drzraf** [comment]
Sorry, to worry, but I'm trying to figure-out an adversary use of `X-Forwarded-Host: 127.0.0.1` (valid value) or a way to inventory multi-sites setup or just tweak some functions' behavior.

There are quite a lot of references to `new Uri()` or `Uri::getCurrentUri()` and some of them make actual use of the `$host` component for possibly critical guesses.

Just one situation as an example (significant or not?):
```php
// system/src/Grav/Common/Page/Markdown/Excerpts.php : processImageExcerpt()
 $local_file = isset($url_parts['path'])
                && (empty($url_parts['scheme']) || in_array($url_parts['scheme'], ['http', 'https'], true))
                && (empty($url_parts['host']) || $url_parts['host'] === $grav['uri']->host());
```

With X-Forwarded-Host, an attacker could easily trick this function to make it think the media within a given markdown input is `local` while it's actually not.

Here there is no subsequent `file_get_contents()` or `fopen()`, but we can imagine how quickly this could be introduced somewhere in the codebase, in the future and easily open unnoticed doors (including within themes/plugins).

mahagr:
@rhukster I think this is a valid concern and can be used for attacks. We cannot trust X_FORWARDED_HOST which means that it should not be used for anything.

https://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

Looks like this feature is disabled by default in most CMSes.

rhukster:
I concur with you guys. I should of done more research before approving. It will be removed in next release.

mahagr:
@rhukster Can we make it configurable and optional instead? I think that in the configurations where it is used, it will be always set by a nearby server/firewall.

rhukster:
I've added configuration options for all of these in system.yaml but only 'host' is disabled by default so it maintains backwards compatibility:

```
http_x_forwarded:                                # Configuration options for the various HTTP_X_FORWARD headers
  protocol: true
  host: false
  port: true
  ip: true
```

PR #2897, opened 2020-04-27
Alternative representation

A----:
I've been fiddling with the idea for a while now. In many cases, I need to have different representations of the same page. There are _ways_ (namely headers) but those have limitations. They does not use the same pipeline as page content (so, no plugins).

Here comes what I called Alternative Representations. It's one or more _other_ markdown files next to the usual one. Its use cases are:
* Summaries but on steroïds
* Other kind-of-similar use-cases : having different representation of the page when displayed on the home page / search results page / …
* Complex content in classic or modular pages

Here's an super-simple illustration of the last use case. I need a two-columns page (or page module). Additionnally, I want to use the image-captions plugins. Here's what I get if I use a header:

image

Now, using the alternative representation:

image

Content pipeline (including plugins) processes the alternative MD file, caption is displayed.

My folder looks like this:
image

In my `left-column` template, I now use this:

```twig
{{ page.alternativeRepresentation('side-content').content | raw }}
```

As of now, implementation might have unexpected results in some cases (namely multi-language sites) and I may need some guidance there.

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -180,6 +180,41 @@ class Page implements PageInterface
     }
 
     /**
+     * Return a different representation of the same page.
+     * Place a new file named _<representation>.md next to the current
+     * markdown file.
+     *
+     * @param string $representation the name of the alternative representation.
+     *
+     * @return Page A new instance of Page or null if the representation does not
+     * exists or is not published.
+     */
+    public function alternativeRepresentation($representation)
+    {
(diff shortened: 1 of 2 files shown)


mahagr:
Well, we had the same idea as I've implemented one to flex already. In there you can have an unlimited number of layouts for both collections and objects.

It should already kind of work for the Flex Pages as well, just define the layouts and use the `render()` method. Unfortunately, as it is now, all the layouts are shared between every page type... So that part may still need some work.

PR #2906, opened 2020-05-05
Updates donatj/phpuseragentparser to 1.0.0

donatj:
I maintain `donatj/phpuseragentparser` and recently released v1.0.0 😄 

It's entirely backwards compatible, but also adds a namespaced variant which is the prefered use. I have changed the code to point to that.

Everything else should continue to work as expected.

I have also added an object oriented wrapper should you wish to change to that. As is, this was simply the clearest route forward.

I am happy to answer any questions.


mahagr:
@donatj Thank you.

Issue #2397, opened 2019-03-05
Drop gpm. Use composer (+ possibly plugins)


drzraf:
Sooner or later, most other mainstream PHP framework/CMS dropped their custom installer. Be it bundle installation on `Symfony` or `drush-make` at `Drupal`.
That does not mean they dropped command line tools, but rather that custom `dependency-resolution/download/extraction/installation` and `pre|post-install|update` scripts where standardized using `composer`. In case of very particular needs (like powerful `Symfony flex` recipes) this gone through a custom `composer` plugin.

I went over previous discussion ¹ and still fail to see any current of future case for `GPM` that wouldn't be solved in a better way by `composer` + plugins (let's recall that composer plugins can be referenced as dependencies of the original `composer.json` file).

Looking at `gpm list`, all of the features and underlying concepts are provided by `composer` : source, repositories, stability. And tarballs isn't worth NIH since it can be done by anyone using an alias for `curl | tar -C`.

Front-page talks about `sophisticated Package Manager to make it super flexible`. I tend to disagree. `composer` is way more flexible and sophisticated and well-known. And if project from `mediawiki` to `WordPress/bedrock` can use it, I don't what would make Grav so specific it could not do it.

Let's take the risk to show some actual use-case that `gpm` fails to address:
* I want to install a plugin from `git`: `composer --prefer-source` or `preferred-install` key
* I want to have a base plugin set and customize some installation: A set `composer-*.json` requiring `wikimedia/composer-merge-plugin`
* I want a default template installation: `post-root-package-install`
* I want a specific plugin post-install processing, assets downloading, ...: `script: { "post-install-cmd": { ... } }`
* I want something more standard in terms of installation solution w.r.t. other CMS: `composer/installers` and `installer-paths`
* I want private repositories: use the `repositories` key
etc...etc...etc...

Granted: `composer` is slow and can't extract a tarball from a command-line parameter URL.
Still, Grav is great, it's worth focusing of what it best does. The PHP package management problem has already been solved by existing mainstream tools.

¹ https://discourse.getgrav.org/t/why-gpm-and-not-composer/3124/4

rhukster:
Composer has come a long way since we started Grav and created GPM.  It's definitely more capable now, and there is a fair bit of overlap, and of course composer has a huge amount of great functionality.  However, there are things we do in GPM that are simply not possible in Composer. Things like support for handling license keys (currently unused, but we plan on utilizing this support soon).

[...]

PR #2913, opened 2020-05-13
gpm: Plugins containing "theme" should not be installed as themes

drzraf:
https://github.com/Sommerregen/grav-plugin-themer/issues/6
`./bin/gpm -vvv direct-install https://github.com/Sommerregen/grav-plugin-themer/archive/v1.1.0.zip`

Up to you to evaluate the pattern themes must match, but I felt that `theme-` would be reasonable... until all this could be replaced by composer (or a composer plugin like symfony flex).

### DIFF
--- a/system/src/Grav/Common/GPM/GPM.php
+++ b/system/src/Grav/Common/GPM/GPM.php
@@ -575,10 +575,10 @@ class GPM extends Iterator
 
         // either theme or plugin
         $name = basename($source);
-        if (Utils::contains($name, 'theme')) {
+        if (Utils::contains($name, 'theme-')) {
             return 'theme';
         }
-        if (Utils::contains($name, 'plugin')) {
+        if (Utils::contains($name, 'plugin-')) {
             return 'plugin';
         }
         foreach (glob($source . '*.php') as $filename) {


mahagr:
@rhukster @w00fz What do you think? I think it might be better to try to find word "plugin" first, right now the code looks for "theme" and only if it's not found, checks for "plugin" in the name.

drzraf:
(One more thing that would be resolved by https://github.com/getgrav/grav/issues/2397)

mahagr:
What do you think of changing the order of the tests (too)?

drzraf:
Will work for me in that specific case. Another possibility is to tighten the match to `\btheme\b`.

w00fz:
I don’t think the order should matter. Matching `^grav-theme.+`, `grav-plugin.+` and `grav-skeleton.+` should be sufficient.
Where do you see this check happening @mahagr, is it in the internal repo generator? I was pretty sure to have a dash in there but might be wrong.

mahagr:
@w00fz The check is in here:

https://github.com/getgrav/grav/blob/c5bef56e05c16662a5b997f79de79960cac5e702/system/src/Grav/Common/GPM/GPM.php#L559-L596

It doesn't check anything else but the folder name of zip file if it matches.

w00fz:
Actually the package type checks for `theme-` in the name and also the `php` class. If it's extending Plugin it's a plugin, if it's extending Theme is a theme..

w00fz:
I'm ok with your change @drzraf but could you do the same thing for the plugin check down below?

mahagr:
@w00fz Actually it does not. The check for folder name comes first and the rest of the checks are never made.

PR #2924, opened 2020-05-25
Fixed enviroment name mapping for localhost

izzyu91:
This part should map the local ips to the name "localhost" to provide a single point for configurations. 

The feature was introduced in
https://github.com/getgrav/grav/commit/380b3be92812e71f9453d6ab15d100dd70588111

but should have never worked as the bug is already there.

Please check my change carefully as it seems to me as a potentially breaking change...


izzyu91:
Sorry for the push mess. Shouldn't have done this today, is not my day...

PR #2937, opened 2020-06-15
Adjusted asset types to enable extension of asset classes

jackthomasatl:


rhukster:
What is the specific use case for extending assets class?

jackthomasatl:
I'd like to handle my front-end assets differently. Are you against the idea of extending grav?

I'm using s3 to store my assets instead of local.

jackthomasatl:
Currently I apply this change via patch file when I build my local. I'd like to see it as part of the official source.

ricardo118:
if you have the assets externally, do you even need to use grav's assets?

rhukster:
I actually don't have a problem making this more extensible, i was really just curious :)

jackthomasatl:
I hear that, I moved my docroot folder below the base directory. For images I host the content via s3 and css files I host them locally but have the webroot deeper into the folder structure. I've got an extension of BaseUtils that modifies the uri method and I've also got extensions to the Assets and Medium classes.

Issue #2942, opened 2020-06-19
:443 requests to getgrav.org after login


elusamus:
According to my community post (https://discourse.getgrav.org/t/grav-external-requests/12633) there seems to be getgrav.org:443 communication during login process and also later.
1.7.rc-12 is Grav version installed. Haven't had any luck to out-configure such communication.

elusamus:
Trial-and-error method revealed, that /system/src/Grav/Common/GPM/Remote/Plugins.php has  "https://getgrav.org/downloads/plugins.json" specified and somewhy it needs to be called right after login.

Why it's important to make that call, still no idea.

mloitm:
This effectively disables Grav admin for environments, where outgoing connections are restricted for security reasons.

I think, there should be a config setting for disabling GPM auto update checks. There's one for admin plugin, I think that could be reused. 

I can create a pull request for that.

PR #2949, opened 2020-06-30
gpm: Disable auto update checking from configuration

mloitm:
Related to https://github.com/getgrav/grav/issues/2942

Disables GPM-s automatic update check in when `plugins.admin.enable_auto_updates_check` configuration value is `false`.

This is required for environments, where outgoing network connections are restricted.


mahagr:
Grav cannot use admin configuration -- it may not be installed. The fix needs to be in admin.

Issue #2882, opened 2020-04-15
Add option to disable breaking markdown links


tg-x:
Grav breaks links by [removing certain attributes](https://github.com/getgrav/grav/blob/ef80e28d1db6a8d59581d77b26231169a7194c8b/system/src/Grav/Common/Page/Markdown/Excerpts.php#L90) if it finds them.
There should be a global configuration option to turn this off instead of adding `&noprocess` one by one to each link.

This is an unintuitive anti-feature that breaks user expectations of being able to link to pages without unintended side-effects of breaking links by silently dropping commonly used attributes.
There have been several reports before from users unwittingly stumbling upon this, e.g. #989, #1877

rhukster:
Only a few reports in 5 years, disabling site-wide would mean huge chunks of functionality suddenly not working, leading to more issues/questions/etc. Might be better to put a good entry in the doc's so people find the solution quickly after googling.

OleVik:
This warrants a global option to disable though, as there are many cases where you either need the pass through of query parameters, or media-processing disabled altogether. It would be advantageous to be able to get the media with query-parameters and have Grav process it afterwards.

otahirs:
I ran into the same issue.

rhukster:
Fixed via https://github.com/getgrav/grav/commit/bfbe4ce1b847ade2df074b1df5900b1daa3a2e7f

Issue #2839, opened 2020-02-26
links with target=_blank together with noprocess do not work


matteomattei:
Hello,
I tried to write a link with target=_blank and noprocess parameter but it seems they do not work together. Basically the attribute target="_blank" will not be inserted in the link but remains in the query string. If I remove the noprocess then it will.

Example:
`[my link](https://www.mydomain.com?id=blablabla&target=_blank&noprocess)`

The html result is this:
`<a href="https://www.mydomain.com?id=blablabla&amp;target=_blank">my link</a>`

Instead of:
`<a href="https://www.mydomain.com?id=blablabla" target="_blank">my link</a>`

Thanks.

rhukster:
This is a known issue because `noprocess` stops ALL processing of links including the target attribute

PR #2954, opened 2020-07-06
added optional options to noprocess

ricardo118:
This enables the option &noprocess to be expanded to &noprocess=id,hl

Example:

`[PlayStore](https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de&target=_blank&noprocess)`

Current result link: `https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de&target=_blank` - not opening in new tab

This will work as intended by not processing the `id`, however the target also gets no processed.

Using 

`[PlayStore](https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de&target=_blank&noprocess=id)`

Will skip the `id` processing, and still process the target. Resulting in a better end result link.

New result link: `https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de` - opening in new tab successfully

This is in reference to the following issues:

https://github.com/getgrav/grav/issues/2839 - FIXED

https://github.com/getgrav/grav/issues/2882

https://discordapp.com/channels/501836936584101899/501854266873348112/729728029063905345

### DIFF
--- a/system/src/Grav/Common/Page/Markdown/Excerpts.php
+++ b/system/src/Grav/Common/Page/Markdown/Excerpts.php
@@ -104,12 +104,18 @@ class Excerpts
             // Valid attributes supported.
             $valid_attributes = Grav::instance()['config']->get('system.pages.markdown.valid_link_attributes');
 
+            $noprocess = [];
             // Unless told to not process, go through actions.
             if (array_key_exists('noprocess', $actions)) {
+                if (!is_bool($actions['noprocess'])) {
+                    $noprocess = explode(',', $actions['noprocess']);
+                }
                 unset($actions['noprocess']);
-            } else {
-                // Loop through actions for the image and call them.
-                foreach ($actions as $attrib => $value) {
+            }
+
+            // Loop through actions for the image and call them.
+            foreach ($actions as $attrib => $value) {
+                if (!empty($noprocess) && !in_array($attrib, $noprocess)) {
                     $key = $attrib;
 
                     if (in_array($attrib, $valid_attributes, true)) {
--- a/system/src/Grav/Common/Page/Markdown/Excerpts.php
+++ b/system/src/Grav/Common/Page/Markdown/Excerpts.php
@@ -104,18 +104,16 @@ class Excerpts
             // Valid attributes supported.
             $valid_attributes = Grav::instance()['config']->get('system.pages.markdown.valid_link_attributes');
 
-            $noprocess = [];
+            $skip = [];
             // Unless told to not process, go through actions.
             if (array_key_exists('noprocess', $actions)) {
-                if (!is_bool($actions['noprocess'])) {
-                    $noprocess = explode(',', $actions['noprocess']);
-                }
+                $skip = is_bool($actions['noprocess']) ? $actions : explode(',', $actions['noprocess']);
                 unset($actions['noprocess']);
             }
 
             // Loop through actions for the image and call them.
             foreach ($actions as $attrib => $value) {
-                if (!empty($noprocess) && !in_array($attrib, $noprocess)) {
+                if (!in_array($attrib, $skip)) {
                     $key = $attrib;
 
                     if (in_array($attrib, $valid_attributes, true)) {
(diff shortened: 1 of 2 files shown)


ricardo118:
theres some bugs found in discord currently testing

ricardo118:
```
[noprocess](https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de&target=_blank&noprocess)  
result:  `https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de&target=_blank`  
should have `target=_blank` in url and **not open in a new tab**  

[noprocess=id](https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de&target=_blank&noprocess=id)  
result:  `https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=de`  
should have `id=org.jitsi.meet&hl=de` in url and **open in a new tab**  
  
[only target](https://play.google.com/store/apps/details?target=_blank)  
result:  `https://play.google.com/store/apps/details`  
should  **open in a new tab**  
  
[Browser](https://meet.weikamp.biz/Support?rel=nofollow&target=_blank)  
result:  `https://meet.weikamp.biz/Support`  
should  **open in a new tab**  
```

Test cases above

rhukster:
be great if you could add these to the test suite @ricardo118

rhukster:
Also will need some updates to learn to show how it can be used with options

rhukster:
I added some tests, and used those to confirm the functionality is good.  perhaps you can use these for future reference.

tbowan:
I needed to remove this "automatic extraction of links attributes" so I removed all of them in the configuration.

Un fortunately, as there is no attributes, the configuration element is no more an array but "null" and I got Exception "in_array() expects parameter 2 to be array, null given" (`/system/src/Grav/Common/Page/Markdown/Excerpts.php` at line 131) :

```php
            // Loop through actions for the image and call them.
            foreach ($actions as $attrib => $value) {
                if (!in_array($attrib, $skip)) {
                    $key = $attrib;
 
                    if (in_array($attrib, $valid_attributes, true)) { // <---- this one is the 131
                        // support both class and classes.
                        if ($attrib === 'classes') {
                            $attrib = 'class';
                        }
                        $excerpt['element']['attributes'][$attrib] = str_replace(',', ' ', $value);
                        unset($actions[$key]);
                    }
                }
            }
```

My workaround without changing the code is to add a random string as a valid attribute in the hope no website will use it as valid attribute (but as I choose it randomly, chances are really tiny).

To fix it in the code, I changed the 117th line as follows (adding the `?: []` at the end) :

```php
            $valid_attributes = $grav['config']->get('system.pages.markdown.valid_link_attributes') ?: [] ;
```

mahagr:
@ricardo118 Can you take a look?

ricardo118:
ya will do

ricardo118:
> 
> 
> I needed to remove this "automatic extraction of links attributes" so I removed all of them in the configuration.
> 
> Un fortunately, as there is no attributes, the configuration element is no more an array but "null" and I got Exception "in_array() expects parameter 2 to be array, null given" (`/system/src/Grav/Common/Page/Markdown/Excerpts.php` at line 131) :
> 
> ```
>             // Loop through actions for the image and call them.
>             foreach ($actions as $attrib => $value) {
>                 if (!in_array($attrib, $skip)) {
>                     $key = $attrib;
>  
>                     if (in_array($attrib, $valid_attributes, true)) { // <---- this one is the 131
>                         // support both class and classes.
>                         if ($attrib === 'classes') {
>                             $attrib = 'class';
>                         }
>                         $excerpt['element']['attributes'][$attrib] = str_replace(',', ' ', $value);
>                         unset($actions[$key]);
>                     }
>                 }
>             }
> ```
> 
> My workaround without changing the code is to add a random string as a valid attribute in the hope no website will use it as valid attribute (but as I choose it randomly, chances are really tiny).
> 
> To fix it in the code, I changed the 117th line as follows (adding the `?: []` at the end) :
> 
> ```
>             $valid_attributes = $grav['config']->get('system.pages.markdown.valid_link_attributes') ?: [] ;
> ```
> 
> (I was going to submit a pull request with this change but did not find the Excerps.php file in develop or 1.7 branches so I don't know how to fix it in the github code).

i just tested this, and while yes what you say does happen, i fixed the error without having any issues by simply setting the config to
`valid_link_attributes: []` instead of `valid_link_attributes: null`.

However, admin saves it as null, so i believe the check suggested here is valid.  Pushed a change here [https://github.com/getgrav/grav/commit/2813934d21b04ef5c9a084dfea3e0873588d69e7](https://github.com/getgrav/grav/commit/2813934d21b04ef5c9a084dfea3e0873588d69e7)

PR #2963, opened 2020-07-11
Update to Caddyfile using Caddy 2 changes.

CanntAim:
Caddy 1 is officially deprecated in favor of Caddy 2. With that there are some changes to directives and such. I'm in the process of updating my blog site to use Caddy 2 over Lighttpd and the following changes seem to be OK. I'm not seeing any server-side or client warnings and/or errors but perhaps other users can confirm.

Official documentation on steps to upgrade, which I followed, are available [here](https://caddyserver.com/docs/v2-upgrade).

I noticed the original configuration listened on a network port vs. a unix socket in regards to `fastcgi` (I'm not sure if you want to keep this.)

### DIFF
--- a/webserver-configs/Caddyfile
+++ b/webserver-configs/Caddyfile
@@ -1,33 +1,24 @@
 :8080
-gzip
-fastcgi / 127.0.0.1:9000 php
+encode gzip
+root * /path/to/grav/root
+php_fastcgi unix//run/php/php7.3-fpm.sock
+file_server
 
 # Begin - Security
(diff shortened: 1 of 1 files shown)


max312:
Instead of rewriting to /403 u could shorten it by directly responding a 403

OleVik:
Would be nice to find a way to generalize this to be applicable to Windows as well, @CanntAim. Are you aware if v2.2.0 of Caddy now supports an equivalent to the `startup` directive of v1?

PR #2987, opened 2020-08-24
Add translation hook for metadata content

nbusseneau:
Problem: I wanted to localize the `description` metadata, but couldn't find a way to localize `site.yaml`.
I settled on using `languages.yaml` as a way to localize `site.yaml` metadata keys from `metadata.html.twig`, when they exist in both.

Users can override `site.yaml` metadata keys by providing a corresponding `GRAV.METADATA.key` via `languages.yaml`.

This is a working PR, but I don't believe this solution is elegant, even if it works fine. Input is welcome. Ideas?


Skymirrh:
While skimming the doc, I discovered that we can actually override the `description` metadata at the page-level. Since pages are easily localized, localizing `description` for the home page is actually very easy.

The translation hook trick would only be useful if one wants to localize the default global `description` tag for when one is not set up at the page level.

I have given it some more thought and I believe a good solution would be to mimic what's currently possible with templates: https://learn.getgrav.org/16/content/multi-language#language-based-twig-templates

Subfolders for overriding config values, e.g. `config/fr/site.yaml`.

What do you think?

PR #844, opened 2016-05-20
Docker + docker-compose

FDiskas:

PR #2988, opened 2020-08-24
Adding Dockerfile

AntonLydike:
I noticed this repository is missing a dockerfile, and [this old pr](https://github.com/getgrav/grav/pull/844) tried to fix that but did not succeed.

I have not tested this Dockerfile with many plugins, maybe some dependencies are still missing, but it'll provide a nice base for future expansions. It allows building of images with preinstalled addons via the `--build-args NSTALL_PLUGINS="..."` flag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+FROM php:7-apache
+
+# install gd
+RUN apt-get update && apt-get install -y \
+		libfreetype6-dev \
+		libjpeg62-turbo-dev \
+		libpng-dev \
+	&& docker-php-ext-configure gd --with-freetype --with-jpeg \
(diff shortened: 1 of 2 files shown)


Skymirrh:
I think it's because the Dockerfile can be found here: https://github.com/getgrav/docker-grav

AntonLydike:
Maybe link it in the Readme? I had trouble finding the repo through google and this projects readme...

Skymirrh:
I agree the documentation could mention it.

Not sure about you but on my end it does show as the first result on Google: 
image

AntonLydike:
Oh well, seems like I did in fact not do proper research^^

PR #3011, opened 2020-09-18
Custom nginx-ddev-site.conf no longer required

jgonyea:
The custom nginx-ddev configuration is no longer required.  DDEV and Grav play much nicer together with current defaults.


jgonyea:
There will be a pull request shortly for updated learn.getgrav.org documentation on ddev and grav setup

PR #3013, opened 2020-09-22
Add option for timeout in selfupgrade command

grantholle:
Upgrading from a server within China was brutally slow. It couldn't update without timing out. This adds an option on the `selfupgrade` command to accommodate a timeout value for downloading the update.


mahagr:
Default timeout should be at least 30 seconds IMHO.

grantholle:
[15 is the default](https://github.com/getgrav/grav/blob/59b3b6cc0250e4db865e55d7e1948852e256a8c6/system/src/Grav/Common/GPM/Response.php#L44) in the Response class, which is what I based the default on here. I agree that 15 is pretty small.

mahagr (approved):
This one looks good to me.

I would still raise the default timeout limit to 30 seconds as your connection may be slow.

grantholle:
Raised it to 30

grantholle:
Any word on getting this approved?

rhukster:
Merged!

PR #3035, opened 2020-10-12
Sub-page redirect with code fails

randoum:
## Situation

Having a redirect set-up with a 3xx code

```
redirects:
  '/old_path': '/new_path[301]'
```

And accessing *http://my.site/old_path* the response is a *301* to *http://my.site/new_path*, as expected

## Issue

But when accessing *http://my.site/old_path/sub-page* the response is a *302* to *http://my.site/new_path[301]/sub-page*, which in turn respond another *302* to *http://my.site/new_path[301]/new_path[301]/sub-page* and so on and so forth until "TooManyRedirect" error is raised by the client.

## Solution

This PR intend to fix it and provide the following behavior : when accessing *http://my.site/old_path/sub-page* the response is a *301* to *http://my.site/new_path/sub-page*

### DIFF
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -320,7 +320,7 @@ class Grav extends Container
         $route = preg_replace("#^\/[\\\/]+\/#", '/', $route);
 
          // Check for code in route
-        $regex = '/.*(\[(30[1-7])\])$/';
+        $regex = '/.*(\[(30[1-7])\]).*/';
         preg_match($regex, $route, $matches);
         if ($matches) {
             $route = str_replace($matches[1], '', $matches[0]);


mahagr:
I don't think that the redirects were intended to be used like this, but I do agree that the redirect from a subpage was not what I'd expect for. It should either display 404 or use the same redirect rule.

That said, I don't agree with the fix either as it may cause other issues.

randoum:
Well, then what would you agree with?

Given redirects with a custom status is a need, and the code exists in Grav core to accept this syntax. 
The bug I declared in the original message exists too (can you reproduce?).

Any way you look at it, something need to be done.

mahagr:
Should be fixed in the next release.

PR #3048, opened 2020-10-23
Fix compatibility with Symfony 4.2 and up.

ViliusS:
This enables running scheduled *shell* commands again because Symfony 4.2 deprecated passing commands as a string to the Process component: https://symfony.com/blog/new-in-symfony-4-2-important-deprecations#deprecated-process-commands-as-strings

One also needs to catch all possible exceptions from the Process component, though. These exceptions are never displayed or sent to any of the logs and I've spent hours trying to debug why my scheduled tasks are failing.

Fix should be applied to 1.6 and 1.7 branches.


rhukster:
Good catch, thanks.

PR #3050, opened 2020-10-24
Fix failing example custom job.

ViliusS:
Since Symfony 4.2 passing chained shell commands to the Process component is not supported anymore and a working directory needs to be set by passing it as a completely separate parameter.

Unless somebody finds a way to use Process() for this and fixes it in the code, rework example custom job.

Related info: https://symfony.com/blog/new-in-symfony-4-2-important-deprecations#deprecated-process-commands-as-strings


ViliusS:
Did anyone had time to look into this?

rhukster:
Also merged.. thanks!

PR #3059, opened 2020-11-09
Add new locator schema `templates://`

NicoHood:
Also see https://github.com/getgrav/grav-learn/issues/851

### DIFF
--- a/system/src/Grav/Common/Twig/Twig.php
+++ b/system/src/Grav/Common/Twig/Twig.php
@@ -103,6 +103,9 @@ class Twig
             $this->twig_paths = array_merge($this->twig_paths, $core_templates);
 
             $this->loader = new \Twig_Loader_Filesystem($this->twig_paths);
+            
+            // Add all template path as schema for locator
+            $locator->addPath('templates', '', $this->twig_paths);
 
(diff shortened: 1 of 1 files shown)


rhukster:
what is the compelling reason for this?

NicoHood:
The reason for adding the docs or adding `templates://`?

1. It is used in multiple plugins but the parameters are not explained anywhere.

2. I want to create a plugin that adds structured data (json-ld) to the page. It processes the `page.header` and if it finds an entry it searches a corresponding template.

For example:
```
structured_data:
    local_business:
        name: 'Pizza Restaurant'
        address: [...]
```

```php
// Check if a template exists for the given data
$template = 'structured-data/json-ld/' . $key . '/' . $key . '.json.twig';
if ($locator->findResource('templates://' . $template, true, false) === false) {
    continue;
}
$this->grav['twig']->processTemplate($template);
[...]
```

mahagr:
It won't work. The reason is that twig isn't using a stream and paths can be added later, too.

NicoHood:
I dont understand, it works for me. And the event to add path is run first, so all templates should be in place. But I am no expert here, however I've need a solution to lookup the templates from a plugin.

Issue #3060, opened 2020-11-10
Dangerous extension .js includes .json


stephan-strate:
Having `.js` listed as Dangerous Extension (Configuration -> Security or `config/security.yaml`) leads to errors when trying to upload a `.json` file using the PageMedia upload.

![image](https://user-images.githubusercontent.com/19595615/98704415-ecc23700-237c-11eb-9323-a7694da36df4.png)

Yes, `.json` files (**JavaScript** Object Notation) are somehow connected with `.js` files. But I think it should be possible to upload them (`.xml` and `.yaml` files can be upload aswell).

PR #3061, opened 2020-11-11
Check exact extension in checkFilename utility

stephan-strate:
As described in #3060 I changed the `checkFilename` function to check the exact extension.


rhukster:
Thanks! looks good to me.

PR #3063, opened 2020-11-11
Allow to set SameSite from system.yaml

randoum:
This patch allow to set `config.session.samesite` from system.yaml.
It does not set any default value, hence do not change the current behavior.


rhukster:
This is good, we also need this to be added to the admin plugin.

randoum:
Could you help? I have no idea how it works

rhukster:
Really all you have to do is add the new field to the `system/blueprints/config.yaml` and translation strings to the `user/plugins/admin/languages/en.yaml` file

randoum:
Thanks I'll check

randoum:
Ok done

PR #3065, opened 2020-11-13
add custom attribute to medium

stephan-strate:
This pull request adds the ability adding custom attributes to medium instances.
Eg. `page.media[...].attribute('data-section', 'my-data').html()` would generate ``

Empty `.attribute()` does nothing, `.attribute('key')` adds only the key without value.


rhukster:
I manually merged this as things have changed in 1.7

PR #3070, opened 2020-11-18
Fixed MIME types in default media settings

1tsi:
Fixed MIME types for .docx, .pptx and .xlsx for default media settings.

The wrong MIME types caused errors when redirecting to these types on mobile browsers.


rhukster:
I had fixed this for a client, but forgot to merge it back into Grav core.  Thanks

PR #3115, opened 2020-12-28
Fixed Link to XSS Security Docs

thekenshow:
Was linking to .local server.


OleVik:
It should be linking to http://learn.grav.local/17/forms/forms/form-options#xss-checks, not https://learn.getgrav.org/17/basics/grav-configuration#security. The latter refers to an addition to 1.5, the former to the addition to 1.7 that the CHANGELOG references.

PR #3122, opened 2021-01-03
Fix #2781 asset manager pipeline order

### LINKED ISSUE #2781: AssetManager creates different hash for same pipelined files when extra asset is added to before/after position
**Grav version:** grav-admin-v1.7.0-rc.2

**Problem:**
AssetManager creates different hashes for pipelined files when extra asset is added in 'before' or 'after' position. The contents of the different pipelined files are exactly the same though.

**Expected result:**
Assets in position before/after should not change hash of pipelined assets.

**Example:**

I have the following in base.html.twig:
```
{% block javascripts %}
   {% do assets.addJs('https://code.jquery.com/jquery-3.4.1.min.js', { priority: 100, loading: 'defer' }) %}
   {% do assets.addJs('theme://js/popper.min.js', { priority: 98, loading: 'defer' }) %}
   {% do assets.addJs('theme://js/bootstrap.min.js', { priority: 97, loading: 'defer' }) %}
 {% endblock %}

{{ assets.js('head', {loading: 'defer'}) }}
```
The output is:
```
<script src="/grav/site-base/assets/3c4701a9a8e3c486b719e9d09e64a839.js" defer></script>
```

On some pages I add a plugin.  My plugins add assets as follows:
```
$this->grav['assets']->addJs("plugin://contactform/js/contactform.min.js", [
   'group' => 'head',
   'position' => 'after',
   'loading' => 'defer',
]);
```
The output is:
```
<script src="/grav/site-base/assets/24180c8c0af4f6c7bc1b1f1f95235033.js" defer></script>
<script src="/grav/site-base/user/plugins/contactform/js/contactform.min.js" defer></script>
```

The contents of '3c4701a9a8e3c486b719e9d09e64a839.js' and '24180c8c0af4f6c7bc1b1f1f95235033.js' are according to diff exactly the same.

NicoHood:
Patch taken from @pamtbaau

### DIFF
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -59,6 +59,8 @@ class Assets extends PropertyObject
     protected $collections;
     protected $timestamp;
 
+    /** @var array to contain next 'order' value per asset 'type', 'group' and 'position' combination */
+    protected $order = [];
 
     /**
      * Initialization called in the Grav lifecycle to initialize the Assets with appropriate configuration
@@ -190,7 +192,14 @@ class Assets extends PropertyObject
         $options['timestamp'] = $this->timestamp;
 
         // Set order
-        $options['order'] = \count($this->$collection);
+        $group = $options['group'] ?? 'head';
+        $position = $options['position'] ?? 'pipeline';
+
+        if (!isset($this->order[$type][$group][$position])) {
+           $this->order[$type][$group][$position] = 0;
+        }
+
+        $options['order'] = $this->order[$type][$group][$position]++;
 
         // Create asset of correct type
         $asset_class = "\\Grav\\Common\\Assets\\{$type}";
--- a/system/src/Grav/Common/Assets.php
+++ b/system/src/Grav/Common/Assets.php
@@ -59,7 +59,7 @@ class Assets extends PropertyObject
     protected $collections;
     protected $timestamp;
 
-    /** @var array to contain next 'order' value per asset 'type', 'group' and 'position' combination */
+    /** @var array Array to contain next 'order' value per asset 'type', 'group' and 'position' combination */
     protected $order = [];
 
     /**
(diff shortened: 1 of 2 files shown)


rhukster:
I will test this when I get caught up. Also, i see a test failing, so need to evaluate if that is related to a new bug, or something that is based on this fixing a bug we were just accepting as the correct value in the test.

pamtbaau:
@rhukster, During the unit tests, the arrays containing the assets in class `Assets` are being reset before each test using `$this->assets->reset();`. Therefor the `$options['order'] = count($this->$collection)' starts with 0 with each new test.

However, the newly added `Assets::order` property is not reset after each test and hence keeps increasing. This leads to the failing tests.

Solution:
Add a reset of the newly added `Assets::order` property in 'TestingAssetsTraits::reset()':
```
public function reset()
{
    $this->resetCss();
    $this->resetJs();
    $this->setCssPipeline(false);
    $this->setJsPipeline(false);

    $this->order = [];  <-- Reset the newly added 'order' property

    return $this;
}
```

All Unit tests pass successfully now.

NB. I believe this new algorithm calls for new tests to check if checksum remains the same when assets are added in the before or after position.

NicoHood:
I've applied the fix.

system/src/Grav/Common/Assets.php
```diff
@@ -59,6 +59,8 @@ class Assets extends PropertyObject
     protected $collections;
     protected $timestamp;
 
+    /** @var array to contain next 'order' value per asset 'type', 'group' and 'position' combination */
```
mahagr, line 62: Be careful with these, the first word should be php type (as it happens to be this time).

I would add extra array: `/** @var array Array to..` just to prevent someone changing the comment and messing it up.

mahagr: I think it would also fix the comment if creating a documentation out of this:

type: array
description: to contain next...

NicoHood: Fixed!

rhukster:
I still need to go over this one.. sorry for the delay.

mahagr (approved):
I think this looks good now.

PR #3128, opened 2021-01-05
Fix #3127 redirect trailing slash using 301 statuscode

### LINKED ISSUE #3127: Trailing slash issues
There are multiple issues with trailing slashes in Grav.

The current situation:
* `https://mywebsite.de/contact` is the canonial url
* `https://mywebsite.de/contact/` redirects to `https://mywebsite.de/contact` with 302 status code
* The homepage is **always** `https://mywebsite.de/`, not `https://mywebsite.de` due to the http protocol. Browsers do not display them though!

Those are the current issues:
1. Redirects should be 301 rather than 302 (Update: Fixed now)
2. `https://mywebsite.de/` is the website root with a canonial of `https://mywebsite.de`
1. The [sitemap plugin](https://github.com/getgrav/grav-plugin-sitemap) lists `https://mywebsite.de` instead of `https://mywebsite.de/` (also an invalid canonial)

This is important for search engines, as they will no correctly crawl the main homepage (which is most important).

**Solutions:**
1. Can be fixed via https://github.com/getgrav/grav/pull/3128

2. & 3. Can be fixed in multiple places (I guess), you should know best. I've edited `Page.php` like this (can be improved possibly):
```php
// trim trailing / if not root
if ($url === '') {
    $url = '/';
}
else if ($url !== '/' && rtrim($url, '/') !== rtrim($grav['uri']->base(), '/') ) {
    $url = rtrim($url, '/');
}
```

NicoHood:
Fixes:
#3127
https://github.com/getgrav/grav-plugin-sitemap/issues/87


mahagr:
This should be already fixed in Grav 1.7: https://github.com/getgrav/grav/blob/1.7/system/src/Grav/Common/Processors/InitializeProcessor.php#L422

NicoHood:
This is also not fixed in 1.7
The redirect still uses 302 instead of 301. I have no solution for this yet.

mahagr:
I mean that the code does not redirect from the server root, which fixes the redirect loop.

I don't want to use 301 for redirects ever, especially on development sites. I've broken my sites so often with 301 and the only way is to clear all the cache from your browser. That itself isn't bad, but I've also wasted my time trying to locate a bug that isn't there.

I can work with a configurable redirect code, though.

NicoHood:
I do not understand the problem with the redirect code. But it is a fact, that 302 is wrong here and will cause bad SEO results.

mahagr:
Manually added the fix.

Issue #3067, opened 2020-11-15
Status: Googlebot blocked by robots.txt


the-s-a-m:
The following resources can not be loaded by googlebot because they are blocked by default by robots.txt

/system/assets/jquery/jquery-2.x.min.js
/user/plugins/form/assets/form-styles.css
/user/plugins/login/css/login.css

Thank you for fixing this error in the default configuration

NicoHood:
The current robots is:
```
User-agent: *
Disallow: /backup/
Disallow: /bin/
Disallow: /cache/
Disallow: /grav/
Disallow: /logs/
Disallow: /system/
Disallow: /vendor/
Disallow: /user/
Allow: /user/pages/
Allow: /user/themes/
Allow: /user/images/
Allow: /
```

Maybe it would make sense to add `Allow: /user/plugins/` to the list?
Why is it important to load the listed resources? Does google even care about javascript and css?

the-s-a-m:
Google told me that the access is not possible therefore I changed it to

`User-agent: *
Allow: /user/pages/
Allow: /user/themes/
Allow: /user/images/
Allow: /user/plugins/
Allow: /system/assets/jquery/jquery-2.x.min.js
Disallow: /backup/
Disallow: /bin/
Disallow: /cache/
Disallow: /grav/
Disallow: /logs/
Disallow: /system/
Disallow: /vendor/
Disallow: /user/
Allow: /
`

and now it is working

rhukster:
Fixed by https://github.com/getgrav/grav/commit/bc22c8d2b1efad6f28608aa937e9dd0485eac6d7

NicoHood:
It seems that the issue is not fixed for the builtin jquery:
`/system/assets/jquery/jquery-2.x.min.js`

The reason why this happens is unclear to me. It might be that `Allow: *.js$` does not include `script.min.js` syntax (at least from my observations, as other `.js` scripts work. Searching around in google i found [this post](http://www.thesempost.com/how-to-unblock-all-css-javascript-for-googlebot-using-robots-txt/) that suggests the following:

```
User-Agent: Googlebot
Allow: .css
Allow: .js
```

This also allows query parameters that are often used for cache busting: `https://example.com/deep/style.css?something=1`. And yes, `User-Agent: Googlebot` is important to use here! Another option is to explicitly allow the jquery path.

NicoHood:
[...]

PR #3129, opened 2021-01-05
Allow builtin jquery to be crawled by bots #3067

NicoHood:
See https://github.com/getgrav/grav/issues/3067#issuecomment-754775815


rhukster:
Already merged

PR #3131, opened 2021-01-06
Redirect file extension fix #3130

### LINKED ISSUE #3130: Page extensions return duplicated content using 200 statuscode instead of 301 redirect
I found som issues with page extensions, eg `https://yourdomain.de/contact.html`:
1. Any characters appended to the specified `config.system.pages.types` will also work. E.g. `https://yourdomain.de/contact.htmlxxx`
2. The website return a 200 instead of a 301 to the original url. This is important for SEO to avoid duplicated content

### Solution
1.: The error comes from the `Uri` class function `isValidExtension()`. The regex should be an exact match using `^exact$` syntax:
https://github.com/NicoHood/grav/commit/4131e9c2494aac463cf4be491f34a8877514b9bb

2.: `Uri` class function `init()` removes the postfix from the url which causes the issue. Note that `html` and `htm` will **always** return a 200, as those are set as default, no matter what `config.system.pages.types` contains. The reason why I even found this, is the weird behavior for the sitemap plugin, which made the sitemap available via xml and "normal" url. It causes lots of trouble for search engines, if there is duplicated content. In this case it would be best to only serve the xml file.

Page processing order:
1. InitializeProcessor -> trailing slash logic happens here
2. Uri.init() -> 200 logic currently happens here
3. PagesProcessor
4. onPageInitialized -> sitemap gets added here
5. onPageNotFound
6. onPageNotFound -> error plugin serves 404
7. PagesProcessor ends

It would be best to only do the redirect if no page was found (or served by any plugin) and serve the sitemap only at `/sitemap.xml` (or whatever is configured). In order to still allow plugins to serve special `.xml` files we must place the redirect after the `onPageInitialized` event but before `onPageNotFound`. A possible solution is to move the code from `Uri` to the `PagesProcessor`:

https://github.com/NicoHood/grav/commit/ecbcdbd5fea1540cb3a7921d484c5d897ee13d18

The code solves the following:
* 301 redirect instead of 200 duplicated content (SEO improvement)
* Allow plugins to server special files like `sitemap.xml` on the extension uri **only**
* Still allow redirects of .html etc to the "real" page
* But also add an option to turn it off: `redirect_file_extensions`
* Backwards compatibility for the sitemap plugin to server the sitemap at `sitemap` and redirect at `sitemap.xml`
* Option to serve the sitemap now at `sitemap.xml` only, without any .htaccess rule

**Fix provided in PR https://github.com/getgrav/grav/pull/3131**

NicoHood:
See details explained here: https://github.com/getgrav/grav/issues/3130


mahagr:
FYI: the fix breaks unit tests.

NicoHood:
I have updated the unit test accordingly. However this of course might break stuff. Now how can I test this? Do you any usecases why those tests were written for?

mahagr:
Now that you fixed unit tests, I think these will break things as route and path should never have extension in them (routing in many places rely on it).

Personally I do not like Uri class at all, it does all the wrong things and is really hard to work with. So I don't use it myself anymore. :)

NicoHood:
Now that is the question. Cause if a page ends with `html` this does not break things, as the page never gets loaded with `.html`, because it will get redirected to the correct url. So it may only break for those pages, that explicitly require a file extension, which is the sitemap plugin for example. But as I said it works there. The question now is, if there are other places where this is an issue.

If you do not like the Uri class, an alternative solution should be documented and the class should be deprecated. Not even the API docs have an alternative solution documented.

mahagr:
Yeah, the issue is that the `Uri` class defines a path as a route and forbids a valid file extension from being in there. It gets very confusing as paths can still have dots in them if they're not ending to a valid extension such as `.html`.

There is a replacement, it is `Route` class (which in turn can return PSR-7 `Uri` class). The issue is that I've never had time to finish it so it could be used as a full replacement; I merely needed it to be able to alter paths in admin and flex.

NicoHood:
Alright, but how do we go on now? This should get fixed.

mahagr:
I am not sure how to fix this (yet).

mahagr:
I merged / improved the fix for `Uri::isValidExtension()`, it'll be in the next version.

What comes to the second issue, I cannot use your fix as it breaks handling other extension types, such as JSON responses.

mahagr:
Fixed with alternative approach. See the original issue.

PR #3139, opened 2021-01-09
Add grav twig extensions to its own `evaluate` and `evaluate_twig` fi…

NicoHood:
…lters

I wanted to use `evaluate_twig()` with a template the uses `theme_var()` or also a simpler twig function `string`. I made them available with this simple call.

### DIFF
--- a/system/src/Grav/Common/Twig/TwigExtension.php
+++ b/system/src/Grav/Common/Twig/TwigExtension.php
@@ -837,6 +837,7 @@ class TwigExtension extends \Twig_Extension implements \Twig_Extension_GlobalsIn
 
         $loader = new \Twig_Loader_Filesystem('.');
         $env = new \Twig_Environment($loader);
+        $env->addExtension($this);
 
         $template = $env->createTemplate($twig);
 


mahagr:
Ugh, using the deprecated Twig methods. Lemme fix those in 1.7.

mahagr:
Looks like the code is already updated in 1.7, good :)

NicoHood:
Should PRs target master or 1.7? I guess you need to rebase/merge on master to pull this into 1.7

mahagr:
Right now develop for minor bug fixes and 1.7 (or 1.8/2.0) for new features in general.

Basically we have two working branches, develop is the next stable bugfix release and the tag with future version is the one where all the real work goes in.

In the beginning of the cycle we do allow small (or separate) features to go to the develop, but because of the need to keep it as stable and backwards compatible as possible, everything new and interesting always happens in the version tag.

mahagr:
If you are planning to stay active in the project, I recommend talking to us also in the Discord as it allows more unofficial way to quickly communicate to each other. 

Thank you for being active, it really helps us to get the bugs fixed before the big release next week.

Issue #3072, opened 2020-11-20
1.7 Final Issues / Enhancements


w00fz:
Here to list all the known and final issues that we do need to get resolved before 1.7 release:

- [x] system:// stream is causing issues in Admin, making Media tab to disappear and possibly causing other issues. First introduced via https://github.com/getgrav/grav/commit/747055f60e3e897a76363161dd16108c4048913a
- [x] Add `user/env` as primary environmental location and fallback to `user/` (for traditional behavior backward compatibility). This will allow to keep all the environments under one place the user folder tidy.
- [x] Autoescape will be enabled by default and to guarantee BC we need postflight installer to be able to handle installations vs updates. (ed. I believe this is already done)
- [x] Support picking up changes in Flex Users (such as manual password reset) when a user yaml is updated.
- [x] `safe_email` filter should remove HTML from the email (tags, entities) and output HTML safe output [see here](https://github.com/getgrav/grav/issues/3069#issuecomment-731774440).
- [x] Fix clockwork to work with the latest client

// cc: @rhukster @mahagr

mahagr:
`system:// stream` issue has been fixed in Toolbox update (needs a release).

mahagr:
Clockwork client 5.0.1  has a bug on rendering user defined data. The bug has been fixed, but isn't released just yet.

mahagr:
@rhukster For object refresh support grav, **login** and flex plugins need to be updated

PR #3140, opened 2021-01-09
Fixed `|safe_email` filter to return safe and escaped UTF-8 HTML [#3072]

NicoHood:
Backported from 1.7:
https://github.com/getgrav/grav/commit/068de42e833af8baeb96fe31c170964fa2df0d4b

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# v1.6.32
+## mm/dd/2021
+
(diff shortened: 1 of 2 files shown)


mahagr:
In 3 days nobody is able to update Grav 1.6 anymore (our updates only support the latest version). This means that Grav 1.6 will be an unsupported version because of the upgrade logic does not allow it to be supported.

PR #3141, opened 2021-01-09
Fix Lithuanian nativeName

Karmalakas:
That deleted second word literally means _language_, which is completely unnecessary in language choosing context (like for all other languages)


Karmalakas:
I don't really mind in Admin to see this long native name, but now I want to add langSwitcher plugin and it uses this name. There's no way I want to show in front-end this wrong string to people who understand how wrong it is :/

Issue #1168, opened 2016-11-10
Webp image format enhancement


tidiview:
#may be that's wrong; if so forgive me.

Would it be possible to add `webp` image format to the native GRAV caching media process?
By that I mean that when there is a `webp` file, it's cached like for a `jpeg` one.

I added a webp setting in a `media.yaml` file in `config` like in the docs, but I can't see anything.

The idea is to be able to have structures like:
```
<figure><picture markdown="1">  
<source srcset="/link/to/my/Webp/image.webp" type="image/webp">  
![image-title](/link/image.jpg) {.class}  
</picture></figure>
``` 

for browser that accept `webp` format.

Thanks in advance

rhukster:
I'm not sure webp has a enough market penetration to make it worth supporting natively.  Maybe someone could create a plugin?

euene:
Market penetration of WebP probably isn't too strong, but [browser support](http://caniuse.com/#feat=webp) is at over 72% globally which is actually quite decent.

flaviocopes:
I don't know how they calculate their stats on caniuse.com but Chrome+Opera+Android doesn't look like 72% to me, anyway been marked as enhancement.

fredrikekelund:
@flaviocopes caniuse.com probably get pretty good statistics, since they're based on statcounter.com. But there's a difference between global and local stats, I only get 52% support for Sweden, for instance.

If we wan't to get WebP support for the `page.media.images` array, then we'll likely need support from Gregwar/Image (and there are additional considerations for cross browser support). There's a ticket in their bug tracker for that (https://github.com/Gregwar/Image/issues/77), but there hasn't been a whole lot of movement in that project for some time. I don't know how far you can get around this with the plugin architecture (I think it might be quite difficult), but there is a ticket for looking into alternative libraries for image manipulation #77

ganar:
I would like to have WebP support as an option: I have a client's site with 64% of views coming from compatible browsers, most of them mobile. 
I would like to be able to apply the enhancement via TWIG in the templates, using a `<picture>` polyfill.
This would be transparent to the users, but would greatly improve the speed of the site to 64% of the audience.

rhukster:
Our image library doesn't really support this,  if we can change the library we might be able to revisit this.

fuslwusl:
[...]

PR #3144, opened 2021-01-11
Added 'format' to magic actions for images - this will allow conversion and caching of webp images

angusmcb:
A very simple change that would make it possible to set the output format of an image from Twig / Markdown.

This would be particularly useful for being able to serve automatically generated webp images, created from jpegs.

Related to issue #1168


NicoHood:
Very nice idea! How would the twig code look like then? I was searching for such an option some time ago.

angusmcb:
@NicoHood It works like this:

`{{ page.media['sample-image.jpg'].format('webp').url|e }}`

NicoHood:
This does not work with `derivatives`. It only changes the main image to webp, but not the others in srcset.

apokaliptis:
Any idea when this feature will hit a stable release?

PR #3148, opened 2021-01-13
Add GRAV_VERSION to twig variables

### LINKED ISSUE #3142: Get grav version from plugin or twig
Is there a way to read out the Grav version from a plugin or twig to support version specific features?

NicoHood:
Fix #3142

### DIFF
--- a/system/src/Grav/Common/Twig/Twig.php
+++ b/system/src/Grav/Common/Twig/Twig.php
@@ -68,6 +68,7 @@ class Twig
     {
         $this->grav = $grav;
         $this->twig_paths = [];
+        $this->twig_vars['GRAV_VERSION'] = GRAV_VERSION;
     }
 
     /**


mahagr (changes requested):
I am really not fan of adding more and more variables into twig variables. They are copied and merged on every single twig context making the code slower. Something like `grav.version` would be much cleaner (in PHP `$grav->getVersion()`).

NicoHood:
Function calls are faster? I am happy to fix that, if you think it is better. Is there also a function to get the version of a plugin? If the function could also do `grav.version('plugin.form')`. That'd be nice!

mahagr:
Method call is not faster. But having 1000 array merges with quadzillion extra variables surely costs more, especially as this is something that will likely be called only a single time per request. :)

Plugin versions are in the plugin blueprints, there's method to get those.

mahagr:
Fixed in https://github.com/getgrav/grav/commit/c7dcb9144c0aed147c32dfa124f38bb2e1204e6c , use `grav.version` to access it.

NicoHood:
Nice!

PR #3149, opened 2021-01-13
Add german translations for GRAV.YES and GRAV.NO

NicoHood:
They were simply missing :-D


NicoHood:
I want to note, that they do not appear in crowdin (the translation tool). So yes and no is very likely not translated for all other languages as well. I dont know how to fix that though.

mahagr:
@rhukster ? How the translations/translating are supposed to work?

PR #3187, opened 2021-01-30
Added ability to filter enabled or disabled with bin/gpm index

jgonyea:
Third time's the charm (hopefully)

Adds ability to filter index results based on enabled/ disabled settings.

### DIFF
--- a/system/src/Grav/Common/GPM/GPM.php
+++ b/system/src/Grav/Common/GPM/GPM.php
@@ -139,6 +139,32 @@ class GPM extends Iterator
         return $this->installed['plugins'];
     }
 
+
+    /**
+     * Returns the plugin's enabled state
+     *
+     * @param  string $slug
+     * @return bool True if the Plugin is Enabled. False if manually set to enable:false. Null otherwise.
+     */
+    public function isPluginEnabled($slug)
+    {
+        $result = null;
+        $grav = new Grav();
+        $config = $grav::instance()['config']['plugins'];
+        $installed = $this->isPluginInstalled($slug);
+
+        $enabled = isset($config[$slug]) && ((bool)$config[$slug]['enabled'] == true);
+        $disabled = isset($config[$slug]['enabled']) && ((bool)$config[$slug]['enabled'] === false);
+
+        if ($enabled) {
+            $result = true;
+        }
+        if ($disabled) {
+            $result = false;
+        }
+        return $result;
+    }
+
     /**
      * Checks if a Plugin is installed
      *
@@ -180,6 +206,28 @@ class GPM extends Iterator
     public function getInstalledThemes()
     {
         return $this->installed['themes'];
+    }
+
+    /**
+     * Checks if a Theme is enabled
+     *
+     * @param  string $slug The slug of the Theme
+     * @return bool
+     *   True if the Theme has been set to the default theme. False if installed, but not enabled. Null otherwise.
+     */
+    public function isThemeEnabled($slug)
+    {
+        $grav = new Grav();
+        $result = null;
+        $current_theme = $grav::instance()['config']['system']['pages']['theme'];
+        $theme_installed = $this->isThemeInstalled($slug);
+
+        if ($current_theme == $slug) {
+            $result = true;
+        } elseif ($theme_installed) {
+            $result = false;
+        }
+        return $result;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


system/src/Grav/Common/GPM/GPM.php
```diff
@@ -139,6 +139,32 @@ public function getInstalledPlugins()
...
+     */
+    public function isPluginEnabled($slug)
+    {
+        $result = null;
+        $grav = new Grav();
+        $config = $grav::instance()['config']['plugins'];
+        $installed = $this->isPluginInstalled($slug);
+
+        $enabled = isset($config[$slug]) && ((bool)$config[$slug]['enabled'] == true);
+        $disabled = isset($config[$slug]['enabled']) && ((bool)$config[$slug]['enabled'] === false);
```
mahagr, line 157: Umm, aren't these two lines the same?

Here's better code:
```
return (bool)($config[$slug]['enabled'] ?? false);
```

system/src/Grav/Common/GPM/GPM.php
```diff
@@ -139,6 +139,32 @@ public function getInstalledPlugins()
...
+    /**
+     * Returns the plugin's enabled state
+     *
+     * @param  string $slug
+     * @return bool True if the Plugin is Enabled. False if manually set to enable:false. Null otherwise.
+     */
+    public function isPluginEnabled($slug)
+    {
+        $result = null;
+        $grav = new Grav();
```
mahagr, line 152: This won't work. Use `$grav = Grav::instance();`

system/src/Grav/Common/GPM/GPM.php
```diff
@@ -182,6 +208,28 @@ public function getInstalledThemes()
...
+    /**
+     * Checks if a Theme is enabled
+     *
+     * @param  string $slug The slug of the Theme
+     * @return bool
+     *   True if the Theme has been set to the default theme. False if installed, but not enabled. Null otherwise.
+     */
+    public function isThemeEnabled($slug)
+    {
+        $grav = new Grav();
```
mahagr, line 220: Same as above, use instance.

system/src/Grav/Common/GPM/GPM.php
```diff
@@ -182,6 +208,28 @@ public function getInstalledThemes()
...
+        $result = null;
+        $current_theme = $grav::instance()['config']['system']['pages']['theme'];
+        $theme_installed = $this->isThemeInstalled($slug);
+
+        if ($current_theme == $slug) {
```
mahagr, line 225: I'd do the same as with plugins. I would also move check for installed outside of this (and plugin) method and check it elsewhere..

system/src/Grav/Console/Gpm/IndexCommand.php
```diff
@@ -138,7 +150,8 @@ protected function serve(): int
                         'Name' => '<cyan>' . Utils::truncate($package->name, 20, false, ' ', '...') . '</cyan> ',
                         'Slug' => $slug,
                         'Version'=> $this->version($package),
-                        'Installed' => $this->installed($package)
+                        'Installed' => $this->installed($package),
+                        'Enabled' => $this->enabled($package),
```
mahagr, line 154: and have something like: `'Enabled' => $this->installed($package) ? $this->enabled($package) : '',` here.

mahagr (changes requested):
A few changes are required, others are nice to have.

system/src/Grav/Common/GPM/GPM.php
jgonyea: I think there are technically 3 states:
1. config exists and is set to enabled
2. config exists and is set to disabled
3. config doesn't exist (grav treats this as disabled)

I was attempting to take into consideration the tricky difference between #2 and #3 cases.

system/src/Grav/Common/GPM/GPM.php
jgonyea: I'll fix this.

system/src/Grav/Common/GPM/GPM.php
jgonyea: Same as above...I'll fix this.

system/src/Grav/Console/Gpm/IndexCommand.php
mahagr: @jgonyea If you check here, this change would check all your 3 cases.

Issue #3180, opened 2021-01-28
Slug based URLs, forms and multi-language


Karmalakas:
So I'm actually not sure how to explain the issue very clearly, but I'll try

Grav 1.7.3 + Form 4.3.0 plugin. All freshly installed

Basically, trying to implement one feature on website (building personal page from scratch) encountered an issue, which uncovered a bunch of problems with Gravs core and how in handles URLs, routes, redirects and pages.

First: the only way to get slug based URL, is `page.route()` (at least that's what I figured), but that's not always an option, because sometimes you want to add a link just from some saved [pages field](https://learn.getgrav.org/17/forms/blueprints/fields-available#pages-field), which saves slug value [without language code](https://github.com/getgrav/grav-plugin-admin/issues/2021). That wouldn't be a problem in a normal page, because we can just `pages.find([SLUG]).route` or `pages.find([FOLDER]).route`, but there's a problem - `pages === null` in modules of modular pages.

OK, we can't get to `page.route`, then lets try redirects. Yes, even though they are one of the worst enemies of SEO, but still - we don't have much of a choice. When in _system.yaml_ you have `pages.redirect_default_route: true`, when you get to folder based URL, this setting enabled, redirects to correct slug based URL. OK it seems we have a really dirty workaround, BUT...

After redirect, `forms()` will return `null` no matter how you try to get the form - by name or by route. And only cache clearing helps. But it helps only if you have a single language page and won't navigate to folder based URLs (so that redirect wouldn't happen). If you have at least two languages and want to switch it on page with a form, again there's no easy way to get slug URL for same page in different language and you have to use whatever URL you can get (most likely folder based URL). After clicking on new language redirect happens and form is gone.

Same applies if you want to implement AJAX form and have different slugs from folder name. You can't use common _frontmatter.yaml_ for all languages, because you need to have separate actions for form in each language depending on slug, because if you use folder based action URL, after submit it redirects to correct slug based URL and whole payload is gone - form never gets submitted.

It's kind of a closed loop
You need to have correct (slug based) URLs to be somewhat good to SEO. There's no way to get slug URLs, because you can't access `page.route` in some situations (especially for different language). So you just use folder based URLs with redirects, but you can't use them either not only because of SEO, but because of redirects after which you lose form payload or lose forms all together.

I was already liking Grav and even considering a donation, but no no matter how easy it is to extend, I'm starting to find it pretty much unusable if you want to have beautiful multilingual slugs and at least a contact form.

---

[...]

PR #3188, opened 2021-01-30
Add method to get route by language

Karmalakas:
Just an idea on how [redirect problem](https://github.com/getgrav/grav/issues/3180) might be fixed maybe. Or at  least partially.
Although this doesn't work - couldn't figure out how to get parent. Without the parent, `route()` returns just a home route.
Maybe someone could take this over, or tell me the approach is completely wrong :) But that issue really needs to be fixed. Otherwise I'll have to search for alternatives and I really like Grav, so it would be sad if I couldn't publish my page and had to ditch a few weeks work :(

Also `Page::init()` calls `$this->id($this->modified() . md5($this->filePath()))`, but it gets overwritten, because `$this->id`  doesn't exist yet, so added a check in `Page::id()`

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -91,6 +91,8 @@ class Page implements PageInterface
     protected $url;
     /** @var array */
     protected $routes;
+    /** @var array */
+    protected $languageRoutes;
     /** @var bool */
     protected $routable;
     /** @var int */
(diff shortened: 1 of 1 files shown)


OleVik:
[Parent needs to be instantiated](https://github.com/getgrav/grav/pull/2709), no? I would assume that this goes into the comprehensive work on improving multi-language, @ricardo118? Both pull requests and related issues address deficiencies in current language-handling.

PR #3193, opened 2021-02-01
Set twig autoescaping to true

NicoHood:
I think this value get overwritten anyways. Maybe we should remove it completely?


mahagr:
It's still being used in old sites.

PR #3226, opened 2021-02-15
read all language files from theme stream

stephan-strate:
Currently when inheriting a theme and creating language files in the child theme, they will override the language files of the base theme. There is no chance to extend from the base theme language files other than copying them. This leads to a lot of manual syncing effort when updating the base theme.
```
themes
- child-theme
-- languages.yaml

- base-theme
-- languages.yaml <- content never loaded
```

I changed the language file loading to read language files from all themes in `theme://`.

This might be interesting for @lazybadger.

### DIFF
--- a/system/src/Grav/Common/Themes.php
+++ b/system/src/Grav/Common/Themes.php
@@ -318,6 +318,7 @@ class Themes extends Iterator
 
     /**
      * Load theme languages.
+     * Reads ALL language files from theme stream and merges them.
      *
      * @param Config $config Configuration class
      * @return void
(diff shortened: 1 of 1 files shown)


NicoHood:
Do you really want to load all themes? Wouldnt it make more sense to only load the base theme?

stephan-strate:
> Do you really want to load all themes? Wouldnt it make more sense to only load the base theme?

I am only loading all themes in inheritance chain (so this will be usually two themes, if using inheritance)! What do you do if you want to create additional language variables in child theme?

NicoHood:
Alright, that makes sense to me.

mahagr (approved):
Looks good to me.

rhukster:
Looks good to me too

PR #3251, opened 2021-03-01
Fix Markdown image attr. 'loading'

pamtbaau:
There are 2 issues in `grav/common/page/markdown/Excerpts::processMediaActions()` that prevents the `loading` attribute for images to be set properly in Markdown.

1. Line 281: `$defaults = $config['images']['defaults'] ?? [];`
   Config settings are not picked up.
   - Should be `$defaults = $this->config['images']['defaults'] ?? [];`
2.  Line 284: `if (!array_search($method, array_column($actions, 'method'))) {`
   `array_search` returns (in above use) an `int`, or `false`. When value is found at position `0` the negation will return `true` and hence, explicit value set in Markdown will be overridden by value from config.
    - Should use strict comparison: `if (array_search($method, array_column($actions, 'method')) === false) {`

### DIFF
--- a/system/src/Grav/Common/Page/Markdown/Excerpts.php
+++ b/system/src/Grav/Common/Page/Markdown/Excerpts.php
@@ -278,10 +278,10 @@ class Excerpts
             );
         }
 
-        $defaults = $config['images']['defaults'] ?? [];
+        $defaults = $this->config['images']['defaults'] ?? [];
         if (count($defaults)) {
             foreach ($defaults as $method => $params) {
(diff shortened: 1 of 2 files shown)


pamtbaau:
What I'm still missing in the implementation of `loading` is the implementation of the default in Twig.

That means that theme templates must be updated because `{{ image.html() | raw }}` will not pickup the default loading.

mahagr (approved):
I agree with both changes. They look like bugs to me.

pamtbaau:
Thanks @mahagr

PR #3256, opened 2021-03-05
Add SRI integrity hash to assets

miguelsalespereira:
Having the website pre-generated and hosted on a CDN, I would like to have integrity hashes included in CSS and JS assets.

This PR will allow:
* User to set if wants the integrity hash to be generated and added to their <script> and <link> elements
    *  controlled via new `system.assets.enable_asset_sri` config option
* When true, this option will generate the integrity hash and add it to both pipelined and not pipelined assets

### DIFF
--- a/system/src/Grav/Common/Assets/BaseAsset.php
+++ b/system/src/Grav/Common/Assets/BaseAsset.php
@@ -180,7 +180,7 @@ abstract class BaseAsset extends PropertyObject
      *
      * @return string
      */
-    public function integrityHash( $input )
+    public static function integrityHash( $input )
     {
         $grav = Grav::instance();
 
--- a/system/src/Grav/Common/Assets/BaseAsset.php
+++ b/system/src/Grav/Common/Assets/BaseAsset.php
@@ -10,6 +10,7 @@
 namespace Grav\Common\Assets;
 
 use Grav\Common\Assets\Traits\AssetUtilsTrait;
+use Grav\Common\Config\Config;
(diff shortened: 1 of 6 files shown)


w00fz:
Sweet, thanks!

w00fz (approved):

system/src/Grav/Common/Assets/BaseAsset.php
```diff
@@ -171,6 +172,31 @@ public function setPosition($position)
...
+     */
+    public static function integrityHash( $input )
+    {
+        $grav = Grav::instance();
+
+        $assetsConfig = $grav['config']->get('system.assets');
+
+        if ( !empty($assetsConfig['enable_asset_sri']) && $assetsConfig['enable_asset_sri'] )
+        {
+            $dataToHash = file_get_contents( GRAV_ROOT . $input);
```
mahagr, line 191: @w00fz I think we should cache this part and maybe just check file modification time when hash recalculation isn't needed.

rhukster: i agree

w00fz: Yeah I agree, also it should probably also use the Grav File instance and locator for the asset content and path, rather than file_get_contents

rhukster: We can probably merge then optimized to be honest.

rhukster (approved):

PR #3266, opened 2021-03-12
Method `$pages->find()` should never redirect

mahagr:
This is my attempt to fix some redirect issues in $pages->find()` method. 

It attempts to keep the old behavior with one exception: `dispatch()` will attempt to find the regexp routes before redirects, which makes more sense than the old behavior.

I have not changed how the blind redirects work (defined in `site.redirects` configuration), IMHO they should take the path as it is and route accordingly. If you want to redirect a path to an existing page, it's better to use `site.routes` instead and force redirect by setting `system.pages.redirect_default_route` to 301/302 instead.

CC: #3004,  #2435


mahagr:
This pull request will not fix the language redirects needing the language code, but will only separate the logic between find() and dispatch() methods and make `site.routes` more usable.

For the language codes, I still think that `site.redirects` should not be based on routes, but they should blindly match the current URL (including query parameters etc) and do transformation based on that.

However it could be a good idea to be able to pass more information to the redirects, such as `'/({LANG})/foo/(\d+)': '/$1/bar/$2'` which would match all language codes.

The reason why I'd like to do it in that way is that you do not want to restrict redirects to Grav routes as you may have migrated from another CMS.

PR #3267, opened 2021-03-12
Correct width and height after image manipulation

Karmalakas:
I have various size images (both horizontal and vertical). All are at least 1900px on longest side (some 2000px)
```twig
{% set image = item.cropResize(1600, 1600) %}
{{ image.width ~ 'x' ~ image.height }}
```
Result is something like `2000x1350` or `1200x1900`, but it should be 1600 on longest side

Debugged a bit and found the solution, but was afraid, that calling ` $this->image->get()` to actually save the image to get correct dimensions might not be the way. Searched a bit and found [this issue on Gregwar/Image](https://github.com/Gregwar/Image/issues/123) and there's [a fix and a comment](https://github.com/Gregwar/Image/issues/123#ref-commit-948240d) from Gregwar, that this is the way.

This change will cache image with default quality from config and return correct cropped/resized dimensions

### DIFF
--- a/system/src/Grav/Common/Page/Medium/ImageMedium.php
+++ b/system/src/Grav/Common/Page/Medium/ImageMedium.php
@@ -310,7 +310,7 @@ class ImageMedium extends Medium implements ImageMediaInterface, ImageManipulate
 
             if (isset(static::$magic_resize_actions[$method])) {
                 try {
-                    $image = $this->image->get('guess', $this->default_quality);
+                    $this->image->get('guess', 1);
 
                     $this->set('width', $this->image->width());
                     $this->set('height', $this->image->height());
--- a/system/src/Grav/Common/Page/Medium/ImageMedium.php
+++ b/system/src/Grav/Common/Page/Medium/ImageMedium.php
@@ -308,10 +308,16 @@ class ImageMedium extends Medium implements ImageMediaInterface, ImageManipulate
             $this->image->{$method}(...$args);
 
             if (isset(static::$magic_resize_actions[$method])) {
-                $this->image->get('guess', $this->default_quality);
+                try {
+                    $image = $this->image->get('guess', $this->default_quality);
 
-                $this->set('width', $this->image->width());
-                $this->set('height', $this->image->height());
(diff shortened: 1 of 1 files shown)


NicoHood:
The linked issue, why is it still open? I am a bit confused. Was this fixed by upstream or not?

Karmalakas:
Right.. Somehow I didn't even pay attention, that issue is still open, so now you got me worried again.
Also not sure I know what you mean by
> by upstream

Maybe this PR is missing something, but I sure hope the issue itself can be fixed somehow. Any advice maybe what could be improved?

mahagr:
We have our own fork from Gregwar image because the project has been stalled for ages. If there are additional fixes needed, we need to commit those into our fork.

mahagr (approved):

NicoHood:
What do you mean by stalled? There have been some recent commits.

Karmalakas:
Just encountered a fatal error in one use case :/ Will investigate what's happening

Karmalakas:
Updated a PR. Apparently, when image is deleted from page, but page is not saved, image still is on page, but file is missing. Added catching of error and a comment what's happening

Karmalakas:
OK, probably not a viable solution :(

On local all seemed fine, but on prod 30 not so small images looped with `cropResize()` twice for each fails to load the page normally. Couple of times got `Maximum execution time of 30 seconds exceeded`. Probably 5 times got this:
image

Finally, when all images were cached, it loaded, but not an option :/
Removed that _fix_, and even it still takes time, but page loads...

mahagr:
@NicoHood It looks like Gregwar image isn't being maintained anymore by its author. We have our own fork on it to fix PHP 8 issue.

NicoHood:
@mahagr It is, checkout recent releases: https://github.com/Gregwar/Image/releases
It might be worth to open a PR with your changes, so everybody can benefit and you will also receive new fixed from upstream.

mahagr:
Unfortunately, he didn't release Cache package, so unfortunately we cannot use the release. I pulled the fixes into my branch, though.

PR #3272, opened 2021-03-15
Add option images.defaults.size to set default height/width

pamtbaau:
Issue:
[Cummulative Layout Shift](https://web.dev/cls/), the shifting of page content when image's height/width is not set, is not only a bad user experience, but has also become a search ranking factor.

Proposal:
This pull request adds a new default option 'size' to `system/config/system.yaml` with which the image's height/width can automatically be added to an &lt;img> tag. 
```
// system/config/system.yaml

images:
  defaults:
    loading: auto                   # Let browser pick [auto|lazy|eager]
    size: none                      # Add height/width to image [auto|none]. See https://web.dev/optimize-cls/
```
- When value is set to 'auto', Markdown will automatically add the height and width values of an image to the generated &lt;img> tag. 
- Config value can be overridden in Markdown per image. 
- When either or both height/width have been set manually, option is ignored.
- Config is not applied in Twig but can be set manually per image tag.

New Markdown syntax:
```
![](image.jpg?size=none|auto)    <-- Param 'size' only needed to override the default set in config/system.yaml
```

New Twig syntax:
```
{{ image.size('none|auto').html() | raw }}   <-- Without param, function will use default set in config/system.yaml
```

Files changed:
- `system/config/system.yaml`: Add 'size' option 
- `system/blueprints/config/system.yaml`: Add 'size' option field definition
- `system/src/Grav/Common/Media/Traits/ImageSizeTrait.php`: Create new function 'size' which adds height/width attributes
- `system/src/Grav/Common/Page/Medium/ImageMedium.php`: Add use of ImageSizeTrait
- `tests\unit\Grav\Common\Markdown\ParsedownTest.php`: Add Unit tests to test 'size' option

PR requires the following translation keys to grav-plugin-admin:
- IMAGES_SIZE: "Image height/width behavior"
- IMAGES_SIZE_HELP: "The size option sets the height/width values of image to &lt;img> tag to prevent vertical layout shifts of pages. See See https://web.dev/optimize-cls"


rhukster:
I've been looking at this and while it is successfully outputting the width and height, it's always the original width and height.  So that means that no matter what you do with all the 'resize' actions, like cropZoom for example, you get the images original width and height set on the `` tag.  This causes funky behavior. 

Manipulating the built in width and height is not really an option as that needs to be the original width and height as it's used by all those resize functions.  I really think that we need to not actually set the width and height attributes from the image width and height, but rather just think of this as an option for the `html()` method.  And that point the processed image has been saved into cache.  We can then get that imagesize from the cached file and display it in the tag.

I think 'most' of this PR can be used, just the setting/rendering approach needs to be changed.  I'm going to look into it more tomorrow.

rhukster:
BTW made some progress on this already. My current implementation won't work in Twig though.. need to look into that.

pamtbaau:
Ouch, I have overlooked the impact of functions resizing the image: ``

A simple unit test I've just added would have detected that.

rhukster:
I've pushed my changes.  Basically this changes how everything worked as explained in my previous post.  It's no longer a default action, but rather a new set of settings:

```
images:
  ...
  cls:                                           # Cumulative Layout Shift: See https://web.dev/optimize-cls/
    auto_sizes: false                            # Automatically add height/width to image
    aspect_ratio: false                          # Reserve space with aspect ratio style
    retina_scale: 1                              # scale to adjust auto-sizes for better handling of HiDPI resolutions
```

to change inline you just do something like: 

```
![](sample-image.jpg?autoSizes=true&aspectRatio=true&retinaScale=3)
```

rhukster (approved):

pamtbaau:
> I think 'most' of this PR can be used

LOL... only a handfull of unit tests and even these have been changed... :-)

Never mind though, I'm glad the feature request is being implemented.

Did some testing:
- Unit tests succeed (which is surprising considering the following)
- Live website **fails** when using:
  - Markdown: ``
  - system.yaml: 
    ```
    images:
      cls: 
        auto_sizes: true
        aspect_ratio: false
        retina_scale: 1
    ```
  - ImageMedium::auto_sizes is null in method ImageMedium::sourceParsedownElement()
 - Live website **succeeds** when using:
    - Markdown: ``
    - system.yaml: 
      ```
      images:
        cls: 
          auto_sizes: true
          aspect_ratio: false
          retina_scale: 1
      ```

rhukster:
I don't see what you are seeing. The system config option works for all images for me (unless i set one specifically off with `autoSizes=false`

pamtbaau:
Did some tests. All test use the following system.yaml:
```
images:
  cls: 
    auto_sizes: true
    aspect_ratio: false
    retina_scale: 1
```

**Unit tests**
I noticed that when commenting out all unit tests except:
```
$this->config->set('system.images.cls.auto_sizes', true);

self::assertSame(
    '<p><img alt="" src="/tests/fake/nested-site/user/pages/02.item2/02.item2-2/sample-image.jpg" width="1024" height="768" /></p>',
    $this->parsedown->text('![](sample-image.jpg)')
);
```
The test fails.

However, when a unit test before above test is run with `autoSizes=true`, the unit test succeeds.
The following test succeeds
```
self::assertSame(
    '<p><img alt="" src="/tests/fake/nested-site/user/pages/02.item2/02.item2-2/sample-image.jpg" width="1024" height="768" /></p>',
    $this->parsedown->text('![](sample-image.jpg?autoSizes=true)')
);

$this->config->set('system.images.cls.auto_sizes', true);

self::assertSame(
    '<p><img alt="" src="/tests/fake/nested-site/user/pages/02.item2/02.item2-2/sample-image.jpg" width="1024" height="768" /></p>',
    $this->parsedown->text('![](sample-image.jpg)')
);
```

**Markdown**
```
![](image.jpg?autoSizes=true)
![](image.jpg)
```
Similarly, when an image in Markdown is defined with `autoSizes=true` before an image using the default, both images are correct. Remove the upper image, and the lower image will not have height/width

rhukster:
I actually think this is an issue related to testing specifically.  it's like the media object is resused (probably because it's the the same media object and image, and we're doing the parsedown method directly, and not a full markdown process that resets after every 'html' or 'link' call.  I can't replicate the behavior in the tests in an actual page. It works as expected there.

rhukster:
Ok, so was that issue with the same media, but i added some logic to reset the values to the config versions.

So really just a test issue, need to make sure to manually `?reset` after setting something explicitly.

pamtbaau:
`` in combination with `auto_sizes=true` is working now in both Unit tests and live site.

rhukster:
should work as expected without inline `auto_sizes=true` right?

pamtbaau:
Single image in Markdown: ``

- system.yaml: `auto_sizes: false` -> No height/width.
- system.yaml: `auto_sizes: true`  -> Height/width are present.

rhukster:
so all good i think?

pamtbaau:
Because of above issues, I haven't gotten to test other scenarios yet... Will keep you posted.

rhukster:
Merged! we'll fix any remaining issues in develop.

PR #3306, opened 2021-04-09
GRAV_USER_PATH & USER_PATH may start with '/'

makuhama:
I'm not 100% sure but maybe I spotted a few locations where the new fact that paths (like e.g. GRAV_USER_PATH) might start with a `/`.

Please mind that I was not able to test these changes. Especially the change in `InstallCommand.php` may have side effects that I cannot foresee.


mahagr:
Well, it kind of cannot start with `/`. This is because it has to always be under webroot or the images, JS, CSS, fonts would not work.

I did it in the other way; GRAV_WEBROOT can now be set by environment variable and GRAV_USER_PATH cannot be an absolute path anymore.

PR #3308, opened 2021-04-11
Watermark Media Action

ricardo118:
added watermark media function.

This enables gregwar's merge action and a watermark media function.

allows for selecting a image file via config and position. 

position can be top/center/bottom, left/center/right

example:
``

``

Currently the watermark_all option is not working, needs review/help


system/src/Grav/Common/Page/Medium/ImageMedium.php
```diff
@@ -325,6 +330,61 @@ public function retinaScale($scale = 1)
...
+                break;
+
+            case 'center':
+                $positionX = ($this->get('width')/2) - ($watermark->width()/2);
+                break;
+        }
+
+        $this->__call('merge', [$watermark,$positionX, $positionY]);
+
+        dump($this);
```
mahagr, line 384: Maybe you shouldn't leave this one ^

system/src/Grav/Common/Page/Medium/ImageMedium.php
```diff
@@ -72,6 +73,10 @@ public function __construct($items = [], Blueprint $blueprint = null)
         if ($config->get('system.images.cache_all', false)) {
             $this->cache();
         }
+
+        if ($config->get('system.images.watermark.watermark_all', false)) {
+            $this->watermark();
```
mahagr, line 78: I'm wondering if we could delay this action as it makes no sense to generate an image unless the user needs it.

mahagr (changes requested):

system/src/Grav/Common/Page/Medium/ImageMedium.php
ricardo118: yeah that doesn't seem to work anyway, i left it there so could ask for some help/advice where it could be added

ricardo118:
adding scaling to improve the watermark so it works with crop zooms and variable sized images (ie images that are smaller than the watermark itself).

Improved the watermark_all function, it works, but the cache_all_images needs to be turned on, otherwise it will only automatically watermark images that get cached (via other methods)

ricardo118:
is there anything else needed to get it approved?

rhukster:
Nope just need to test it then merge.

mahagr (approved):
Looks good to me (code review).

PR #3321, opened 2021-04-20
query-string redirection, fix #3320

### LINKED ISSUE #3320: 1.6 Fails to redirect is source has a query string.

drzraf:
Consider the presence of a query-string in the source URL when looking to extract a redirection code

So that with:

```yaml
redirects:
  /foo-2019: '/foo[301]'
```

`/foo-2019?bar` now redirects to `/foo?bar` instead of `/foo[301]?bar`

fix #3320

### DIFF
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -320,7 +320,7 @@ class Grav extends Container
         $route = preg_replace("#^\/[\\\/]+\/#", '/', $route);
 
          // Check for code in route
-        $regex = '/.*(\[(30[1-7])\])$/';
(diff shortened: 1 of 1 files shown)


rhukster:
We don’t plan on releasing any more  1.6 releases.

drzraf:
If you don't consider this a PR for 1.6, you may at least want to consider it as a useful bug report for 1.7.

mahagr:
Or maybe rebase it for develop. I'm wondering, though, if there's a better way to handle redirects with redirect code in the URL.

mahagr:
Actually, this has already been fixed in 1.7+.

PR #3330, opened 2021-04-25
nginx: added some security related configs

makuhama:
added an extra rule to deal with ACME challenges
added another rule to allow caching for some static content
added an alternative how to deal with PHP scripts (uwsgi)

### DIFF
--- a/webserver-configs/nginx.conf
+++ b/webserver-configs/nginx.conf
@@ -17,12 +17,11 @@ server {
     ## End - Index
 
     ## Begin - ACME (LetsEncrypt) 
-    # adapt the root folder to where your certification job (e.g. CertBot)
+    # adapt the alias folder to where your certification job (e.g. CertBot)
     # puts the domain verification challenges
-    location ^~ /.well-known/acme-challenge/ {
-        root            /srv/http/acme;
-        rewrite         ^/\.well-known/acme-challenge/(.*)$ /$1 break;
-    }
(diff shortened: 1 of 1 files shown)


webserver-configs/nginx.conf
```diff
@@ -16,7 +16,23 @@ server {
...
 
+    ## Begin - ACME (LetsEncrypt) 
+    # adapt the root folder to where your certification job (e.g. CertBot)
+    # puts the domain verification challenges
+    location ^~ /.well-known/acme-challenge/ {
+        root            /srv/http/acme;
+        rewrite         ^/\.well-known/acme-challenge/(.*)$ /$1 break;
+    }
+    ## End - ACME (LetsEncrypt) 
+
```
rhukster, line 27: Should this be commented out if optional?

webserver-configs/nginx.conf
```diff
@@ -39,6 +57,19 @@ server {
         include fastcgi_params;
         fastcgi_param SCRIPT_FILENAME $document_root/$fastcgi_script_name;
     }
+    # or if you use uwsgi
+    location ~ \.php$ {
+        include uwsgi_params;
+        uwsgi_modifier1 14;
+        uwsgi_pass unix:/run/uwsgi/grav.sock;
+    }
```
rhukster, line 65: I feel this is great but should be commented out as it's optional

rhukster (changes requested):

PR #3336, opened 2021-04-28
Add back Yiddish to Language Codes

yankl:


yankl:
@mahagr You said a friendly reminder was ok so here is one! 😉

Note that Yiddish was among the language codes until [this point](https://github.com/getgrav/grav/commit/d7f286f601229bb0e485c8d1713b5b235f66f10c#diff-765cf749b006e22a32813faba5337377cd53b116485ab8d90b6f902ebbff4247) – that's why it's "add back".

Issue #3345, opened 2021-05-02
daterange without start fails


tbowan:
Hi,

After reading the documentation of [daterange in collection](https://learn.getgrav.org/16/content/collections#date-range), I understood that I can only specify the end date of the range, without providing the start date.

Indeed, if I specify a collection without the start date, grav encounter an exception :

```
content:
  items: '@self.descendants'
  dateRange:
    end: now
  limit: 10
```

The exception is as follows :

```
(An exception has been thrown during the rendering of a template ("DateTime::__construct(): Failed to parse time string (0) at position 0 (0): Unexpected character").)
```

The problem occurs first in the method that build the collection (/system/src/Grav/Common/Page/Pages.php at line 546). As no start is provided, it uses the value 0. Then call `$collection->dateRange(0, 'now', false)` which will always convert the start value (0) into a timestamp with Utils::date2timestamp (which call a basic `new DateTime(0)' which fails).

tbowan:
Here is a diff that fix my problem.

```diff
--- a/system/src/Grav/Common/Page/Collection.php
+++ b/system/src/Grav/Common/Page/Collection.php
@@ -331,7 +331,7 @@ class Collection extends Iterator implements PageCollectionInterface
      */
     public function dateRange($startDate, $endDate = false, $field = null)
     {
-        $start = Utils::date2timestamp($startDate);
+        $start = is_numeric($startDate) ? $startDate : Utils::date2timestamp($startDate);
         $end = $endDate ? Utils::date2timestamp($endDate) : false;

         $date_range = [];
```

tbowan:
The PR has a better fix (_i.e._ fix all implementation of the dateRange method).

mahagr:
I'm currently merging/testing the change.

PR #3350, opened 2021-05-07
daterange - allow optionnal start date

tbowan:
While testing dateRange filtering for collection, I encountered error when the start value is not specified (see #3345 for the issue).

After looking deeper in the source code, I found that the method dateRange() come from an interface and has three implementations and the fix I submited in the issue was not complete. That's why I submit this pull request.

I've divided the fix into three commit so you may discard some of them :
* First commit deal only with the fix of behaviour and touch only minimum lines of codes,
* Second commit simplify a little bit the code,
* Third commit is about the interface (method definition) : using null instead of false (as the behaviour is not specified if you pass true`), updating the documentation of the code (and synchronize the documentation accross all implemtnations) and fixing the only part of the code that call the method (with null instead of a mix of 0, false and null).

As 0, null and false have the same behaviour, the new version is retro-compatible with hypothetical legacy plugins that use the method.

The code have been manually tested on page collection with a fresh install of the fork and default theme :

* Three pages, one in the pase (1800's), one in the present (2000's) and one in the future (2200's),
* Test with start only, end only, both set,
* Same test but with the field attribute specified.

### DIFF
--- a/system/src/Grav/Common/Flex/Types/Pages/PageCollection.php
+++ b/system/src/Grav/Common/Flex/Types/Pages/PageCollection.php
@@ -438,7 +438,7 @@ class PageCollection extends FlexPageCollection implements PageCollectionInterfa
      */
     public function dateRange($startDate, $endDate = false, $field = null)
     {
-        $start = Utils::date2timestamp($startDate);
+        $start = $startDate ? Utils::date2timestamp($startDate) : false;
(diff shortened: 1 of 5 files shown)


mahagr:
FYI: it's good practice to set return value as tight as possible from the docblock -- it allows chaining the method calls without static analyzer reporting errors on undefined methods.

I wish we were using PHP 7.4 or 8.0 minimum as it would allow us to define the return values.

PR #3376, opened 2021-06-04
Fixes to make filters work with Unicode text

qumuq-til:
Putting an end to [A-Z] hegemony one fix at a time


mahagr:
Just a note: It would be nice to have just one regexp per method as it's way faster.

Issue #531, opened 2015-12-17
Collection nextSibling() and prevSibling() are Switched (Reversed)


sturple:
When do a page.collection() loop.  I was using the nextSibling() function and found that the sibling functions were reverse.

## Using 
```
bin/grav -V 
Grav CLI Application version 0.1.0
```

## Page YAML
```yaml
content:
    items: '@self.modular'
    order:
        by: default
        dir: asc
        custom:
            - _hero
            - _feature
            - _about-us
            - _announcements
            - _showcase
```

```twig
    {% for module in page.collection() %}
    	<div class="modular-anchor" id="{{ _self.pageLinkName(module.menu) }}" ></div>
        {{ module.content }}
		{% if module.isLast()%}			
		{% else%}		
		<div class="modular-next-nav"><a data-scroll href="#{{ _self.pageLinkName(module.nextSibling().menu) }}">{{module.prevSibling().menu}} - {{module.nextSibling().menu}}</a></div>
		{% endif%}
    {% endfor %}
```

## Results

```
_feature -
_showcase - _feature
_about-us - _announcements
```

hotovec:
@rhukster @sturple Confirming this behavior.

my case
Hey Guys,

seems there is issue with ordering. Here is my case:
a) I have root pages only
```
user/pages/
   01.home/
   02.themes/
   03.features/
   04.pricing/
   05.sign-in/
```
b) all pages don't specify collection in the header. Pages also do not have order parameter there. This is how each page header look like: (with different titles)
```
title: Home
```

c) I have defined order type in system.yaml
```
pages:
  order:
    by: default
    dir: asc
```

asc(0,1,2,3 .....)

d) I'm using antimatter template and this code to pickup my pages
{% set cp = page.parent().children() %}
next sibling: {{ cp.nextSibling(page.path).title }}
prev sibling: {{ cp.prevSibling(page.path).title }}

But my result are inverted. 
For page features (03), I'm getting next sibling page as 02.themes
and previous siblings is giving me 04.pricing

hotovec:
@rhukster I tried Page.php collections like @self.siblings, seems it is also returning reversed results. But depend on sort direction. in case you have desc it looks like OK, but asc direction is reversed.

rhukster:
Ok taken me a while to decipher this.  I think the problem is simply that we chose a different terminology for `previous` and `next`.  

In my thinking, `next` is **up** the list and `previous` is **down** the list.

[...]

PR #3398, opened 2021-07-06
Add `setCurrent()` method to Page Collection

Karmalakas:
Lets us have a workaround for #531
Also was [a discussion on Discourse](https://discourse.getgrav.org/t/page-prevsibling-nextsibling-inverted/17248/17)

This new method would allow to have expected (not inversed) `prev` and `next` items of the collection in Twig
Eg.:
```twig
    {% set collection = page.parent.collection({items: '@self.children', filter: {published: true, routable: true}}) %}
    {% set isFirst = collection.isFirst(page.path) %}
    {% set isLast = collection.isLast(page.path) %}

    <section>
        {% if not isFirst %}
            {% do collection.setCurrent(page.path) or collection.prev %}
            <a href="{{ collection.current.url }}">Previous page</a>
        {% endif %}

        {% if not isLast %}
            {% do collection.setCurrent(page.path) or collection.next %}
            <a href="{{ collection.current.url }}">Next page</a>
        {% endif %}
    </section>
```

### DIFF
--- a/system/src/Grav/Common/Page/Collection.php
+++ b/system/src/Grav/Common/Page/Collection.php
@@ -152,7 +152,7 @@ class Collection extends Iterator implements PageCollectionInterface
     {
         reset($this->items);
 
-        while (key($this->items) !== $path && key($this->items) !== null) {
+        while (($key = key($this->items)) !== null && $key !== $path) {
             next($this->items);
         }
     }
--- a/system/src/Grav/Common/Page/Collection.php
+++ b/system/src/Grav/Common/Page/Collection.php
@@ -146,6 +146,18 @@ class Collection extends Iterator implements PageCollectionInterface
     }
 
     /**
+     * Set current page.
+     */
+    public function setCurrent(string $path): void
+    {
+        reset($this->items);
+
+        while (key($this->items) !== $path && key($this->items) !== null) {
+            next($this->items);
+        }
+    }
+
+    /**
      * Returns current page.
      *
      * @return PageInterface


mahagr (changes requested):
This is not a bad idea as it's not limited to siblings but next/previous items in the whole collection, which can contain any page on the site.

However, there's minor room for improvement.

system/src/Grav/Common/Page/Collection.php
```diff
@@ -145,6 +145,18 @@ public function intersect(PageCollectionInterface $collection)
...
     }
 
+    /**
+     * Set current page.
+     */
+    public function setCurrent(string $path): void
+    {
+        reset($this->items);
+
+        while (key($this->items) !== $path && key($this->items) !== null) {
```
mahagr, line 155: I would get the key into a variable to speed up the loop.

Karmalakas: But key setting will have to be inside the loop anyway 🤔 Will see what I can come up with. Also will check how expensive this is

Karmalakas: Is it worth introducing a variable to gain 250ms in a 10 mil item collection? 🙂 Ran a few times and one time result was opposite even 🙂 
image

mahagr: It's 9% faster. :) Generally I prefer well-written code over repeating the same thing over and over again.

```
while (($key = key($this->items) !== $path) && $key !== null) {
```

Karmalakas:
Ping @mahagr 🙂 
Not sure if re-request of review is enough or should I ping next time right after re-request 😄

mahagr (approved):

PR #3424, opened 2021-08-13
add mime for `.avif` image format

Rotzbua:
new image format developed by google

references:
https://codelabs.developers.google.com/codelabs/avif#0
https://caniuse.com/avif

@rhukster Is `type: image` automatically processed? If yes and there is no additional check for supported image formats it may cause issues. The support for `avif` by `gd` is not yet available. Maybe with php8.1.


mahagr:
Grav only processes jpegs, pngs, and gifs.

Issue #3428, opened 2021-08-25
How to set  attribute `poster` to video?


dallaslu:
Both `1280x720_1mb.mp4` and `1280x720_1mb.mp4.thumb.jpg` was uploaded to `/user/pages/02.typography/`. And the followed code works well:
```twig
{% set thumb = page.media['1280x720_1mb.mp4'].thumbnail('page').display('thumbnail').url %}
{{ page.media['1280x720_1mb.mp4'].display('source').playsinline(1).autoplay(1).muted(1).loop(1).controls(0).poster(thumb).html()|raw }}
```
Result:
```html
<video playsinline="1" autoplay="1" muted="1" loop="1" poster="/user/pages/02.typography/1280x720_1mb.mp4.thumb.jpg" alt="">
<source src="/user/pages/02.typography/1280x720_1mb.mp4">
Your browser does not support the video tag.
</video>
```
But:

## No poster by default
```twig
{{ page.media['1280x720_1mb.mp4'].playsinline(1).autoplay(1).muted(1).loop(1).controls(0).html()|raw }}
```
Result:
```twig
<video playsinline="1" autoplay="1" muted="1" loop="1" alt="">
<source src="/user/pages/02.typography/1280x720_1mb.mp4">
Your browser does not support the video tag.
</video>
```

## No video without adding variables
```twig
{{ page.media['1280x720_1mb.mp4'].display('source').playsinline(1).autoplay(1).muted(1).loop(1).controls(0).poster(page.media['1280x720_1mb.mp4'].thumbnail('page').display('thumbnail').url).html()|raw }}
```
Result:
```twig
<img playsinline="1" autoplay="1" muted="1" loop="1" poster="/user/pages/02.typography/1280x720_1mb.mp4.thumb.jpg" alt="" src="/user/pages/02.typography/1280x720_1mb.mp4.thumb.jpg">
```

Is there a easy way to set attribute poster to video automatically? (twig & markdown)

PR #3429, opened 2021-08-25
Set thumb image as the attribute poster of video automatically

dallaslu:
ref #3428

### DIFF
--- a/system/src/Grav/Common/Media/Traits/VideoMediaTrait.php
+++ b/system/src/Grav/Common/Media/Traits/VideoMediaTrait.php
@@ -9,6 +9,9 @@
 
 namespace Grav\Common\Media\Traits;
 
+use Grav\Common\Page\Medium\MediumFactory;
+use Grav\Common\Page\Medium\ThumbnailImageMedium;
+
 /**
  * Trait VideoMediaTrait
  * @package Grav\Common\Media\Traits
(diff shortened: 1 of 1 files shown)


system/src/Grav/Common/Media/Traits/VideoMediaTrait.php
```diff
@@ -51,14 +54,27 @@ public function playsinline($status = false)
...
-     * @param  bool $reset
+     * @param array $attributes
+     * @param bool $reset
      * @return array
      */
     protected function sourceParsedownElement(array $attributes, $reset = true)
     {
         $location = $this->url($reset);
 
+        if (!isset($attributes['poster']) || ($attributes['poster'] !== 0 && $attributes['poster'] !== '0')) {
```
kniziol, line 65: It might be simplified in this way:
```suggestion
        $this->setPoster($attributes);
        $this->removePoster($attributes);
```

To make it work, you need to create the following methods in the `VideoMediaTrait` trait:
```php
    private function isPosterUnknown(array $attributes): bool
    {
        return !isset($attributes['poster']) || (int) $attributes['poster'] !== 0;
    }

    private function removePoster(array $attributes): void
    {
        if ($this->isPosterUnknown($attributes)) {
            return;
        }

        unset($attributes['poster']);
    }

    private function setPoster(array $attributes): void
    {
        if ($this->isPosterUnknown($attributes) && $this->thumbnailExists('page')) {
            $thumb = $this->get('thumbnails.page', false);

            if ($thumb) {
                $thumb = $thumb instanceof ThumbnailImageMedium
                    ? $thumb
                    : MediumFactory::fromFile($thumb, ['type' => 'thumbnail']);

                $attributes['poster'] = $thumb->url();
            }
        }
    }
```

The names of new methods may be different. I want to show you the idea. Of course, you'll need to double-check that and ensure it works as expected.

Issue #3269, opened 2021-03-14
system.languages.session_store_active: Language never stored in, or retrieved from session


pamtbaau:
The following setting in `system.yaml` is supposed to store the chosen language extension in the url into the session:
```
languages:
  supported: [it, en]
  session_store_active: true
```
However, the chosen language will never be stored in, nor retrieved from, the session, because function `Language::setActiveFromUri($uri)` is called before the session has been initialized.

**InitializeProcessor.php** 
In `InitializeProcessor::process()`, the line  `$this->initializeUri($config);`, which calls `Language::setActiveFromUri($uri)` is executed before `$this->initializeSession($config);`
https://github.com/getgrav/grav/blob/a67a538dfdeda51cd4cc5e92e4340438e2d188ad/system/src/Grav/Common/Processors/InitializeProcessor.php#L104-L123

**Language.php**
Because of the above, the following if-statements, meant to store and retrieve the language value from the session, will always evaluate to `false`:
https://github.com/getgrav/grav/blob/a67a538dfdeda51cd4cc5e92e4340438e2d188ad/system/src/Grav/Common/Language/Language.php#L249-L261

mahagr:
The language code should really be inside an event `SessionStartEvent` (class).

mahagr:
@pamtbaau Can you test my fix? :)

pamtbaau:
Just did and it seems to be working fine.

- Fresh install Grav 1.7.10
- system.yaml
  ```
  languages:
    supported: [de, en]
    session_store_active: true
  ```
- browse to 'localhost/**en**/typography'
- browse to 'localhost/typography'      ---redirect---> localhost/**en**/typography
- browse to 'localhost/**de**/typography'
- browse to 'localhost/typography'      ---redirect---> localhost/**de**/typography
```

PR #3464, opened 2021-10-12
fix `session_store_active` functionality

nbusseneau:
At the moment and since 1.7.19, the `system.languages.session_store_active` setting has no effect.

Session must be initialized before URI for `$language->setActiveFromUri($uri)` (called from `$this->initializeUri($config) -> $uri->init()`) to properly retrieve / store `active_language` in Session.

This was previously detected in #3269 as per the code comment, but got reversed in 2e9fe80e33b1f6a80b061239fe2f425aa1f54b06.


nbusseneau:
@mahagr Revisiting old PRs -- well this one is not that old but I was on a good run 😁

PR #3467, opened 2021-10-16
Update th.yaml

lisandi:
Added missing terms


lisandi:
We added the Thai Translation to Crowdin - now 100% 5 words were still missing in the actual English translation file which we used as master! I suggest correcting that too.

rhukster:
let me look at the ability to upload the files directly.

rhukster:
Can you attach the thai translation here and i'll try to upload myself?

PR #3468, opened 2021-10-16
Create my.yaml

lisandi:
Adding Myanmar/Burmese Language


lisandi:
How can I add the complete file as is in crowdin - it is a very translater unfriendly environment!

PR #3469, opened 2021-10-16
Update LanguageCodes.php

lisandi:
Adding 
my - Myanmar Burmese
sw - Swahili
lo - Lao
km - Khmer


lisandi:
could you please add those language codes to Grav as we need them here for our new projects? Most urgently Swahili is the main language of many African countries in East Africa. i.e. Tanzania, Ruanda, Uganda, Congo, Kenya, etc. We try to implement Grav in several school projects. Thanks.

PR #3504, opened 2021-12-25
add php 8.1 to test

Rotzbua:
now released https://www.php.net/releases/8.1/en.php

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        php: [ 8.0, 7.4, 7.3]
+        php: [ 8.1, 8.0, 7.4, 7.3]
         os: [ubuntu-latest]
 
     steps:


mdestafadilah:
'mbstring' extension is not loaded. This is required for Grav to run correctly.

grav use 'ext-mbstring'.

Rotzbua:
That is not correct, even if the doc claims it:
https://learn.getgrav.org/17/basics/requirements#php-requirements

The truth is that the extension `mbstring` is just optional:
https://github.com/getgrav/grav/blob/2c252c43b4162fda7dd258927250a603f95fa589/composer.json#L78-L79

Because of the polyfill:
https://github.com/getgrav/grav/blob/2c252c43b4162fda7dd258927250a603f95fa589/composer.json#L22

If this is really a problem there must be at least one single test failing.

mdestafadilah:
@Rotzbua hi, i see too.. but i'm testing it in heroku.

any update here?

Rotzbua:
@mdestafadilah I am not familiar with your setup nor your use case. The best is to create a detailed issue or provide a proper solution as pull request.

mdestafadilah:
> @mdestafadilah I am not familiar with your setup nor your use case. The best is to create a detailed issue or provide a proper solution as pull request.

hems. may your self test it with fresh install to heroku. in my own server still using PHP 8.0 for production but at heroku using PHP 8.1 for development.

just wait others to test it.

mahagr:
@mdestafadilah Error should be fixed now, but I recommend installing the PHP extension as it is much faster.

Rotzbua:
@mdestafadilah Stop pointing at me. Stop trying to make your problems to my problems. If you are too lazy to create a proper issue it is your fault.

PR #3533, opened 2022-02-11
Add Vector image auto_sizes support

xf-:
Read images sizes from vector files to be more consistent with "photo" images.


mahagr:
I made some minor improvements to improve error handling.

xf-:
@mahagr i used that code nested and adjusted it more or less to your code style in grav and removed some PHP8 stuff like `$xml?->` and next line checks if it is an element :)

I thought about using try catch and also xml is requirement in composer. Normally every webserver should use it and avoid inconsistant behaviour. i noticed mbstring is suggested in composer and index.php basiclly stops without it.

mahagr:
I didn't have any PHP8 code, `$attr = $xml ? $xml->attributes() : null;` works in PHP5 and is required as loading the file may fail (non-XML file).

mahagr:
Fixed mb extension issue, too.

PR #3535, opened 2022-02-15
Merge index exceptions with same handling

xf-:
Tested with PHP 7.3 and 8.1

### DIFF
--- a/index.php
+++ b/index.php
@@ -45,10 +45,7 @@ $grav = Grav::instance(array('loader' => $loader));
 // Process the page
 try {
     $grav->process();
-} catch (\Error $e) {
-    $grav->fireEvent('onFatalException', new Event(array('exception' => $e)));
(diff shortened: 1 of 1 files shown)


mahagr:
I'm not going to approve this as the index file needs to compile in PHP 5 to display a proper error to the user if he accidentally used an older version of PHP.

xf-:
@mahagr Line 16 with PHP 7.0 so this file won't work in PHP 5 (`??` syntax) and 65 is long time EOL and should run for last 3 years on anything production related. `exception|exception` was added in 7.1 and 7.2 is much more common of all the EOL PHP versions.

mahagr:
Hmm, true. We haven't had any complaints on it lately and that has been there for a few releases... So OK, it looks like people aren't trying to install Grav to PHP 5 servers anymore.

PR #3546, opened 2022-02-27
Add phpdocs with deprecated

Rotzbua:
Needed for IDE and linter support.

### DIFF
--- a/system/defines.php
+++ b/system/defines.php
@@ -67,63 +67,83 @@ if (!defined('GRAV_BACKUP_PATH')) {
 unset($path);
 
 // INTERNAL: Do not use!
+/**
+ * @internal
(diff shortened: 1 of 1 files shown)


mahagr (changes requested):
I would make them internal instead of deprecated for now. Some of those are still being used before Grav initialization.

Rotzbua:
I do not understand.

Your own comment claims:
https://github.com/getgrav/grav/blob/3dd0cabeac9835fe64dcb4b68c658b39f1f6be2f/system/defines.php#L73

Deprecation just means that the variables should not be used for new code and old code should be refactored in the future.
I see no contradiction between the current use and documenting a deprecation.

PR #3501, opened 2021-12-25
Remove security and privacy issue Gravatar #3491

Rotzbua:
Fixes:  Remove security and privacy issue Gravatar #3491

PR #3550, opened 2022-03-06
Optional gravatar

Rotzbua:
Improved version of  #3501 with optional activation to restore feature.

### DIFF
--- a/system/config/system.yaml
+++ b/system/config/system.yaml
@@ -15,6 +15,7 @@ http_x_forwarded:                                # Configuration options for the
   host: false
   port: true
   ip: true
+avatar_use_gravatar: false                       # If enabled and no custom user avatar is set, then gravatar is used as external avatar source. Possibly relevant to your privacy policy.
 
 languages:
   supported: []                                  # List of languages supported. eg: [en, fr, de]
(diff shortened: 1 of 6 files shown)


system/config/system.yaml
```diff
@@ -15,6 +15,7 @@ http_x_forwarded:                                # Configuration options for the
   host: false
   port: true
   ip: true
+avatar_use_gravatar: false                       # If enabled and no custom user avatar is set, then gravatar is used as external avatar source. Possibly relevant to your privacy policy.
```
mahagr, line 18: To me, this is an issue. We cannot change the existing behavior in a patch release without consent from the site owner. If we do that, it's considered to be a bug.

For the new sites this change is good, but if this will be the default, we need to update configuration in the existing sites to set the variable to true.

mahagr:
If we decide to go this route, I can write an update script that sets the new option to true in existing sites. I tend to agree with this; we should give the choice to the administrator (without needing to guess how to do it in the template overrides).

rhukster:
I propose we drop this for my new Multiavatar-by-default PR: https://github.com/getgrav/grav/pull/3551

rhukster:
We've added Multiavatar... right now Gravatar is still default.  We'll switch to Multiavatar as default in Grav 1.8 when we can warn people about the change and how it impacts them.

PR #3562, opened 2022-03-17
Include modulars for last modification date computation

### LINKED ISSUE #3561: Grav serving stalled content: Last-Modified ignore modular pages last modification date
In #2168 someone requested that the template modification be taken into account when computing the Last-Modified header.
It happens that the time computed from this header is based exclusively on the **main page** last modification time **even if has multiple (more recent) children modular pages**.

The consequence is that when someone reaches such a Grav page their browser first check `Cache-Control: max-age` (see #3560) and, if the content is deemed expired, the browser will send a `If-Modified-Since` header (using the `Last-Modified`) date.

If any caching-proxy exist between the user and Grav, it's likely to send a 304 `Not Modified` (because of the wrong `Last-Modified` date) **leading the stale content being served to the users**.

The consequence depends on the browsers but I've got reports pointing this pervasive behavior affecting many users (including our own editors).
Some browser vendors may simply not send the `If-Modified-Since` is loading the URL from the address-bar or from a the `Back` button, or from a `hard-refresh` or if the debugger is opened. But most browser regularly visiting production-grade Grav instances are suffering this issue.

drzraf:
Fix #3561

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -608,7 +608,15 @@ class Page implements PageInterface
 
         // Set Last-Modified header
         if ($this->lastModified()) {
-            $last_modified_date = gmdate('D, d M Y H:i:s', $this->modified()) . ' GMT';
+            $last_modified = $this->modified();
+            foreach ($this->children()->modular() as $cpage) {
+                $modular_mtime = (int)filemtime($cpage->filePath());
+                if ($modular_mtime > $last_modified) {
+                    $last_modified = $modular_mtime;
+                }
+            }
+
+            $last_modified_date = gmdate('D, d M Y H:i:s', $last_modified) . ' GMT';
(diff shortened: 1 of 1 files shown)


system/src/Grav/Common/Page/Page.php
```diff
@@ -608,7 +608,15 @@ public function httpHeaders()
 
         // Set Last-Modified header
         if ($this->lastModified()) {
-            $last_modified_date = gmdate('D, d M Y H:i:s', $this->modified()) . ' GMT';
+            $last_modified = $this->modified();
+            foreach ($this->children()->modular() as $cpage) {
+                $modular_mtime = (int)filemtime($cpage->filePath());
```
rhukster, line 613: is there a reason you didn't use $cpage->lastModified() ??

rhukster:
cheers.. tweaked slightly

PR #3661, opened 2022-12-24
Fixed exception: "Property 'jsmodule_pipeline_include_externals' does not exist in the object!"

vitaldevru:


rhukster:
Going through older PRs.. thanks for this.

PR #3662, opened 2022-12-25
add php 8.2 to test

Rotzbua:
now released https://www.php.net/releases/8.2/en.php


rhukster:
tests pass, bu there are some issues with admin i need to resolve.

PR #3667, opened 2022-12-28
Handle the situation when GRAV_ROOT or GRAV_WEBROOT are `/`

thebodzio:
I am using a hosting service which separates the site files in such a way that both GRAV_ROOT and GRAV_WEBROOT are `/`. Unfortunately, this tends to break some of the Grav and plugin's functionality—most importantly backups and upgrades (in grav-plugin-admin).

This is caused by the use of `rtrim` to remove trailing slashes from path names. When GRAV_ROOT or GRAV_WEBROOT are `/` this effectively sets both to an empty string, which, in turn, makes the aforementioned scenarios to fail miserably.

The fix is very simple. Whenever GRAV_ROOT or GRAV_WEBROOT are to be evaluated to an empty string (it can only happen if their paths contain only `/` characters), they are arbitrarily set to `/`.


AnaelMobilia:
@thebodzio ,
In order to be compatible with Windows, you should maybe use `DIRECTORY_SEPARATOR` instead of `/`.
Best regards,
Anael

thebodzio:
@AnaelMobilia,

Thanks for the suggestion, however, I don't think this will ever be an issue. First, I don't think we will have a Windows path starting with `/`—that is unless some specific “mapping” will be done by this or that piece of software in order to “hide” drive letters. Second, it would be counterproductive, since one line before all `DIRECTORY_SEPARATOR`s were intentionally replaced by `DS`s.

That being said, I do agree that it would be much more elegant to consistently use `DS` since it was pretty much globally defined for the whole Grav. I will modify my commit accordingly.

Have a great day!

thebodzio:
Corrections have been made. Awaiting your kind consideration 👍😊

PR #3669, opened 2023-01-02
Update php-fpm socket file to use generic socket file

idanoo:
I've had to manually change this on recent installs on Debian Ubuntu and Arch. All php-fpm packages (That I have observed) create a generic php-fpm.sock file which symlinks to the latest installed PHP version. 

PHP 7.X is EOL so I believe this should be a trivial change.


rhukster:
Cheers

PR #3684, opened 2023-02-08
Fix LogViewer regex to reduce greediness, closes #3683

### LINKED ISSUE #3683: Bug: LogViewer parser regex is greedier than it should be
The [regex used by the `Grav\Common\Helpers\LogViewer` class](https://github.com/getgrav/grav/blob/c732bfaeefaa56c1c679ba848349f01da706a6e8/system/src/Grav/Common/Helpers/LogViewer.php#L24) to parse lines from the `grav.log` is too greedy, which means that some log lines are not parsed as expected.

For example, the following lines
```
[2023-02-08 05:05:03] grav.DEBUG: [IP Blacklist] Request matched filter '/\.env': 'https://cooper-davis.net/.env' [] []
[2023-02-08 05:05:03] grav.DEBUG: [IP Blacklist] IP blacklisted: '144.48.81.111' [] []
[2023-02-08 05:05:04] grav.DEBUG: [IP Blacklist] IP reported: '144.48.81.111' [] []
[2023-02-08 05:05:04] grav.DEBUG: [IP Blacklist] Request rejected: 'https://cooper-davis.net/.env' [] []
```
are parsed and displayed in the admin log viewer tool as

image

This is any easy fix; I'll submit a pull request.

aricooperdavis:
By making the date match group non-greedy, and escaping the dot between the logger and level match groups, the regex behaves better for log entries that contain square brackets and colons.

See #3683 for full bug report.

### DIFF
--- a/system/src/Grav/Common/Helpers/LogViewer.php
+++ b/system/src/Grav/Common/Helpers/LogViewer.php
@@ -21,7 +21,7 @@ use function is_string;
 class LogViewer
 {
     /** @var string */
-    protected $pattern = '/\[(?P<date>.*)\] (?P<logger>\w+).(?P<level>\w+): (?P<message>.*[^ ]+) (?P<context>[^ ]+) (?P<extra>[^ ]+)/';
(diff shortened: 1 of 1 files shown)


rhukster:
cheers.

PR #3702, opened 2023-03-17
Fix force_ssl redirect in case of undefined hostname

ViliusS:


rhukster:
Looks solid.  Thanks.

Issue #1622, opened 2017-08-24
Markdown strips escape sequences from image urls


j000:
Escape sequences should be left alone. The same thing does not happen to links.
```
![Failed](http://example.com/?q=a%20b+c%2bd)
```
will render as 
```
<img src="http://example.com/?q=a b c+d" alt="Failed" />
```
The same sequence works fine on http://parsedown.org/demo: 
```
<img src="http://example.com/?q=a%20b+c%2bd" alt="Failed" />
```

rhukster:
I'll have a look at this tomorrow.  Thanks.

rhukster:
Should be fixed, please try develop branch

dirkjf:
This problem appears to be back.

rhukster:
I can't recreate it.. tried your example:

```
![Failed](http://example.com/?q=a%20b+c%2bd)
```

and it renders as:

```
<img src="http://example.com/?q=a%20b+c+d" alt="Failed">
```

With latest Grav `1.7.41.1`

rhukster:
well it changed the %2b to the `+` but that's about it.

dirkjf:
I should have specified, the query params render fine, but the path does not. So if you try: `![image](image%20wtih%20spaces.jpeg)`, it will render: `<img src="/typography/image wtih spaces.jpeg" alt="image">` instead of `<img src="/typography/image%20wtih%20spaces.jpeg" alt="image">`. I've created a PR that should fix this.

rhukster:
I did also try a filename with spaces in it like the example you had.  And it did not convert those spaces to %20 and was broken when rendering. However, I think it's always been this way, at least I went back version to a couple of years ago and it  didn't work there either.

I will review your PR, thanks!

PR #3719, opened 2023-05-18
fix whitespace encoding in urls

dirkjf:
Keep encoded links encoded. With this PR markdown like this: 

```
![internal](image%20wtih%20spaces.jpeg)
![exernal](https://examle.com/image%20wtih%20spaces.jpeg)
```

Will result in:
```
<img src="image%20wtih%20spaces.jpeg" alt="internal">
<img src="https://examle.com/image%20wtih%20spaces.jpeg" alt="exernal">
```

Instead of not: 
```
<img src="image wtih spaces.jpeg" alt="internal">
<img src="https://examle.com/image wtih spaces.jpeg" alt="exernal">
```

See also issue  #1622

### DIFF
--- a/system/src/Grav/Common/Page/Markdown/Excerpts.php
+++ b/system/src/Grav/Common/Page/Markdown/Excerpts.php
@@ -244,6 +244,7 @@ class Excerpts
             $id = $element_excerpt['id'] ?? '';
 
             $excerpt['element'] = $medium->parsedownElement($title, $alt, $class, $id, true);
+            $excerpt['element']['attributes']['src'] = str_replace(' ', '%20', $excerpt['element']['attributes']['src']);
         } else {
             // Not a current page media file, see if it needs converting to relative.
             $excerpt['element']['attributes']['src'] = Uri::buildUrl($url_parts);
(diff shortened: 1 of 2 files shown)


system/src/Grav/Common/Page/Markdown/Excerpts.php
```diff
@@ -159,20 +160,13 @@ static function ($carry, $item) {
...
                 unset($url_parts['stream'], $url_parts['scheme']);
             }
-
-            $excerpt['element']['attributes']['href'] = Uri::buildUrl($url_parts);
-
-            return $excerpt;
         }
 
-        // Handle paths and such.
-        $url_parts = Uri::convertUrl($this->page, $url_parts, $type);
```
rhukster, line 172: This call to `Utils::convertUrl()` is important as it's waht handles the relative pathing calculation.  It can't simply be removed.

rhukster (commented):
Without even running any tests, I know this will be an issue because you've removed an important bit of logic.

system/src/Grav/Common/Page/Markdown/Excerpts.php
dirkjf: I will have a look at it.

dirkjf: I see what you mean, I have updated to only encode whitespaces in the paths.

dirkjf:
> Without even running any tests, I know this will be an issue because you've removed an important bit of logic.

This is restored and currently this PR will only encode whitespaces in a URL.

PR #3724, opened 2023-06-08
Fix invalid input to foreach

ricardo118:
This happens with discord oauth, possibly others


rhukster (commented):
Thanks for the fix.  I actually prefer a slightly different approach for these scenarios:

```php
$user_groups = Grav::instance()['user_groups'] ?? [];
```

That way the original loop can be left without adding more nesting.

rhukster (changes requested):
Thanks for the fix.  I actually prefer a slightly different approach for these scenarios:

```php
$user_groups = Grav::instance()['user_groups'] ?? [];
```

That way the original loop can be left without adding more nesting.

PR #3759, opened 2023-10-24
Make Inflector::ordinalize() work

jeffwjensen:
It's day 2 of learning Grav. I noticed the ordinalize example wasn't working on the docs https://learn.getgrav.org/17/themes/twig-tags-filters-functions/filters#ordinalize 

I almost understand 0.1% of the code, but it feels right.


rhukster:
Eeek, not sure how that happened! Thanks for catching it.

Issue #3770, opened 2023-11-07
[FeatureRequest] [Scheduler] Add an option to the scheduler command to run past due jobs


maelanleborgne:
Hi, I'm hosting one of my Grav project on OVH, which doesn't allow to set the cron 'minute' field for the cron periodicity. That means We can only get the cron to run every hour at a fixed minute time.
I think for this use case it would be great to have the option to run the scheduler command with an option to make it run jobs that are past due (if nextRunDate <= currentTime). 
I'd suggest something like 'catchup' or 'overdue' for the option name, but I'm open to any suggestion :)

Let me know if you like the idea and I'll submit a PR when I get some time to work on this.

rhukster:
I'm open to this if it's an optional setting, and defaults to off (i.e. previous behavior of not running past due).  Cheers!

PR #3771, opened 2023-11-07
[Draft] [Scheduler] Added 'outdated' option to scheduler command

maelanleborgne:
## Description

This PR follows #3770 and adds an `outdated` option to the scheduler command.

This allows the scheduler to run jobs that should have been processed earlier but did not (use cases : server downtime, cron not running every minute).

### Usage

``` bash
php bin/grav scheduler --outdated
```

## Advice wanted

First of all if you have another idea for the option name I'm open to discussion, I don't think `outdated` is really relevant.

Now the main concern :
While working on this feature I realized the Job `creationTime` is not stored in the config, but rather set when instantiating the job. The logic I had in mind involved checking if the job had been created later than the previous scheduled execution, to prevent running a job before its first scheduled run.
https://github.com/getgrav/grav/commit/4bb4069f5f92d1c149356c049673e112d4cce7b3#diff-d62f46dc32787f2a5ecbcca58aa4601fc942e8b293a6cfa7a9c6286475cbac02R212

I see a few options to work around this :
- Removing this test. The downside is that running the scheduler with the 'outdated' option may result in running new jobs to early for their first run. 
- Storing the `creationTime` in the `scheduler.yaml` config. This would work for custom jobs, but the default jobs don't store their config there so I don't think it would be such a good idea.
- Create a new scheduler state in addition to success and error ('new' for example) and for each newly created/modified jobs, add an entry to `data/scheduler/status.yaml` with `state: new` and `last-run: previous-scheduled-time`. This way we can keep track of every change in the scheduler configuration and make sure we don't run the jobs before the next occurrence.

IMO the third option would be the best to cover all cases, but I'd be happy to hear your thoughts about this :)

### DIFF
--- a/system/src/Grav/Common/Scheduler/Job.php
+++ b/system/src/Grav/Common/Scheduler/Job.php
@@ -201,6 +201,28 @@ class Job
     }
 
     /**
+     * Check if the Job should have run previously.
+     *
+     * @param  DateTime|null $date
+     * @return bool
+     */
+    public function isOverdue(?DateTime $date = null, ?DateTime $lastRun = null)
+    {
+        // If the time elapsed since the creation is inferior to the interval, it's not overdue
+        if ($this->creationTime > $this->executionTime->getPreviousRunDate($date)) {
+            return false;
+        }
+        // Else, if the job has never run, it's overdue
+        if (null === $lastRun) {
+            return true;
+        }
+        $date = $date ?? new DateTime('now');
+
+        // Else if the last run time is inferior to the previous scheduled time, it's overdue
+        return $lastRun < $this->executionTime->getPreviousRunDate($date);
+    }
+
+    /**
      * Check if the Job is overlapping.
      *
      * @return bool
--- a/system/src/Grav/Common/Scheduler/Scheduler.php
+++ b/system/src/Grav/Common/Scheduler/Scheduler.php
@@ -188,7 +188,7 @@ class Scheduler
      * @param DateTime|null $runTime Optional, run at specific moment
      * @param bool $force force run even if not due
      */
-    public function run(DateTime $runTime = null, $force = false)
+    public function run(DateTime $runTime = null, $force = false, $overdue = false)
     {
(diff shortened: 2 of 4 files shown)


rhukster:
I agree, the 3rd option would probably be the best solution, seems the most robust approach.

rhukster:
Thanks for this

rhukster:
@maelanleborgne i'm seeing tests fail:

```
Time: 00:01.018, Memory: 112.00 MB

There was 1 failure:

---------
1) JobTest: Is overdue | "hourly job created 1 hour ago and last run 30 mn ago"
 Test  tests/unit/Grav/Common/Scheduler/JobTest.php:testIsOverdue
Failed asserting that false matches expected true.
#1  /home/runner/work/grav/grav/tests/unit/Grav/Common/Scheduler/JobTest.php:14
#2  /home/runner/work/grav/grav/vendor/bin/codecept:119

FAILURES!
Tests: 219, Assertions: 1877, Failures: 1.
Error: Process completed with exit code 1.
```

maelanleborgne:
Hi Andy, I'm sorry but I did not get the time to finish this PR, been quite
busy at work lately. The PR was still marked as draft, do you think you
could revert the commit on the target branch and reopen the PR? I'll try to
make some time for this next week.

rhukster:
Weird, it's passing now.. was in the automated tests.. i couldn't replicate locally:

https://github.com/getgrav/grav/actions/runs/7421540097

Must of been a time related issue?

rhukster:
And appologies, i missed the DRAFT bit.. worked find in my quick local testing and tests ran fine... 

I'll revert though.

maelanleborgne:
> Weird, it's passing now.. was in the automated tests.. i couldn't replicate locally:
> 
> https://github.com/getgrav/grav/actions/runs/7421540097
> 
> Must of been a time related issue?

Yes I messed up the tests, will fix with the new implementation.
Sorry for the inconvenience, I'll make a new PR once it's fully operational.

PR #3772, opened 2023-11-08
Adds mime type for vCards (.vcf files)

JS-Media-Creation:
Adds support for vCards (.vcf files) in case of e.g. scanning a qr-code with the direct url to the file, so that it can be downloaded. 

Only a thumb-vcf.png should be added then too.


rhukster:
thanks

PR #3796, opened 2024-01-29
New Trait for decoding attribute in images

pmoreno-rodriguez:
The decoding property of the HTMLImageElement interface provides a hint to the browser as to how it should decode the image. More specifically, whether it should wait for the image to be decoded before presenting other content updates or not.
With this trait you can add the decoding attribute to load images in an easy way. It basically works the same way as the 'loading' attribute (see [Grav Docs](https://learn.getgrav.org/17/content/media#loading)), with `async`, `sync` and `auto ` options. More info in [Mozilla.org](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding)

### DIFF
--- a/system/blueprints/config/system.yaml
+++ b/system/blueprints/config/system.yaml
@@ -1300,6 +1300,17 @@ form:
                 auto: Auto
                 lazy: Lazy
                 eager: Eager
+            
+            images.defaults.decoding:
+              type: select
+              size: small
+              label: PLUGIN_ADMIN.IMAGES_DECODING
+              help: PLUGIN_ADMIN.IMAGES_DECODING_HELP
+              highlight: auto
+              options:
+                auto: Auto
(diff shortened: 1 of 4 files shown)


pmoreno-rodriguez:
Sorry for the oversight. Shortly, I will make the pull request from the admin plugin with the translations corresponding to the 'decoding' trait.

rhukster:
As yoiuy can see the latest commit is now showing a bunch of failed tests.  I think its becaues you the code is ALWAYS adding 'encoding="auto"' to every single link, even if you don't provide an encoding option.

pmoreno-rodriguez:
Thank you very much for your help. Now the 'decoding' attribute behaves the same as 'loading' and I hope there is no problem.

rhukster:
Looks good thanks!

PR #3821, opened 2024-04-24
Fix running Grav under Load Balancer with custom_base_url

### LINKED ISSUE #3511: custom_base_url vs assets base-url and page lookup
Whenever I setup a custom base url I can't get Grav to work (starting with 1.7)

I'm getting the opposite behavior described in #3238 

Setup:
`https://mydomain.com/foo/ -> [reverse proxy to http://127.0.0.4:8000/] -> Grav`

- With 1.6, setting `custom_base_url: /foo` was fine.
- With 1.7, I'm really blocked:
  - No `custom_base_url` : Assets are missing (looked for at `https://mydomain.com/` without `foo/`)
   - With `custom_base_url: /foo` : Pages are not found (even /foo/admin is not found)

**`custom_base_url` clearly affects pages retrieval** (and more generally, the whole routing process), but I can't find how to make it work.
Help very welcomed!

ViliusS:
Previous [change](https://github.com/getgrav/grav/commit/0af33850a68dd2e9d3e3a9e1bd2d25215b958eb5) broke running Grav under load balancer with custom_base_url set. All dynamic urls (pages, admin, etc.), returns 404. All static content is not affected.

dump($bits) just after [this parse_url()](https://github.com/getgrav/grav/blob/ee8d783d050588aaf35261b7d1832e2bdc433fe6/system/src/Grav/Common/Uri.php#L209)
**Original Grav 1.7.45 code accessed directly from http://localhost:8080/korteles/kodai**
```
array:3 [▼
  "scheme" => "http"
  "host" => "domain.com"
  "path" => "/korteles/kodai"
]
```
Result - OK.

**Grav 1.7.45 with this PR applied accessed directly from http://localhost:8080/korteles/kodai**
```
array:1 [▼
  "path" => "/korteles/kodai"
]
```
Result - OK.

**Original Grav 1.7.45 code accessed from load balancer https://mydomain.example.com/test/korteles/kodai (`custom_base_url: /test`)**
```
array:3 [▼
  "scheme" => "http"
  "host" => "domain.comhttps"
  "path" => "//mydomain.example.com/korteles/kodai"
]
```
**Result - 404 or even some nasty httpd cycles, depending on the configuration of load calancer.**

**Grav 1.7.45 with this PR applied accessed from load balancer https://mydomain.example.com/test/korteles/kodai (`custom_base_url: /test`)**
```
array:1 [▼
  "path" => "/korteles/kodai"
]
```
**Result - OK.**

Load balancer is configured to rewrite /test/(.*) to /$1, and I ensured that it is reaching Grav under correct path. The patch probably also fixes https://github.com/getgrav/grav/issues/3511 .

Also, it's not a good idea to add 3rd party domain into host value, which could be accidently reused some time in the future.

### DIFF
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -205,8 +205,8 @@ class Uri
         // set active language
         $uri = $language->setActiveFromUri($uri);
 
-        // split the URL and params (and make sure that the path isn't seen as domain)
-        $bits = parse_url('http://domain.com' . $uri);
+        // split the URL and params
+        $bits = parse_url($uri);
 
         //process fragment
         if (isset($bits['fragment'])) {


rhukster:
The `domain.com` domain is used only to force the domain to be populated so that the $uri an be parsed as a path, and it's there specifically to ensure that host header based attacks are not possible.  This is also the reason for the script/localhost tests.

With your change, it makes Grav vulnerable again to host-based attacks as it then allows the poor parsing in PHP's parse_url() to be fooled by escaped script tags.  The approach of using `domian.com` is simply to ensure that will not work. the host and scheme are not used after this point, that's why it's not an issue. However, i'm still not sure why your custom_base_url is getting tripped up. I think its something else going on.

ViliusS:
P.S. I also found custom `parseUrl()` method in Grav code itself. I see that it is used in other places, but not here. Not sure what is the purpose of it, but maybe it can be reused for better results.

ViliusS:
I'm not sure what should be put into initializeWithURL() but keep in mind, that `https://mydomain.example.com/test/korteles/kodai` gets rewritten by the load balancer to `https://mydomain.example.com/korteles/kodai`, only then processed by Grav. If I dump $uri object before parse_url() part I get `https://mydomain.example.com/korteles/kodai`.

ViliusS:
Maybe this will be clearer what I'm trying to do:
image

This basically serves as an automated infrastructure for different versions of the same Grav website to preview changes on infinite branches before these changes go into production. Setup is pretty simple, and as I said static elements (JS, CSS, Images) works just fine. It's just dynamic content which goes into redirect loop without my patch.

ViliusS:
P.S. I have tested your changes in https://github.com/getgrav/grav/commit/c97a0ffb161856c48571a3d79771f62acd4640f8 . It still doesn't work. I get 404 for all pages. Dumped objects in the screenshot is $uri before static::parseUrl() ant $bits after it:

image

ViliusS:
Is there anything else I could do to progress this PR?

PR #3844, opened 2024-07-12
Support to Fediverse Creator meta tag

pmoreno-rodriguez:
Added support to [Fediverse Creator meta tag](https://blog.joinmastodon.org/2024/07/highlighting-journalism-on-mastodon/)


rhukster:
merged. thanks.

PR #3856, opened 2024-10-21
Annoying output comments fix for json Flex

Sogl:
These comments break the json output.
Tired of commenting them every time after Grav update 😄


rhukster:
I was going to comment that it should also check content type for application/json rather than relying on extension.

PR #3862, opened 2024-10-26
Feature 3164- Aliases are not found during cli commands

jgonyea:
Adds aliases into the commands array during lookup.


jgonyea:
Ack! Wrong branch. Sorry

PR #3886, opened 2025-01-10
Update code blocks in README

odolbeau:
The goal of this PR is to removed the leading `$` in order to not have it copied when using the "copy button" automatically proposed by Github.

The `bash` activate syntax highlighting.

Before :
image

After :
image


rhukster:
cheers

Issue #1893, opened 2018-02-27
[BUG] js asset added inline doesn't respect order


jimblue:
Hi there,

It seems that when adding a `JS` file inline trough param of `Grav assets manager`, the order is not respected. `Inlined assets` are always rendered on the last position.

**e.g. this:**

```twig
{% block javascripts %}
  {% do assets.addJs('theme://js/manifest.js, {'priority':100, 'loading':'inline'}) %}
  {% do assets.addJs('theme://js/main.js', {'priority':10}) %}
{% endblock %}
{{ assets.js }}
```

**render into:**

```html
<script src="/user/themes/mytheme/js/main.js" type="text/javascript" defer=""></script>
<script>inlined content of manifest.js file</script>
```

It's a problem that js rendering order doesn't respect the given order in `Grav assets manager` params as it can break the js code.

Cheers

rhukster:
As discussed in slack, this is really a significant change, and i've marked it as an enhancement because it will require a fair bit of work that has backwards compatibility implications.  I'll look at it when i get a chance.

jimblue:
Thank you 😄

hughbris:
I've just encountered this error and set up a test for reporting it, then found this. I think it's important the order is respected. I don't think it's an enhancement, rather a bug.

_(I guess I can work around this using small assets groups or moving my inline assets to files. I'm making a rough interim port of a terrible WP site, so it's probably a poor setup to start with.)_

Since I've gone to the trouble, here are my test steps in case you need them:

Create a stub page at 99.test/assets.md ..

```
---
title: Test
never_cache_twig: true
---
Here's some content.
```

Create some asset files in the theme's `js` directory to output to console. Here's `js/test/file1.js`, repeat and modify for `file2.js` and `file3.js`.

```js
console.log('file 1');
```

Create a template `templates/assets.html.twig` in the theme ..

```twig
<html>
<head>
<title>Assets test only</title>

{% set inline_js1 %}
console.log('inline 1');
{% endset %}

{% set inline_js2 = "console.log('inline 2');" %}

{% block javascripts %}
	{% do assets.addInlineJs(inline_js1, 100) %}
	{% do assets.addJs('theme://js/test/file1.js', 90) %}
	{% do assets.addJs('theme://js/test/file2.js', 80) %}
	{% do assets.addInlineJs(inline_js2, 70) %}
	{% do assets.addJs('theme://js/test/file3.js', 60) %}
{% endblock %}

{% block assets deferred %}
[...]

PR #3924, opened 2025-05-26
Possible fix for inline JS being added out of sequence

hughbris:
Adds a new Assets test assertion to verify order of output per #1893. Fixes asset order by removing default `position=after` property in newly instantiated `inlineJs` objects.

Notes/limitations:

* May be incomplete because there are other asset types where `position=after` is set by default in constructor.
* Not clear if there was a good reason to set `position=after` by default, _or_ unintended consequences not setting it.
* Not actually clear on the intended purpose of asset `position` property vs `priority`.
* Though it's a bug fix restoring expected behaviour, this is a potentially breaking change for existing Grav sites. So might be wise to create a toggle setting in system assets config.

### DIFF
--- a/system/src/Grav/Common/Assets/InlineJs.php
+++ b/system/src/Grav/Common/Assets/InlineJs.php
@@ -26,7 +26,6 @@ class InlineJs extends BaseAsset
     {
         $base_options = [
             'asset_type' => 'js',
-            'position' => 'after'
         ];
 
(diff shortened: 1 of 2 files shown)


hughbris:
Converted this to draft PR after reading what that means. With the caveats listed above, I am sure this PR will need changing before it should be pulled.

PR #4101, opened 2026-05-31
security: validate ZIP entry paths in ZipArchiver::extract() to prevent Zip Slip

XananasX7:
## Summary

`ZipArchiver::extract()` calls `$zip->extractTo($destination)` without first validating that every entry path inside the ZIP archive stays within the intended destination directory.

A crafted ZIP file containing entries like:
- `../../config/security.yaml`
- `../../../../etc/cron.d/evil`
- `../user/config/system.yaml` (overwrites Grav configuration)

would be silently extracted outside $destination — a **Zip Slip** vulnerability.

## Attack Vector

Grav's package manager (`InstallCommand`, `SelfupgradeCommand`, `DirectInstallCommand`) uses `ZipArchiver` to extract theme, plugin, and core update packages. An attacker who can supply a malicious ZIP (e.g. via a compromised GPM package, a man-in-the-middle, or the admin panel's direct install) can overwrite arbitrary files on the server — including PHP files, configuration, cron scripts, or SSH authorized_keys.

## Fix

Before calling `extractTo()`, iterate over all entries with `ZipArchive::count()` / `getNameIndex()` and reject any entry whose resolved path does not begin with the canonicalized `$destination` prefix.

This is the same defensive pattern used by:
- **TYPO3** `ZipService::verify()`
- **PrestaShop** (our fix: #41578)
- **Joomla** template installer (our fix: #47859)
- **OWASP** Zip Slip guidance

## Impact

| | |
|---|---|
| **CWE** | CWE-22: Path Traversal |
| **Attack** | Crafted ZIP → arbitrary file write → RCE |
| **Precondition** | Attacker can supply a ZIP to any Grav ZIP-consuming code path |

## No Behaviour Change

All well-formed archives (entries that stay within the destination) are unaffected.

### DIFF
--- a/system/src/Grav/Common/Filesystem/ZipArchiver.php
+++ b/system/src/Grav/Common/Filesystem/ZipArchiver.php
@@ -13,6 +13,8 @@ use InvalidArgumentException;
 use RuntimeException;
 use ZipArchive;
 use function extension_loaded;
+use function realpath;
+use function str_starts_with;
 use function strlen;
 
 /**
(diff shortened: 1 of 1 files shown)


rhukster:
Thanks for catching this @XananasX7, the Zip Slip vector is real and worth closing off.

I went with a slightly different implementation. The proposed check leans on `realpath()`, which returns `false` for a path that does not exist yet, and that is exactly the attack case (creating a brand new file outside the destination). When `realpath()` fails, the fallback keeps the unresolved `../` segments, so `str_starts_with($entryPath, $destination)` is still true and the malicious entry slips through. It catches overwrites of files that already exist, but not new-file writes.

The version going into 2.0 resolves `.` and `..` lexically with no filesystem access, and also rejects absolute paths and Windows drive letters:

```php
protected function isSafeEntryPath(string $name): bool
{
    $name = str_replace('\\', '/', $name);
    if (str_starts_with($name, '/') || preg_match('#^[a-zA-Z]:#', $name)) {
        return false;
    }
    $depth = 0;
    foreach (explode('/', $name) as $segment) {
        if ($segment === '' || $segment === '.') {
            continue;
        }
        if ($segment === '..') {
            if (--$depth < 0) {
                return false;
            }
            continue;
        }
        $depth++;
    }
    return true;
}
```

This is called for every entry before `extractTo()`, and the extraction is aborted if any entry would escape the destination. You are credited in the CHANGELOG. Closing this PR since the fix is landing in 2.0, but thank you for the report and the references.

PR #4102, opened 2026-05-31
security: restrict allowed_classes in JobQueue::reconstructJob() to prevent PHP Object Injection

XananasX7:
## Summary

`JobQueue::reconstructJob()` calls `unserialize(base64_decode($item['serialized_job']))` without an `allowed_classes` restriction. The queue state is stored in a JSON file under the Grav data directory. An attacker who can write to that file (e.g. via a path traversal, insecure file permissions, or an admin-panel upload bypass) can inject a PHP Object Injection payload that instantiates arbitrary classes available in the autoloader during deserialization — potentially leading to Remote Code Execution via a gadget chain.

Note: the existing `if ($job instanceof Job)` check only validates the result **after** deserialization, so the gadget chain fires before the check rejects the payload.

## Fix

Pass `['allowed_classes' => [Job::class]]` to `unserialize()`. The serialized value is always a `Grav\Common\Scheduler\Job` object (serialized on line 95 in the same class), so this restriction is safe and does not change behaviour for legitimate queue entries.

## Impact

| | |
|---|---|
| **CWE** | CWE-502: Deserialization of Untrusted Data |
| **Attack** | Write crafted JSON to queue file → `unserialize()` instantiates gadget → RCE |
| **Precondition** | Write access to the Grav scheduler queue JSON file |

## No Behaviour Change

All legitimate queue entries (serialized `Job` objects) are unaffected. Invalid entries that previously returned `null` continue to return `null`.

### DIFF
--- a/system/src/Grav/Common/Scheduler/JobQueue.php
+++ b/system/src/Grav/Common/Scheduler/JobQueue.php
@@ -460,9 +460,10 @@ class JobQueue
     protected function reconstructJob(array $item): ?Job
     {
         if (isset($item['serialized_job'])) {
-            // Unserialize the job
+            // Unserialize the job; restrict to Job::class to prevent PHP Object Injection
+            // via a crafted queue backend entry that could trigger a gadget chain.
             try {
-                $job = unserialize(base64_decode($item['serialized_job']));
+                $job = unserialize(base64_decode($item['serialized_job']), ['allowed_classes' => [Job::class]]);
                 if ($job instanceof Job) {
                     return $job;
                 }


rhukster:
Thanks for this @XananasX7, and for the detailed writeup.

This vector is already addressed in Grav 2.0. As part of GHSA-vj3m-2g9h-vm4p (shipped in `2.0.0-beta.2`), `reconstructJob()` now HMAC-signs the serialized job blob and refuses to unserialize any queue item whose signature is missing or does not match. A tampered or attacker-planted queue file can no longer reach `unserialize()` at all, so the object-injection path is closed.

Your suggestion is still worth doing as defense in depth, so on top of the HMAC check I have restricted deserialization to the `Job` class:

```php
$job = unserialize($serialized, ['allowed_classes' => [Job::class]]);
```

That is going into 2.0. Thanks again for the report and the references, much appreciated.

PR #4141, opened 2026-06-18
Block Makefiles in .htaccess

3e33:
I'm using a Makefile in my own theme building, and had to add my own .htaccess to block it, but it would be better to add it to the global .htaccess which already includes other common scripting filetypes such as py, sh and bat. This change simply adds Makefile to that list.


rhukster:
Thanks, but stock Grav doesn't ship Makefiles, so this belongs in your theme's own .htaccess rather than core. Also note the rule matches file extensions after a dot, so Makefile wouldn't actually be caught by this pattern as written.

3e33:
Oops that was a silly mistake, glad you caught it. It might still be nice to have it in the main .htaccess because it's an easy thing to overlook.

Issue #732, opened 2016-03-16
Certain twig functions are deleting accented characters


bovisp:
I have been translating UI-related items in my blog and noticed that when I apply the `titleize` twig function, all accented characters are stripped out. Once I took this function out, the characters came back. A quick check showed that `camelize` behaved the same way. 

Strangely, `humanize` will not strip the characters out, but if an accented character is the first word of a phrase or sentence, Grav will not strip out the character, but will also not capitalize it either. As an example:

`échelle synoptique`

should be changed to

`Échelle synoptique`

but in fact remains

`échelle synoptique`

All other accented characters in `humanize` remain, just as they should.

`@Sommerregen` suggested that I try `underscorize` to test if the accented characters are stripped out, which they are.

rhukster:
will look into this.

bovisp:
Thanks @rhukster !

KrakenEXE:
I can look into this tonight if you haven't done this yet. What do you say @rhukster?

rhukster:
That would be appreciated

Sylk:
This is beyond my technical level. I apologize.

pmoreno-rodriguez:
Hi.
I have also had this problem recently. It seems that the problem is in the following lines of the Inflector.php file in the system/src/Grav/Common folder:
```
public static function titleize($word, $uppercase = '')
     {
         $uppercase = $uppercase === 'first' ? 'ucfirst' : 'ucwords';

         return $uppercase(static::humanize(static::underscorize($word)));
     }
```

If you delete the reference to the `underscorize` function, it works perfectly, but I don't know if that change will affect anything else.

pmoreno-rodriguez:
Hi again.
I think this code is a possible solution. Replace line 201 of Inflector.php to:

`$regex3 = preg_replace('/[^\p{L}^0-9]+/u', '_', $regex2);`

pmoreno-rodriguez:
This code seems to fix the problem:
`$regex3 = preg_replace('/[^\p{L}^0-9]+/u', '_', $regex2);`

PR #4143, opened 2026-06-18
Fix unicode handling issue in inflector method hyphenize

Xoriander:
## Summary

On a modular site the on-page menu anchor links are not shown correctly if they are containing Unicode letters like `ä`, `ü`, `ö` and so on. This does not have a negative effect on the behavior of the links in my tests but looks strange in the addressbar and on mouseover.

## Environment

- Grav v2.0.0-rc.9
- Firefox 151
- PHP 8.5
- Apache Server

## Steps to Reproduce

1. Create a modular site with activated on-page menu and at least one module.
2. Name the module `Käsekuchen`.
3. Go to the frontend and click `Käsekuchen` in the navbar. 
4. Look at the adressbar and see `https://my-grav-page/#k?-sekuchen`

The `?` you're seeing is likely how your system displays the corrupted/replaced character.
The regex isn't recognizing `ä` as a valid letter.

## Reason

The quark2 module template uses the inflector class method hyphenize 
https://github.com/getgrav/grav-theme-quark2/blob/develop/templates/modular.html.twig#L45
and the pattern `/[^\p{L}^0-9]+/` from this method
https://github.com/getgrav/grav/blob/2.0/system/src/Grav/Common/Inflector.php#L223-L233
uses `\p{L}` (Unicode letter property), but **`\p{L}` only works in PHP when the `u` flag is set** to enable PCRE Unicode mode.

Without the `u` flag, `\p{L}` is treated as a literal character class that doesn't match Unicode letters properly, causing `ä` to be replaced  by the corrupted/replaced character. 

## Solution

Add the **`u` flag** at the end of the regex:

```php
$regex4 = preg_replace('/[^\p{L}^0-9]+/u', '-', (string) $regex3);
```
The `u` flag tells PCRE to treat the pattern as UTF-8 and properly interpret Unicode escape sequences like `\p{L}`.

## Testcase

The following assertion was added to https://github.com/getgrav/grav/blob/2.0/tests/unit/Grav/Common/InflectorTest.php#L83-L89 to confirm the changes are working as expected:

```php
self::assertSame('wörter-mit-bindestrich-getrennt', $this->inflector->hyphenize('Wörter_mit_Bindestrich_getrennt'));
```

Without the **`u` flag** at the end of the regex this test fails. 

```diff
There was 1 failure:
1) InflectorTest: Hyphenize
 Test  tests/unit/Grav/Common/InflectorTest.php:testHyphenize
Failed asserting that two strings are identical.
- Expected | + Actual
@@ @@
-'wörter-mit-bindestrich-getrennt>>+'w�-rter-mit-bindestrich-getrennt'
```

With the **`u` flag** at the end of the regex all tests succeeded. 

`OK (748 tests, 2705 assertions)`

## Similar Issue
https://github.com/getgrav/grav/issues/732

### DIFF
--- a/system/src/Grav/Common/Inflector.php
+++ b/system/src/Grav/Common/Inflector.php
@@ -225,7 +225,7 @@ class Inflector
         $regex1 = preg_replace('/([A-Z]+)([A-Z][a-z])/', '\1-\2', $word);
         $regex2 = preg_replace('/([a-z])([A-Z])/', '\1-\2', $regex1);
         $regex3 = preg_replace('/([0-9])([A-Z])/', '\1-\2', $regex2);
-        $regex4 = preg_replace('/[^\p{L}^0-9]+/', '-', $regex3);
+        $regex4 = preg_replace('/[^\p{L}^0-9]+/u', '-', $regex3);
 
         $regex4 = trim($regex4, '-');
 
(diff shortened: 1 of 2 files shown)


rhukster:
cheers!

PR #4146, opened 2026-06-19
Replace maximebf/debugbar with php-debugbar/php-debugbar

erikn69:


rhukster:
1.7 is approaching EOL.  Grav 2.0 already has this dependency updated to php-debugbar/php-debugbar.

PR #4205, opened 2026-07-11
Add Twig Intl extension support

### LINKED ISSUE #4204: Advanced intl date twig functions are not available
## Problem

Grav supports php `intl` extension for better multi language support.
Unfortunately the advanced twig filters using `intl` are not available if the grav configuration is enabled.

## Background

With grav v1 it was possible to use the outdated functions with a plugin enabling the (now abandoned) twig extension.

Since grav v2 uses twig 3 which restructured the extra extensions. There is a dedicated twig intl extensions available.

## Implementation

Enable twig intl extensions if php intl is available and the grav configuration is enabled.

## Reference

- https://github.com/twigphp/intl-extra
- https://twig.symfony.com/doc/3.x/filters/format_datetime.html

Rotzbua:
fixes #4204

Tests are missing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,7 @@
         "symfony/process": "^7.0",
         "symfony/http-client": "^7.0",
         "twig/twig": "3.x-dev",
+        "twig/intl-extra": "^3.26",
         "monolog/monolog": "^3.0",
         "doctrine/collections": "^2.2",
         "nyholm/psr7-server": "^1.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "cbe7176dcb90b38c0bb15fe41e4b6987",
+    "content-hash": "31cfdc8effb44b51cb6871de34274549",
     "packages": [
         {
             "name": "antoligy/dom-string-iterators",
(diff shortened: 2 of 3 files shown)


rhukster (commented):
Thanks for putting this together, and for wiring the registration up cleanly behind `extension_loaded('intl')` and a config flag. The feature is genuinely useful and I want Grav to support it. My one blocker is how the package gets pulled in.

I don't want `twig/intl-extra` in core's `require`. It hard-depends on `symfony/intl`, which bundles a large chunk of ICU data (several MB in vendor), and in practice relatively few sites need intl date, number, or currency formatting. Core needs to stay lean, so I'm not comfortable shipping that to every install by default. The `-extra` in the name is just Twig's packaging convention, it doesn't make the dependency optional: anything in `require` lands on everyone.

What I'd prefer is to keep this core-supported but not core-included. Two ways to get there:

1. Minimal change to this PR. Move `twig/intl-extra` out of `require` and into `suggest` (which installs nothing, it just documents the option), and add a `class_exists()` guard to the registration so it's a clean no-op when the package isn't present:

```php
if (extension_loaded('intl')
    && class_exists(\Twig\Extra\Intl\IntlExtension::class)
    && $this->grav['config']->get('system.intl_enabled', false)) {
    $this->twig->addExtension(new IntlExtension());
}
```

Anyone who wants it then runs `composer require twig/intl-extra` and sets `system.intl_enabled: true`.

2. Better still, we ship this as a small official plugin, essentially a class loader for `symfony/intl` that registers the `IntlExtension` behind the same ext-intl and config guard. That keeps the dependency with the people who opt in and leaves core untouched. Happy to help author that.

Either path is fine with me, but the package can't go into core `require`. Could you also add the tests you flagged as missing? Once it's optional and covered I'm glad to get this in.

Rotzbua:
Thanks for your feedback and explanation 👍 
I think the solution with a separate plugin does not require any core changes so closing.

rhukster:
https://github.com/getgrav/grav-plugin-intl

PR #4209, opened 2026-07-16
Add HTTP language fallback mapping

mirkotomasek:
## Summary

Adds configurable fallbacks for unsupported browser languages during `HTTP_ACCEPT_LANGUAGE` negotiation.

```yaml
languages:
  supported: [en, bs, de, cs]
  http_accept_language: true
  http_accept_language_fallback:
    hr: [bs]
    sr: [bs]
    sk: [cs]
```

For this configuration, Croatian and Serbian browser preferences select Bosnian, while Slovak selects Czech. Only the configured supported languages become active languages and URL prefixes, so the fallback sources do not expose additional routes such as `/hr`, `/sr`, or `/sk`.

## Implementation

- Includes configured fallback sources in browser-language negotiation.
- Resolves the winning fallback source to the first configured target present in `languages.supported`.
- Preserves direct supported-language matches and browser quality ordering.
- Ignores malformed sources, mappings from already-supported sources, and unsupported targets.
- Adds the default configuration, system blueprint field, and changelog entry.

Regional browser tags are handled by the existing negotiation library; for example, `hr-HR` matches the configured `hr` fallback and `sr-Latn-RS` matches `sr`.

This setting applies only to browser negotiation. It does not turn fallback sources into supported URL language prefixes.

## Validation

- `php -l system/src/Grav/Common/Language/Language.php`
- Parsed `system/config/system.yaml` and `system/blueprints/config/system.yaml` with Symfony YAML.
- Checked `hr-HR -> bs`, `sr-Latn-RS -> bs`, and `sk-SK -> cs`.
- Checked browser quality ordering between direct and mapped languages.
- `git diff --check`

## Follow-up

The new `PLUGIN_ADMIN.HTTP_ACCEPT_LANGUAGE_FALLBACK*` blueprint labels need corresponding Admin2 translation strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 1. [](#new)
     * You can now filter, sort, and group a page's media by the values in their `.meta.yaml` metafiles directly in Twig, with new `filterBy`, `where`, `findBy`, `sortBy`, `groupBy`, and `withMeta` methods on `page.media`. Fixes [getgrav/grav#4200](https://github.com/getgrav/grav/issues/4200).
+    * Added per-language fallbacks for unsupported browser languages during `HTTP_ACCEPT_LANGUAGE` negotiation, allowing them to resolve to supported languages without exposing additional language routes.
 1. [](#bugfix)
     * [security] A page editor can no longer read arbitrary files from the server by pointing an image watermark at a traversal path such as `carrier.png?watermark=../secret.png`; an editor-supplied watermark path is now confined to the site's media, while operator-configured watermarks and stream URIs are unaffected ([GHSA-w3f4-8pj2-599w](https://github.com/getgrav/grav/security/advisories/GHSA-w3f4-8pj2-599w)).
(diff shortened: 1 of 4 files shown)


rhukster:
Thanks @mirkotomasek — merged into `develop` 🎉

Nicely scoped: browser-negotiation-only, no new URL routes, and it correctly prefers direct supported-language matches over fallback sources on equal browser quality. Modeling the blueprint field on `content_fallback` was the right call.

I pushed one small follow-up on top ([18cb0278b](https://github.com/getgrav/grav/commit/18cb0278b)): the negotiation library lowercases matched language types, but `getHttpLanguageFallbacks()` keyed the map by the raw config source. So a source with an uppercase script/region subtag (e.g. `sr-Latn: [bs]`) never matched the map and silently fell back to the default language — a realistic case when distinguishing script variants like `sr-Latn` / `sr-Cyrl`. Lowercasing the source key when building the map fixes it; the plain lowercase cases (`hr`, `sr`, `sk`) were unaffected either way.

Two follow-ups still open, if you're up for them:
- The `PLUGIN_ADMIN.HTTP_ACCEPT_LANGUAGE_FALLBACK*` labels need matching strings in `grav-plugin-admin` (you flagged this) — otherwise the admin UI shows raw keys.
- A unit test for the negotiation path (`hr → bs`, mixed-case source, malformed source, unsupported target, quality ordering) would lock the behavior in.

Appreciate the contribution!

PR #4218, opened 2026-07-20
Add new Twig filter/function array_group_by for grouping arrays

pmoreno-rodriguez:
#### **Description**

This PR introduces a new Twig **filter and function** named `array_group_by` that allows grouping arrays or Grav collections by a property, method, or callback result.

It provides a convenient way to organize content directly in Twig templates — for example, grouping pages by category, year, or author — without writing custom PHP logic.

#### **Usage Example**

```twig
{# Prepare collection with category #}
{% set collection_with_category = [] %}
{% for post in page.collection() %}
    {% set category = post.taxonomy.category|first ?: 'uncategorized' %}
    {% set collection_with_category = collection_with_category|merge([{
        post: post,
        category: category
    }]) %}
{% endfor %}

{# Group by category #}
{% set posts_by_category = collection_with_category|array_group_by('category') %}

{# Display grouped posts #}
{% for category, posts in posts_by_category %}
    <h3>{{ category|capitalize }}</h3>
    {% for item in posts %}
        <div>{{ item.post.title }}</div>
    {% endfor %}
{% endfor %}
```

### DIFF
--- a/system/src/Grav/Common/Twig/Extension/GravExtension.php
+++ b/system/src/Grav/Common/Twig/Extension/GravExtension.php
@@ -145,6 +145,7 @@ class GravExtension extends AbstractExtension implements GlobalsInterface
             new TwigFilter('wordcount', [$this, 'wordCountFilter']),
             new TwigFilter('json_decode', $this->jsonDecodeFilter(...)),
             new TwigFilter('array_unique', 'array_unique'),
+            new TwigFilter('array_group_by', [$this, 'arrayGroupByFilter'], ['needs_environment' => true]),
             new TwigFilter('basename', 'basename'),
             new TwigFilter('dirname', 'dirname'),
             new TwigFilter('print_r', $this->printRGuarded(...), ['needs_environment' => true]),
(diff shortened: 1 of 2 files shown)


rhukster:
Thanks for tidying this into a single-file change, and for the clear docblock — much easier to review.

A few things before we can take it:

1. **Twig sandbox (the blocker).** Grav ships an optional Twig sandbox (`security.twig_sandbox.enabled`) that guards untrusted content rendering, and we recently hardened several filters against a member-access bypass (advisory GHSA-mc5q-6hpj-rp7j). As written, `array_group_by` can (a) call an arbitrary method on each item by name in the string branch (`item->$callback()`) and (b) invoke an arbitrary array callable via `call_user_func` — both without going through the sandbox policy. Twig's own callback filters (`sort`/`map`/`filter`/`column`/`reduce`/`find`) all register with `'needs_is_sandboxed' => true` and reject any non-Closure callback in sandbox mode (see `CoreExtension::checkArrow`). Could you follow that same pattern: add `needs_is_sandboxed`, and in sandbox mode require a Closure and route any object member access through `GravSecurityPolicy`? Outside the sandbox the full behavior can stay, exactly like our existing dump-filter guard in the same file.

2. **Placement.** The method landed between `getCookie()`'s docblock and its signature, so `getCookie` ends up undocumented and its `@param` block is left dangling. Mind moving the method out of that spot? There's also some trailing whitespace our PSR-12 check will flag.

3. **Small correctness bits:** the `try/catch` that maps errors to `'undefined'` also hides real problems (and misses `\Error`); `isset()` treats a legitimately-null group value the same as a missing key (`array_key_exists` would separate them); and a non-scalar grouping value would trigger an "Illegal offset type" error.

4. The title mentions a filter **and** a function, but only the filter is registered — did you mean to add the function too, or should we drop it from the title?

5. Since this is a public Twig filter, a couple of unit tests in `GravExtensionTest.php` (group by key, by property, by closure, plus the sandbox-rejection case) would help it land.

For what it's worth, the feature itself is welcome — Twig core doesn't have a `group_by` filter, so this fills a real gap. It's mainly the sandbox handling we need to get right first. Happy to look again once that's in.

pmoreno-rodriguez:
Hi @rhukster,

I've implemented all your feedback:
- Sandbox security with `needs_is_sandboxed` (Closure-only in sandbox)
- Fixed method placement and trailing whitespace
- Better null/type handling with `array_key_exists` and proper error propagation
- Registered both filter and function
- Added 5 unit tests (array key, object property, closure, sandbox accept/reject)

Ready for review!

rhukster:
Cheers,

PR #4223, opened 2026-07-22
Update nginx.conf - anchoring location directives

santacruz-usa:
Modified 'location' directives in the stanza "## <Begin/End> - Security" to reflect intention captured in 'htaccess.txt': anchoring the start of some regular expressions used in location matching.

For sites served by nginx, this fixes an error message received in Admin2, selecting "Tools -> Logs".  

Worth noting - this behavior (start-anchored regex) existed in earlier versions, but was removed (perhaps inadvertently) in commit 8899b3ebb86ed14e5718178afe922482d9433a0f 

I believe the behavior of non-anchored expressions has been acceptable for quite a long time, since the vast majority of matches would have been equivalent (whether using anchored or non-anchored). However, with the new(ish) reliance on API calls (most notably Admin2), the api call to fetch logs ("/api/v1/system/logs/files") is caught by the non-anchored location check to deny access to certain files/folders: "/(\.git|cache|bin|logs|backup|tests)/.*$"  

Arguably, the case could be made for adding a higher-priority match expression for URLs matching "/api/v1" (or similar), and keeping location expressions as-is (no start anchor), but given that "htaccess", "lighttpd", and "web.config" are all similarly configured (start anchor), it would seem prudent to conform nginx to match.

### DIFF
--- a/webserver-configs/nginx.conf
+++ b/webserver-configs/nginx.conf
@@ -18,29 +18,29 @@ server {
 
     ## Begin - Security
     # deny all direct access for these folders
-    location ~* /(\.git|cache|bin|logs|backup|tests)/.*$ { return 403; }
+    location ~* ^/(\.git|cache|bin|logs|backup|tests)/.*$ { return 403; }
     # deny all direct access to these sensitive user folders, whatever the file type
-    location ~* /user/(config|env)/.*$ { return 403; }
+    location ~* ^/user/(config|env)/.*$ { return 403; }
     # allow avatar images under user/accounts to be served directly, whether stored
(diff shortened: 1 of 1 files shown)


rhukster:
Thanks for this — nicely traced. You're right: the `location` regexes lost their start-anchor in that 2015 rewrite, and the unanchored `/(\.git|cache|bin|logs|backup|tests)/` rule was catching the Admin2 logs API call (`/api/v1/system/logs/files`) because it matches `logs` mid-path. Anchoring them brings nginx back in line with `htaccess.txt`, `web.config`, and the lighttpd config, and leaving `.env` unanchored is exactly right. Merging.

PR #4240, opened 2026-08-14
CompiledFile: regenerate compiled cache instead of a fatal ParseError on a corrupt compiled file

### LINKED ISSUE #4239: CompiledFile::content() throws uncaught ParseError (fatal 500) on a corrupt compiled cache instead of regenerating
### Summary

`CompiledFile::content()` reads a page's compiled cache file (`cache/compiled/files/*.md.php`) via `include`. If that compiled file is **corrupt or partially written**, the `include` throws a **`ParseError`**. `ParseError` extends `Error`, not `Exception` — and the slow-path read is only guarded by the method's outer `catch (Exception)`, so the fatal escapes uncaught instead of the cache being regenerated. Under the API (Admin2 / Grav-API plugin) this surfaces as a **500**; on the front end it's a white-screen fatal.

The fast path in the same method already handles this correctly with `catch (Throwable)`; only the slow-path read was left unprotected.

### Environment

- Grav **2.0.14** (also present on `develop`, current tip `8fdf7ed`)
- PHP 8.3

### Where

`system/src/Grav/Common/File/CompiledFile.php`, in `content()`:

```php
// Fast path — already resilient to a broken compiled file:
if ($modified && is_file($cacheFilename)) {
    try {
        $cache = (array)include $cacheFilename;
        // ... validate ...
    } catch (Throwable) {
        // If the compiled file is broken, we can safely ignore the error and continue.
    }
}
// ...
// Slow path — NOT resilient: $file->content() include()s the same compiled file,
// but is only covered by the method's outer `catch (Exception)` further down:
$size = filesize($filename);
$cache = $file->exists() ? $file->content() : null;   // <-- ParseError escapes here
```

The outer handler is `} catch (Exception $e) { throw new RuntimeException(...); }`, which never catches the `ParseError`.

### Real-world trigger

Saving a page that forces a full page-tree rebuild is enough to hit a transiently-corrupt compiled cache. In our case, renaming a page's folder in Admin2 → the editor requests a preview token for the new route → `PagesController::previewToken()` → `enablePages()` → `Pages::init()` → `Pages::recurse()` reads every page's compiled cache while those caches are being regenerated concurrently:

```
API unhandled exception: syntax error, unexpected string content "O **Xen Project 4.22.0** foi l..."
  file: cache/compiled/files/4605e38230fcaf36fa46e84c2fb42c1b.md.php  line: 43

#0  RocketTheme\Toolbox\File\PhpFile->decode()
#1  RocketTheme\Toolbox\File\AbstractFile->content()
#2  RocketTheme\Toolbox\File\File->content()
#3  system/src/Grav/Common/File/CompiledFile.php(78): RocketTheme\Toolbox\File\PhpFile->content()
#4  RocketTheme\Toolbox\File\MarkdownFile->content()
#5  system/src/Grav/Common/Page/Page.php(553): MarkdownFile->frontmatter()
#6  system/src/Grav/Common/Page/Page.php(225): Page->header()
#7  system/src/Grav/Common/Page/Pages.php(2236): Page->init()
#8  Pages->recurse()  ... (recursion) ...
#11 system/src/Grav/Common/Page/Pages.php(1966): Pages->resetPages()
#12 Pages->buildRegularPages()  ->  #13 buildPages()  ->  #14 Pages->init()
#15 user/plugins/api/classes/Api/Controllers/PagesController.php(1899): Pages->enablePages()
#16 PagesController->enablePages()  ->  #17 PagesController->previewToken()
#18 ApiRouter->handleRoute()  ->  ... -> index.php
```

The page whose compiled cache was corrupt is unrelated to the one being edited — any page's momentarily-bad compiled cache takes down the whole request. The folder rename itself succeeds; only the follow-up request 500s (so the folder ends up renamed but the UI reports an error).

### Why the compiled file is momentarily corrupt

The compiled `.md.php` is valid PHP again after the fact (Grav regenerates it), so the corruption is transient — consistent with a partial/racing write while the page tree is rebuilt and caches are regenerated concurrently. Regardless of the exact cause, `CompiledFile::content()` is designed to fall back to regenerating from source when the compiled cache is unusable; today it only does so when the failure is an `Exception`.

### Minimal proof of the `Exception`/`Throwable` gap

```php
file_put_contents('/tmp/corrupt.php', "<?php\nreturn ['data' => 'O **Xen' unexpected];\n");
try { include '/tmp/corrupt.php'; }
catch (\Exception $e) { echo "caught\n"; }   // NOT reached — ParseError is an Error
// PHP Fatal error: Uncaught ParseError ...

try { include '/tmp/corrupt.php'; }
catch (\Throwable $e) { echo "caught: ".get_class($e)."\n"; }  // caught: ParseError
```

### Suggested fix

Guard the slow-path read with `catch (Throwable)` and treat a failure as a cache miss, so the existing regeneration path runs — mirroring the fast path. PR attached.

jonata:
Fixes #4239.

### Problem

`CompiledFile::content()` reads a compiled cache file (`cache/compiled/files/*.md.php`) on its slow path via `$file->content()`, which `include`s the file. When that compiled file is corrupt or partially written (e.g. a concurrent regeneration race while the page tree is rebuilt), the `include` throws a **`ParseError`**. `ParseError` extends `Error`, not `Exception`, so it is **not** caught by the method's outer `catch (Exception)` — it escapes as a fatal (a **500** under the API, a white screen on the front end) instead of the compiled cache being regenerated from source.

The fast path a few lines above already handles the exact same situation with `catch (Throwable)`; only the slow-path read was unguarded.

### Fix

Wrap the slow-path read in `catch (Throwable)` and treat a failure as a cache miss (`$cache = null`). The existing "cache isn't up to date (or is invalid)" branch immediately below then regenerates the compiled cache from the raw source — the intended fallback. No behavior change on the happy path.

```php
$size = filesize($filename);
try {
    $cache = $file->exists() ? $file->content() : null;
} catch (Throwable $e) {
    // corrupt/partial compiled cache (incl. ParseError, an Error) -> treat as a
    // cache miss and regenerate from source below, mirroring the fast path.
    $cache = null;
}
```

`Throwable` is already imported in this file.

### Why `Exception` doesn't catch it

```php
file_put_contents('/tmp/corrupt.php', "<?php\nreturn ['data' => 'O **Xen' unexpected];\n");
try { include '/tmp/corrupt.php'; } catch (\Exception $e) { echo "caught\n"; }
// -> PHP Fatal error: Uncaught ParseError (Exception does not catch it)
try { include '/tmp/corrupt.php'; } catch (\Throwable $e) { echo "caught: ".get_class($e); }
// -> caught: ParseError
```

### Observed impact

Renaming a page's folder in Admin2 triggered `PagesController::previewToken()` → `enablePages()` → a full `Pages::recurse()` that read every page's compiled cache; one unrelated page's momentarily-corrupt compiled cache produced an uncaught `ParseError` and a 500 (full stack trace in #4239). The rename itself succeeded, so the folder was renamed but the request errored.

### DIFF
--- a/system/src/Grav/Common/File/CompiledFile.php
+++ b/system/src/Grav/Common/File/CompiledFile.php
@@ -75,7 +75,17 @@ trait CompiledFile
                 }
 
                 $size = filesize($filename);
-                $cache = $file->exists() ? $file->content() : null;
+                try {
+                    $cache = $file->exists() ? $file->content() : null;
+                } catch (Throwable $e) {
+                    // A corrupt or partially written compiled cache file (e.g. from a
+                    // concurrent regeneration race) can throw while being read/included —
+                    // including ParseError, which is an Error and would otherwise escape
+                    // this method's outer `catch (Exception)` as a fatal. Treat it as a
+                    // cache miss and regenerate from the raw source below, mirroring the
+                    // fast-path `catch (Throwable)` above.
+                    $cache = null;
+                }
 
                 // Load real file if cache isn't up to date (or is invalid).
                 if (!isset($cache['@class'])


rhukster:
Merged, thank you — this was an excellent report and a clean fix. The `Exception` vs `Throwable` proof made it trivial to verify, and I confirmed `git log -S'catch (Throwable'` returns a single 2020 commit that hardened the fast path only, so line 78 had never been guarded on any release from 1.7 through 2.0.19.

One thing added on the way in: the corrupt-cache event is now logged, on both the fast and slow paths. Regenerating silently is the right behaviour, but with no trace a *recurring* corruption problem is invisible to an operator — the compiled file is valid PHP again by the time anyone goes looking, which is exactly the situation you were in. The logger is resolved defensively so reporting a degraded cache can never itself become the fatal we're recovering from.

You were also right about the cause. `AbstractFile::save()` takes a non-atomic `ftruncate` + `fwrite` branch whenever a lock handle is already open, and `CompiledFile` always calls `lock()` first — while readers `include` the file without taking a shared lock, so the writer's `LOCK_EX` excludes nothing. That's the real root cause and I'm raising it upstream against the toolbox separately. Your patch is the right fix for Grav regardless, since the reader should be resilient either way.

Shipping in 2.0.20.

PR #4242, opened 2026-08-16
Fix Page::url() concatenating absolute routes.canonical value (#4023)

### LINKED ISSUE #4023: canonical() does not return correct value, if routes -> canonical is set
If my page contains a specific entry for canonical route and the server url differs from that page, canonical returns a concatenated value, which breaks the sitemap and therefor other processes and plugins like seo-magic.

Example:

I have my page open with url:
`https://my-url.tld`

In my `page.md`, the following header is defined:
```
---

routes:
    canonical: https://www.my-url.tld
---

```

Because the prefixed `www` should be the canonical one.

In that case, the canonical() function returns a concatenated value of both:
https://my-url.tldhttps://www.my-url.tld

This will break further processes, relying on a correct sitemap.

Any help/fix appreciated! Thank you!

wakqasahmed:
## What changed

`Page::url($canonical = true)` always concatenated `routeCanonical()` onto `rootUrl()`. If a page's `routes.canonical` header is set to an **absolute** URL (a different scheme/host than the site itself, e.g. to force `www` or a canonical domain), this produced a broken, doubled URL:

```
routes:
    canonical: https://www.my-url.tld
```
resulted in `https://my-url.tldhttps://www.my-url.tld` instead of `https://www.my-url.tld`.

This fix checks whether `routeCanonical()` returned an absolute URL via `Uri::isExternal()` (already used elsewhere in the codebase for the same purpose) and, if so, returns it verbatim — bypassing the `rootUrl()` prefix entirely. Relative canonical routes (the common case, e.g. `routes.canonical: /some-alt-route`) are unaffected and keep going through `rootUrl()` as before.

### Scope note: `custom_base_url` / `absolute_urls`

Not applicable here — when `routes.canonical` is already a fully-qualified absolute URL, the author has explicitly specified the exact string to use as canonical, so there's nothing left for `custom_base_url`/`system.absolute_urls` to contribute; those settings only affect how the *relative* branch builds `rootUrl()`, which is untouched by this change.

Closes #4023

## Test plan

- Added `tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php` (Codeception/PHPUnit unit test), covering:
  - absolute `routes.canonical` returned verbatim (both `$include_host = false` and `true`)
  - relative `routes.canonical` still correctly prefixed with the root URL
  - unset `routes.canonical` falls back to the page's normal route (existing behavior unchanged)
- Ran, locally via Docker (`composer:2` + `php:8.3-cli`):
  - `vendor/bin/codecept run unit tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php` — 3/3 pass
  - `vendor/bin/codecept run unit tests/unit/Grav/Common/Page/` — all `Page`/`Pages`/`PageOrdering` tests pass (63 tests, 190 assertions); the only failure in that run is a pre-existing, unrelated `ImageMediumUrlTest` error (`Undefined constant "IMG_JPG"`) caused by the GD extension/constants not being available in the plain `php:8.3-cli` image used for local verification — unrelated to this change and reproducible on `develop` without this patch.

### DIFF
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -2130,7 +2130,12 @@ class Page implements PageInterface
         }
 
         if ($canonical) {
-            $route .= $this->routeCanonical();
+            $routeCanonical = $this->routeCanonical();
+            if (Uri::isExternal($routeCanonical)) {
+                return $routeCanonical;
+            }
+
+            $route .= $routeCanonical;
         } elseif ($raw_route) {
             $route .= $this->rawRoute();
         } else {
--- a/system/src/Grav/Common/Page/Page.php
+++ b/system/src/Grav/Common/Page/Page.php
@@ -2131,7 +2131,7 @@ class Page implements PageInterface
 
         if ($canonical) {
             $routeCanonical = $this->routeCanonical();
-            if (Uri::isExternal($routeCanonical)) {
+            if (is_string($routeCanonical) && Uri::isExternal($routeCanonical)) {
                 return $routeCanonical;
             }
 
--- a/tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php
+++ b/tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php
@@ -40,6 +40,37 @@ class PageUrlCanonicalTest extends \PHPUnit\Framework\TestCase
         self::assertSame('https://www.example.tld/about', $page->url(true, true));
     }
 
+    /**
+     * @dataProvider absoluteCanonicalProvider
+     */
+    public function testCanonicalUrlWithVariousAbsoluteShapesIsReturnedVerbatim(string $canonical): void
+    {
+        $page = $this->pages->find('/about');
+
+        $page->routeCanonical($canonical);
+
+        self::assertSame($canonical, $page->url(false, true));
+    }
+
+    public function absoluteCanonicalProvider(): array
+    {
+        return [
+            'https with path' => ['https://www.example.tld/about'],
+            'https origin only' => ['https://www.my-url.tld'],
+            'protocol relative' => ['//www.example.tld/about'],
+            'query and fragment preserved' => ['https://www.my-url.tld/a?b=c#frag'],
+        ];
+    }
+
+    public function testCanonicalUrlWithArrayRouteDoesNotCrash(): void
+    {
+        $page = $this->pages->find('/about');
+
+        $page->routeCanonical(['/some-route']);
+
+        self::assertIsString(@$page->url(false, true));
+    }
+
     public function testCanonicalUrlWithRelativeRouteIsStillPrefixedWithRootUrl(): void
     {
         $page = $this->pages->find('/about');
--- a/tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php
+++ b/tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php
@@ -1,0 +1,60 @@
+<?php
+
+use Codeception\Util\Fixtures;
+use Grav\Common\Grav;
+use Grav\Common\Page\Pages;
+
+/**
+ * Regression tests for getgrav/grav#4023: Page::url($canonical = true) must
+ * return an absolute `routes.canonical` value verbatim instead of
+ * concatenating it onto the site's root URL.
+ */
+class PageUrlCanonicalTest extends \PHPUnit\Framework\TestCase
+{
+    /** @var Grav $grav */
+    protected $grav;
+
+    /** @var Pages $pages */
+    protected $pages;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $grav = Fixtures::get('grav');
+        $this->grav = $grav();
+        $this->pages = $this->grav['pages'];
+        $this->grav['config']->set('system.home.alias', '/home');
+
+        $locator = $this->grav['locator'];
+        $locator->addPath('page', '', 'tests/fake/simple-site/user/pages', false);
+        $this->pages->init();
+    }
+
+    public function testCanonicalUrlWithAbsoluteRouteIsReturnedVerbatim(): void
+    {
+        $page = $this->pages->find('/about');
+
+        $page->routeCanonical('https://www.example.tld/about');
+
+        self::assertSame('https://www.example.tld/about', $page->url(false, true));
+        self::assertSame('https://www.example.tld/about', $page->url(true, true));
+    }
+
+    public function testCanonicalUrlWithRelativeRouteIsStillPrefixedWithRootUrl(): void
+    {
+        $page = $this->pages->find('/about');
+
+        $page->routeCanonical('/about-canonical');
+
+        $rootUrl = $this->grav['uri']->rootUrl(false) . $this->pages->baseRoute();
+
+        self::assertSame($rootUrl . '/about-canonical', $page->url(false, true));
+    }
+
+    public function testCanonicalUrlDefaultsToRegularRouteWhenUnset(): void
+    {
+        $page = $this->pages->find('/about');
+
+        self::assertSame($page->url(false, false), $page->url(false, true));
+    }
+}
--- a/system/src/Grav/Framework/Flex/Pages/Traits/PageRoutableTrait.php
+++ b/system/src/Grav/Framework/Flex/Pages/Traits/PageRoutableTrait.php
@@ -143,7 +143,12 @@ trait PageRoutableTrait
         }
 
         if ($canonical) {
-            $route .= $this->routeCanonical();
+            $routeCanonical = $this->routeCanonical();
+            if (is_string($routeCanonical) && Uri::isExternal($routeCanonical)) {
+                return $routeCanonical;
+            }
+
+            $route .= $routeCanonical;
         } elseif ($raw_route) {
             $route .= $this->rawRoute();
         } else {


system/src/Grav/Common/Page/Page.php
```diff
@@ -2130,7 +2130,12 @@ public function url($include_host = false, $canonical = false, $include_base = t
         }
 
         if ($canonical) {
-            $route .= $this->routeCanonical();
+            $routeCanonical = $this->routeCanonical();
+            if (Uri::isExternal($routeCanonical)) {
```
wakqasahmed: **Nit / robustness:** `routeCanonical()` on this class is not type-guaranteed to return a string, and `str_starts_with()` inside `Uri::isExternal()` is strict about it.

`Page::routeCanonical()` (`Page.php:2244-2255`) returns `$this->routes['canonical']` verbatim with no type check, so hand-written front matter like

```yaml
routes:
    canonical:
        - /some-route
```

yields an array. Before this change that produced a PHP *warning* ("Array to string conversion") and a cosmetically wrong URL; after it, `Uri::isExternal($routeCanonical)` throws `TypeError: str_starts_with(): Argument #1 ($haystack) must be of type string, array given` — a fatal, i.e. a whole-page 500 instead of a bad `<link rel="canonical">`. I verified both behaviours on PHP 8.3.

Notably the Flex counterpart already guards for exactly this (`PageRoutableTrait.php:290`: `return is_string($canonical) ? $canonical : $this->route();`), and its own setter at line 285 casts with `(array)$var`, so array-shaped `header.routes.canonical` is demonstrably reachable in this codebase.

One-token fix:

```php
if (is_string($routeCanonical) && Uri::isExternal($routeCanonical)) {
```

system/src/Grav/Common/Page/Page.php
```diff
@@ -2130,7 +2130,12 @@ public function url($include_host = false, $canonical = false, $include_base = t
...
+                return $routeCanonical;
```
wakqasahmed, line 2135: **Parity gap:** `Grav\Framework\Flex\Pages\Traits\PageRoutableTrait::url()` carries a byte-for-byte copy of the same bug and is not fixed here.

`system/src/Grav/Framework/Flex/Pages/Traits/PageRoutableTrait.php:145-146`:

```php
if ($canonical) {
    $route .= $this->routeCanonical();
}
```

Same `canonical()` wrapper (`->url(true, true, $include_lang)`, line 107), same `rootUrl()` concatenation at line 155. So on a site using Flex Pages, `page.canonical` with an absolute `routes.canonical` still produces `https://my-url.tldhttps://www.my-url.tld` after this PR lands — the reporter's symptom in #4023 would only be half-fixed depending on which page implementation is in play.

Worth applying the identical three-line change there (its `routeCanonical()` already returns `?string`, so no `is_string` guard needed) and adding one assertion so the two implementations don't drift further.

tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->pages->init();
+    }
+
+    public function testCanonicalUrlWithAbsoluteRouteIsReturnedVerbatim(): void
+    {
+        $page = $this->pages->find('/about');
+
+        $page->routeCanonical('https://www.example.tld/about');
+
+        self::assertSame('https://www.example.tld/about', $page->url(false, true));
```
wakqasahmed, line 39: **Test coverage:** the two shapes actually named in #4023 aren't the ones asserted here.

The issue reports `routes.canonical: https://www.my-url.tld` — a bare origin with **no path** — and protocol-relative `//host/path` is the other form `Uri::isExternal()` accepts. This test only covers `https://www.example.tld/about`.

I ran both against the patched build and they behave correctly (`https://www.my-url.tld` and `//www.example.tld/about`, both verbatim), so this is a coverage gap rather than a defect. Cheap to close by turning this into a data provider, which also pins `Uri::isExternal()`'s `//` branch against future edits:

```php
public static function absoluteCanonicalProvider(): array
{
    return [
        'https with path'    => ['https://www.example.tld/about'],
        'https origin only'  => ['https://www.my-url.tld'],
        'protocol relative'  => ['//www.example.tld/about'],
    ];
}
```

Also worth one assertion that a query string/fragment survives (`https://www.my-url.tld/a?b=c#frag`) — that case is a real behavioural *improvement* of the early return, since the relative branch still runs through `Uri::filterPath()` and percent-encodes `?` to `%3F`. Locking it in stops someone "helpfully" reinstating `filterPath()` on the new return path later.

wakqasahmed (commented):
### Independent review (cold-start, verified locally)

I verified this against a fresh `develop` clone at `b91a2fd` (the PR's base) using a `php:8.3-cli` image **with `ext-gd` enabled**, so the GD-related noise the description mentions is out of the picture.

**What I confirmed works**

- `Uri::isExternal()` (`system/src/Grav/Common/Uri.php:779-782`) matches `http://`, `https://` **and** `//`, so both shapes from #4023 are covered.
- Behavioural check on the patched build (`Page::url(true, true)` after setting `routes.canonical`):

  | `routes.canonical` | result |
  |---|---|
  | `https://www.my-url.tld` (exact issue case) | `https://www.my-url.tld` |
  | `//www.example.tld/about` (protocol-relative) | `//www.example.tld/about` |
  | `https://www.my-url.tld/a?b=c#frag` | preserved verbatim, query+fragment intact |
  | `/rel?b=c` (relative) | `http://localhost/en/rel%3Fb=c` — identical to pre-patch |

- **The new test is a genuine regression test, not a stand-in.** It drives the real `Pages::init()` → `Page::url()` path. Reverting only the `Page.php` hunk and re-running it fails with `'/enhttps://www.example.tld/about'` vs `'https://www.example.tld/about'` — i.e. it reproduces the exact concatenation shape from #4023, and passes only with the fix.
- **No regressions:** `vendor/bin/codecept run unit tests/unit/Grav/Common/Page/` → `OK (63 tests, 193 assertions)` on the PR branch; `OK (60 tests, 189 assertions)` on unpatched `develop`.
- **The `IMG_JPG` / `ImageMediumUrlTest` failure is confirmed environmental, not pre-existing breakage.** With GD compiled in, `ImageMediumUrlTest` passes on plain `develop` *without* this patch. It fails only because the stock `composer:2` / `php:8.3-cli` images ship without `ext-gd` (`composer install` there even refuses outright: *"Root composer.json requires PHP extension ext-gd but it is missing"*). Unrelated to this change — but the description would be more accurate saying "environmental, GD not installed" rather than "pre-existing".
- **The `custom_base_url` / `absolute_urls` scope-out is sound.** `custom_base_url` only ever mutates `Uri::$base`/`$root`/`$root_path` (`Uri.php:169-190`), which are consumed exclusively through `rootUrl()` — the branch this fix bypasses. Both settings describe *the site's own* base; an absolute `routes.canonical` is a complete author-specified URL to a different origin, so there is nothing for them to rewrite. I could not construct an edge case where an author-supplied absolute canonical should be re-based through `custom_base_url`.

**Findings** — three non-blocking items, left as inline comments: a missing `is_string()` guard that converts a warning into a fatal, an identical unfixed copy of the bug in the Flex page path, and a small test-coverage gap. None of them make the current diff wrong.

Not approving — no merge rights here; this is review feedback only.

system/src/Grav/Common/Page/Page.php
wakqasahmed: Fixed — added the `is_string($routeCanonical)` guard before calling `Uri::isExternal()` in `Page::url()`. A non-string `routes.canonical` now falls through to the pre-existing relative-route behavior (concatenation onto the base route, same as before this PR) instead of throwing a TypeError. Added a regression test (`testCanonicalUrlWithArrayRouteDoesNotCrash`) confirming no crash.

system/src/Grav/Common/Page/Page.php
wakqasahmed: Applied the identical early-return fix to `Flex\Pages\Traits\PageRoutableTrait::url()`. Since that trait's own `routeCanonical()` already guards with `is_string($canonical) ? $canonical : $this->route()`, it can never itself return an array, so the `url()` fix only needed the `is_string() && Uri::isExternal()` early-return (no additional guard needed there, matching your note).

tests/unit/Grav/Common/Page/PageUrlCanonicalTest.php
wakqasahmed: Added the data-provider test covering bare-origin (`https://www.my-url.tld`) and protocol-relative (`//www.example.tld/about`) shapes, plus a query/fragment-preserving case (`https://www.my-url.tld/a?b=c#frag`) to pin the early-return against `filterPath()` regressions, and a case asserting a non-string (array) `routes.canonical` no longer crashes. Full unit suite still green: 979 tests / 3133 assertions on PHP 8.3 with ext-gd.

rhukster:
Merged, thanks — clean diagnosis and a genuinely minimal fix. Verified before merging: 8/8 of your new tests, and 990 tests / 3154 assertions across the full unit suite with no failures.

A few things I want to note as deliberately right, since they're easy to mistake for noise later: the `is_string()` guard is load-bearing (classic `Page::routeCanonical()` returns the raw header value, so malformed YAML can hand back an array and `Uri::isExternal()` would throw a `TypeError`); skipping `Uri::filterPath()` on the early return is correct, since it's a *path* filter that percent-encodes `?` and `#`; and keeping the two hunks as byte-identical copies rather than extracting a helper matches how `PageSystemValidatorCommand` validates classic `Page` against Flex `PageObject` method by method.

Three things added on the way in:

- A CHANGELOG entry under the existing `v2.0.20` block. For future reference there's no PR template telling contributors this, so it's on us rather than on you.
- A docblock note that `$include_host` is ignored when the canonical is absolute. That's the right behaviour — an absolute canonical names a different origin and has no host-less form — but silently ignoring a parameter deserves saying out loud.
- A fix for the adjacent Flex `routeCanonical()` setter, which cast to `(array)` while the getter required `is_string()`, so a value set through it could never be read back. That's why the trait side was awkward to test; header-loaded canonicals were unaffected, so your fix worked in production either way.

I also strengthened `testCanonicalUrlWithArrayRouteDoesNotCrash` — it used `@` suppression plus `assertIsString`, so it passed even if a warning fired, and that guard is the one line preventing a new fatal. It now asserts the concrete fallback.

Shipping in 2.0.20.

PR #4243, opened 2026-08-16
Per-file mtime cache-busting for non-pipelined local assets (#4049)

### LINKED ISSUE #4049: Bug: `enable_asset_timestamp` does not invalidate cached assets on file changes
## Summary

The `system.assets.enable_asset_timestamp` setting is misleadingly named. When enabled, it appends a query string to asset URLs that implies per-file cache busting (a "timestamp"), but the value is a global cache key that only changes when Grav configuration or version changes — never when individual JS/CSS files change on disk.

Combined with long-lived `Cache-Control` headers, this causes browsers to serve stale assets indefinitely after deployment.

## Expected Behavior

When `enable_asset_timestamp: true`:

```
user/themes/july/js/script.js   (modified at 14:00)
  → script.js?1774017600          (or similar mtime-based hash)

user/themes/july/js/script.js   (modified at 15:00)
  → script.js?1774021200          (different value, cache invalidated)
```

A developer edits `script.js`, deploys, and browsers fetch the new version.

## Actual Behavior

When `enable_asset_timestamp: true`:

```
user/themes/july/js/script.js   (modified at 14:00)
  → script.js?g-5fb7ccf4

user/themes/july/js/script.js   (modified at 15:00)
  → script.js?g-5fb7ccf4          (unchanged — same value)
```

The query string does not change. Browsers serve the old file from cache.

## Root Cause

`Grav\Common\Assets.php:159-161`:

```php
if ($this->enable_asset_timestamp) {
    $this->timestamp = Grav::instance()['cache']->getKey();
}
```

`getKey()` returns the global cache key, computed once in `Grav\Common\Cache.php:154-157`:

```php
$prefix = $this->config->get('system.cache.prefix');       // default: 'g'
$uniqueness = substr(md5($uri->rootUrl(true) . $this->config->key() . GRAV_VERSION), 2, 8);
$this->key = ($prefix ?: 'g') . '-' . $uniqueness;         // e.g. 'g-5fb7ccf4'
```

This hash is derived from:

- `rootUrl` — the site's base URL
- `config key` — fingerprint of the merged YAML configuration
- `GRAV_VERSION` — the Grav version string

**None of these change when a JS/CSS file is edited.**

## Why This Is Dangerous

Many production configurations — including Grav's own documentation recommendations — serve assets with the `g-` query string using aggressive caching headers:

```
Cache-Control: public, max-age=31536000, immutable
```

The `immutable` directive tells the browser: never revalidate, serve from cache until the URL changes. The browser trusts this because the URL contains what appears to be a unique, changing token (`g-5fb7ccf4`).

But the token doesn't change on file edits. The result:

| Scenario                                       | Browser behavior                 |
|------------------------------------------------|----------------------------------|
| User visits site, loads `script.js?g-5fb7ccf4` | Caches it for 1 year (immutable) |
| Developer deploys updated `script.js`          | New file on server               |
| User visits site again                         | Same URL `script.js?g-5fb7ccf4`  |
| Browser sees `immutable`, serves from cache    | **Stale JS served**              |
| Bug manifests: broken functionality            | User sees old behavior           |

The user has no way to recover except clearing browser cache manually. This can affect hundreds or thousands of users after every deployment that touches JS/CSS.

## Reproduction

1. Enable asset timestamps in `user/config/system.yaml`:
   ```yaml
   assets:
     enable_asset_timestamp: true
   ```

2. Note the URL of a JS file in page source:
   ```
   /user/themes/july/js/script.js?g-5fb7ccf4
   ```

3. Edit `user/themes/july/js/script.js` (add a comment or change functionality).

4. Reload the page. The URL is still:
   ```
   /user/themes/july/js/script.js?g-5fb7ccf4
   ```

5. Clear Grav cache (`bin/grav cache --all`). The URL is still:
   ```
   /user/themes/july/js/script.js?g-5fb7ccf4
   ```

6. The only way to change the URL is to edit a YAML config file or update Grav itself.

### Option A: Per-file mtime-based timestamp (recommended)

Replace the global cache key with per-file modification time:

```php
// In renderQueryString() or equivalent
if ($this->enable_asset_timestamp) {
    $file = GRAV_ROOT . '/' . $this->asset;
    if (file_exists($file)) {
        $this->timestamp = dechex(filemtime($file));
    }
}
```

This gives each asset a unique, content-aware hash that changes on every file edit.

I prefer human-readable (and easy to debug) timestamp strings:

```php
        $this->timestamp = date('Ymd-His', filemtime($file));
        // gives a nice `20260101-123456`
```

Ideally the cache string is a hash – `md5` or `xxHash/xxh3` (since PHP 8.1) – over the assets content: Same hash = same content = caching forever, live’s good.

### Option B: Hash of all asset mtimes

Compute a single hash from the modification times of all registered assets. Changes when any file changes, but not per-file. Simpler than A but coarser.

### Option C: Rename the feature

If the global cache key is intentional behavior, rename the config key and documentation to accurately describe what it does:

- `enable_asset_timestamp` → `enable_cache_key`
- Document that it invalidates on config/version change only
- Document that developers must use a build step to rename files or use content-hashed filenames

## Additional Notes

The name "timestamp" strongly implies `filemtime()` behavior. Any developer enabling this setting would reasonably expect it to bust cache on file changes. The current behavior silently fails to do so, creating a latent bug that only manifests in production with long cache headers.

The setting defaults to `false`, which may mask the issue during development (assets are not cached aggressively without the query string). Problems only appear when `enable_asset_timestamp: true` is deployed alongside immutable cache headers — exactly the scenario the feature is meant to enable.

wakqasahmed:
## Problem

Closes #4049.

When `enable_asset_timestamp` is on, `Assets` computes one global cache-busting
token (`Grav::instance()['cache']->getKey()`) and stamps it onto every
asset's URL as a `?<token>` query string. That token only changes when the
global config/cache key changes, so editing a single CSS/JS file's contents
does not invalidate that file's cached URL until something unrelated bumps
the global key — the bug reported in #4049.

## Change

For non-pipelined, local (non-remote) assets, `BaseAsset::init()` now
overrides the asset's own `$timestamp` property with `dechex(filemtime())` of
that specific file, right after the file is resolved and stat'd. The rewrite
only fires when `system.assets.enable_asset_timestamp` is actually enabled —
not just whenever a truthy timestamp happens to be present — so an explicit
token set via the public `Assets::setTimestamp()` API (independent of that
config flag) is preserved instead of being silently overridden. Everything
else about the existing timestamp plumbing (`Assets::$timestamp`,
`options['timestamp']`, `renderQueryString()`) is unchanged; this only
swaps the value that ends up in the per-asset property for this one case.

### Why pipelined and remote assets are effectively unaffected

- **Pipelined assets**: the *rendered* `<link>`/`<script>` tag's query string
  for a pipelined bundle comes from `Pipeline::$timestamp`, which is set from
  `Assets::$timestamp` (the global key) when the `Pipeline` object is
  constructed in `Assets::render()` — a separate property from the
  individual `BaseAsset` objects whose `$timestamp` this PR changes. So the
  pipelined *output URL* is unaffected.

  Correction from an earlier version of this description: the pipeline's
  internal bundle uid *does* consult this per-asset timestamp.
  `Pipeline::renderCss()`/`renderJs()` hash `json_encode($assets)`, and
  `BaseAsset::jsonSerialize()` includes the `timestamp` property via
  `getElements()`, so mutating it here does perturb the bundle filename hash.
  This is harmless in practice (bundle mtimes already flow into that same
  hash before this PR, and the effect here is a one-time bundle regeneration
  plus an orphaned old bundle file on upgrade, not a functional bug) — but
  it's not "no effect" as originally claimed here, so I'm correcting the
  record.
- **Remote assets**: `BaseAsset::init()` only enters the "local asset" branch
  when `!$this->remote`; remote URLs have no local file to `stat()`, so they
  keep using the existing global cache key.

## Test plan

`AssetsTest::testPerFileTimestamps()` now exercises the real
`enable_asset_timestamp` config flag (not just a manually set token), and
covers:
- two local CSS fixtures each get a distinct, `filemtime()`-derived query
  string instead of the shared global key
- touching one fixture only changes that fixture's own token
- a stream-based (`tests://`) local asset gets the same per-file rewrite
- a local JS asset gets the same per-file rewrite
- a remote asset keeps the unmodified global cache key
- a pipelined local asset's rendered query string keeps the unmodified
  global cache key

Added `AssetsTest::testManualTimestampPreservedWithoutConfigFlag()`: a token
set via `Assets::setTimestamp()` while `enable_asset_timestamp` is off (its
default) is preserved verbatim, proving the regression flagged in review is
fixed.

Fixture mtimes touched by the test are restored in a `finally` block so
repeated runs stay deterministic.

Ran via Docker (`composer:2` + `php:8.3-cli`, matching `composer.json`'s
`"php": "^8.3"`):

```
docker run --rm -v <worktree>:/app -w /app composer:2 install --no-interaction --ignore-platform-reqs
docker run --rm -v <worktree>:/app -w /app php:8.3-cli vendor/bin/codecept run unit Grav/Common/AssetsTest
docker run --rm -v <worktree>:/app -w /app php:8.3-cli vendor/bin/codecept run unit
```

- `AssetsTest`: 26/26 tests passing, 148 assertions.
- Full unit suite: 973 tests, 3060 assertions, 17 errors / 1 failure — all
  pre-existing and unrelated to this change (`ZipArchive` extension missing
  from the bare `php:8.3-cli` image used for `ZipBombSecurityTest`, and a
  `ComposerTest` path assertion that depends on the host filesystem layout).
  Same baseline as before this branch's changes (972/3057/17/1), plus the
  one new test method and its assertions.

## Scope note

Also updated the `enable_asset_timestamp` comment in
`system/config/system.yaml` to describe the new per-file-mtime behavior.
Not addressed in this PR (raised in review, left out as separable/low-risk
follow-ups): translating the admin-panel help string
(`PLUGIN_ADMIN.ENABLED_TIMESTAMPS_ON_ASSETS_HELP`) lives in the Admin
plugin's own language files, not in this repo.

### DIFF
--- a/system/src/Grav/Common/Assets/BaseAsset.php
+++ b/system/src/Grav/Common/Assets/BaseAsset.php
@@ -149,14 +149,20 @@ abstract class BaseAsset extends PropertyObject
                 // instead of the global config/version-derived cache key, so
                 // editing one file invalidates only that file's URL.
                 //
-                // Scope: this only affects the query string built for a single,
-                // individually rendered asset (this class's own renderQueryString()).
-                // Pipelined/bundled assets never consult this per-asset timestamp:
-                // Pipeline builds its own uid from a hash of the bundled asset
-                // list and content (see Pipeline::renderCss()/renderJs()), so it
-                // already has its own, finer-grained invalidation and is left
-                // using the global cache key set in Assets::render().
-                if ($this->timestamp && $this->modified) {
+                // Gated on the enable_asset_timestamp config flag itself (not
+                // just on $this->timestamp being truthy), so an explicit,
+                // caller-supplied token set via the public Assets::setTimestamp()
+                // API is never silently discarded when the config flag is off.
+                //
+                // Note: the rendered <link>/<script> tag for a *pipelined* bundle
+                // still uses the global cache key (Pipeline::$timestamp, set from
+                // Assets::render()), so this has no effect on pipelined output
+                // URLs. It does still feed into the pipeline's internal bundle
+                // uid hash (Pipeline::renderCss()/renderJs() hash the serialized
+                // BaseAsset objects, which include $timestamp) — harmless, but
+                // it does mean upgrading can produce one new bundle file instead
+                // of reusing the previous one.
+                if ($this->timestamp && $this->modified && $config->get('system.assets.enable_asset_timestamp')) {
                     $this->timestamp = dechex($this->modified);
                 }
             }
--- a/system/src/Grav/Common/Assets/BaseAsset.php
+++ b/system/src/Grav/Common/Assets/BaseAsset.php
@@ -144,6 +144,21 @@ abstract class BaseAsset extends PropertyObject
                 $asset = $this->buildLocalLink($file->getPathname());
 
                 $this->modified = $file->isFile() ? $file->getMTime() : false;
+
+                // Rewrite the cache-busting token to this asset's own filemtime
+                // instead of the global config/version-derived cache key, so
+                // editing one file invalidates only that file's URL.
+                //
+                // Scope: this only affects the query string built for a single,
+                // individually rendered asset (this class's own renderQueryString()).
+                // Pipelined/bundled assets never consult this per-asset timestamp:
+                // Pipeline builds its own uid from a hash of the bundled asset
+                // list and content (see Pipeline::renderCss()/renderJs()), so it
+                // already has its own, finer-grained invalidation and is left
+                // using the global cache key set in Assets::render().
+                if ($this->timestamp && $this->modified) {
+                    $this->timestamp = dechex($this->modified);
+                }
             }
         }
 
--- a/tests/unit/Grav/Common/AssetsTest.php
+++ b/tests/unit/Grav/Common/AssetsTest.php
@@ -714,67 +714,143 @@ class AssetsTest extends \PHPUnit\Framework\TestCase
     {
         $fileA = GRAV_ROOT . '/tests/unit/data/assets/timestamp-a.css';
         $fileB = GRAV_ROOT . '/tests/unit/data/assets/timestamp-b.css';
+        $fileJs = GRAV_ROOT . '/tests/unit/data/assets/timestamp-a.js';
 
-        // Distinct, known mtimes for each fixture (not "now") so the assertions
-        // don't depend on filesystem timing/precision.
-        $mtimeA = time() - 3600;
-        $mtimeB = time() - 60;
-        touch($fileA, $mtimeA);
-        touch($fileB, $mtimeB);
-        clearstatcache(true, $fileA);
-        clearstatcache(true, $fileB);
+        $origMtimeA = filemtime($fileA);
+        $origMtimeB = filemtime($fileB);
+        $origMtimeJs = filemtime($fileJs);
 
-        // Local, non-pipelined assets: each gets its own filemtime-derived
-        // token, not the shared global cache key ('foo' here, via setTimestamp()
-        // which stands in for the value enable_asset_timestamp would compute).
+        try {
+            // Distinct, known mtimes for each fixture (not "now") so the assertions
+            // don't depend on filesystem timing/precision.
+            $mtimeA = time() - 3600;
+            $mtimeB = time() - 60;
+            touch($fileA, $mtimeA);
+            touch($fileB, $mtimeB);
+            clearstatcache(true, $fileA);
+            clearstatcache(true, $fileB);
+
+            // Exercise the real enable_asset_timestamp config flag (the actual
+            // trigger for #4049), not just a manually set token.
+            $this->assets->reset();
+            $this->grav['config']->set('system.assets.enable_asset_timestamp', true);
+            $this->assets->config(['enable_asset_timestamp' => true]);
+            // Pin the resulting global key so remote/pipeline assertions below
+            // (which are unaffected by the per-file override) are deterministic.
+            $this->assets->setTimestamp('foo');
+
+            // Local, non-pipelined assets: each gets its own filemtime-derived
+            // token, not the shared global cache key.
+            $this->assets->addCss('/tests/unit/data/assets/timestamp-a.css');
+            $css = $this->assets->css();
+            self::assertSame(
+                '<link href="/tests/unit/data/assets/timestamp-a.css?' . dechex($mtimeA) . '" type="text/css" rel="stylesheet">' . PHP_EOL,
+                $css
+            );
+
+            $this->assets->reset();
+            $this->grav['config']->set('system.assets.enable_asset_timestamp', true);
+            $this->assets->config(['enable_asset_timestamp' => true]);
+            $this->assets->setTimestamp('foo');
+            $this->assets->addCss('/tests/unit/data/assets/timestamp-b.css');
+            $css = $this->assets->css();
+            self::assertSame(
+                '<link href="/tests/unit/data/assets/timestamp-b.css?' . dechex($mtimeB) . '" type="text/css" rel="stylesheet">' . PHP_EOL,
+                $css
+            );
+
+            // Touching one file changes only its own token.
+            $mtimeANew = $mtimeA + 1800;
+            touch($fileA, $mtimeANew);
+            clearstatcache(true, $fileA);
+
+            $this->assets->reset();
+            $this->grav['config']->set('system.assets.enable_asset_timestamp', true);
+            $this->assets->config(['enable_asset_timestamp' => true]);
+            $this->assets->setTimestamp('foo');
+            $this->assets->addCss('/tests/unit/data/assets/timestamp-a.css');
+            $css = $this->assets->css();
+            self::assertSame(
+                '<link href="/tests/unit/data/assets/timestamp-a.css?' . dechex($mtimeANew) . '" type="text/css" rel="stylesheet">' . PHP_EOL,
+                $css
+            );
+
+            // Stream-based local asset (e.g. theme://): goes through
+            // $locator->findResource() rather than GRAV_WEBROOT concatenation,
+            // and must be rewritten to its own mtime too.
+            $this->assets->reset();
+            $this->grav['config']->set('system.assets.enable_asset_timestamp', true);
+            $this->assets->config(['enable_asset_timestamp' => true]);
+            $this->assets->setTimestamp('foo');
+            $this->assets->addCss('tests://unit/data/assets/timestamp-a.css');
+            $css = $this->assets->css();
+            self::assertSame(
+                '<link href="/tests/unit/data/assets/timestamp-a.css?' . dechex($mtimeANew) . '" type="text/css" rel="stylesheet">' . PHP_EOL,
+                $css
+            );
+
+            // Local JS asset: addJs() goes through a different concrete asset
+            // class than addCss(), but must be rewritten the same way.
+            $mtimeJs = time() - 120;
+            touch($fileJs, $mtimeJs);
+            clearstatcache(true, $fileJs);
+
+            $this->assets->reset();
+            $this->grav['config']->set('system.assets.enable_asset_timestamp', true);
+            $this->assets->config(['enable_asset_timestamp' => true]);
+            $this->assets->setTimestamp('foo');
+            $this->assets->addJs('/tests/unit/data/assets/timestamp-a.js');
+            $js = $this->assets->js();
+            self::assertSame(
+                '<script src="/tests/unit/data/assets/timestamp-a.js?' . dechex($mtimeJs) . '"></script>' . PHP_EOL,
+                $js
+            );
+
+            // Remote assets have no local file to stat, so they keep the global
+            // cache key untouched.
+            $this->assets->reset();
+            $this->grav['config']->set('system.assets.enable_asset_timestamp', true);
+            $this->assets->config(['enable_asset_timestamp' => true]);
+            $this->assets->setTimestamp('foo');
+            $this->assets->addCss('http://somesite.com/test.css');
+            $css = $this->assets->css();
+            self::assertSame('<link href="http://somesite.com/test.css?foo" type="text/css" rel="stylesheet">' . PHP_EOL, $css);
+
+            // Pipelined assets: the *rendered* query string keeps using the
+            // global cache key (Pipeline::$timestamp, set from Assets::render()),
+            // regardless of the per-asset mtime rewrite above.
+            $this->assets->reset();
+            $this->grav['config']->set('system.assets.enable_asset_timestamp', true);
+            $this->assets->config(['enable_asset_timestamp' => true]);
+            $this->assets->setTimestamp('foo');
+            $this->assets->setCssPipeline(true);
+            $this->assets->addCss('/tests/unit/data/assets/timestamp-a.css');
+            $css = $this->assets->css();
+            self::assertMatchesRegularExpression('#<link href="/assets/(.*)\.css\?foo" type="text/css" rel="stylesheet">#', $css);
+        } finally {
+            touch($fileA, $origMtimeA);
+            touch($fileB, $origMtimeB);
+            touch($fileJs, $origMtimeJs);
+            clearstatcache(true, $fileA);
+            clearstatcache(true, $fileB);
+            clearstatcache(true, $fileJs);
+        }
+    }
+
+    public function testManualTimestampPreservedWithoutConfigFlag(): void
+    {
+        // enable_asset_timestamp defaults to off. A caller-supplied token set
+        // via the public Assets::setTimestamp() API must not be silently
+        // overridden by the per-file mtime rewrite, since that rewrite is only
+        // meant to fire for the enable_asset_timestamp config path.
         $this->assets->reset();
-        $this->assets->setTimestamp('foo');
+        $this->assets->setTimestamp('v1.2.3');
         $this->assets->addCss('/tests/unit/data/assets/timestamp-a.css');
         $css = $this->assets->css();
         self::assertSame(
-            '<link href="/tests/unit/data/assets/timestamp-a.css?' . dechex($mtimeA) . '" type="text/css" rel="stylesheet">' . PHP_EOL,
+            '<link href="/tests/unit/data/assets/timestamp-a.css?v1.2.3" type="text/css" rel="stylesheet">' . PHP_EOL,
             $css
         );
-
-        $this->assets->reset();
-        $this->assets->setTimestamp('foo');
-        $this->assets->addCss('/tests/unit/data/assets/timestamp-b.css');
-        $css = $this->assets->css();
-        self::assertSame(
-            '<link href="/tests/unit/data/assets/timestamp-b.css?' . dechex($mtimeB) . '" type="text/css" rel="stylesheet">' . PHP_EOL,
-            $css
-        );
-
-        // Touching one file changes only its own token.
-        $mtimeANew = $mtimeA + 1800;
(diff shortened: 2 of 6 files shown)


system/src/Grav/Common/Assets/BaseAsset.php
```diff
@@ -144,6 +144,21 @@ public function init($asset, $options)
...
 
                 $this->modified = $file->isFile() ? $file->getMTime() : false;
+
+                // Rewrite the cache-busting token to this asset's own filemtime
+                // instead of the global config/version-derived cache key, so
+                // editing one file invalidates only that file's URL.
+                //
+                // Scope: this only affects the query string built for a single,
+                // individually rendered asset (this class's own renderQueryString()).
+                // Pipelined/bundled assets never consult this per-asset timestamp:
```
wakqasahmed: **This comment is factually incorrect — the pipeline *does* consult this per-asset timestamp.**

`Pipeline::renderCss()` computes `$uid = md5(json_encode($assets) . ...)`, and `$assets` is the array of `BaseAsset` objects. `BaseAsset::jsonSerialize()` returns `["type" => ..., "elements" => $this->getElements()]`, and `ObjectPropertyTrait::getElements()` returns *all* defined properties — including `timestamp`. So mutating `$this->timestamp` here feeds straight into the bundle filename hash.

Verified empirically on this branch (`php:8.3-cli`, same fixture, same fixed mtime `1700000000`, pipelined CSS):

```
with this patch:    /assets/1a6ee842880d3bf127496d686d2c9a9b.css?foo
patch line removed: /assets/865167985240bb41f9533c00967fbd12.css?foo
```

Consequences (neither is a correctness bug, but the reasoning in the comment and in the PR body should be corrected):

1. Every existing pipelined bundle filename changes once on upgrade — new files written into `assets://`, the old ones orphaned until a cache clear.
2. The stated rationale is also imprecise: the uid is a hash of the *serialized asset objects* (which already include `modified`, the mtime), not of the bundled *content*. I confirmed that without this patch the uid already changes when a fixture's mtime changes (`865167...` → `3133c6...`), which is why the pipeline path was never affected by #4049 in the first place. Worth saying that plainly rather than "hash of ... content".

Suggested rewrite: the *rendered query string* of a pipelined bundle comes from `Pipeline::$timestamp` (set from the global key in `Assets::render()`), so pipelined output is not switched to per-file mtimes; the per-asset timestamp does still perturb the bundle uid hash, harmlessly.

system/src/Grav/Common/Assets/BaseAsset.php
```diff
@@ -144,6 +144,21 @@ public function init($asset, $options)
...
+                // Pipeline builds its own uid from a hash of the bundled asset
+                // list and content (see Pipeline::renderCss()/renderJs()), so it
+                // already has its own, finer-grained invalidation and is left
+                // using the global cache key set in Assets::render().
+                if ($this->timestamp && $this->modified) {
```
wakqasahmed: **Main functional concern: this triggers on `$this->timestamp` being truthy, not on `enable_asset_timestamp` specifically — so it silently overrides an explicitly set token.**

`Assets::setTimestamp()` is public API (`TestingAssetsTrait:227`, with a matching `getTimestamp()`), and themes/plugins use it to pin a manual cache-bust token, e.g. `$assets->setTimestamp("v1.2.3")` or a build/release hash. With this change, any such explicit value is now discarded for every local asset and replaced by an mtime, with no way to opt out. That is a behavioural regression for a supported public API.

Consider gating on the config flag rather than on the token being non-empty — i.e. only rewrite when the token actually came from `enable_asset_timestamp` (`Grav::instance()["cache"]->getKey()`), leaving a caller-supplied token alone. Passing an explicit flag through `$options` from `Assets::addType()` (where `$options["timestamp"] = $this->timestamp` is assigned) would be the least invasive way.

Two smaller notes on the same block:

- **No extra `stat()` — worth correcting the PR description.** The body says the change calls `filemtime()`; it does not. `$this->modified` is already populated on the line above by `SplFileInfo::getMTime()`, which existed before this PR, so the added code is a pure reuse of an already-stat'd value. There is genuinely zero added syscall cost even on pages with hundreds of assets. Good — but the PR text as written invites a perf objection that does not apply.
- **Silent no-op when the file is not found.** For non-stream assets `$path = GRAV_WEBROOT . $asset` is never `false`, so a missing file reaches `SplFileInfo` with `isFile() === false`, `$this->modified === false`, and the asset keeps the global key. That is why the pre-existing `testTimestamps` (which uses a nonexistent `/test.css`) still passes. The fallback is sensible, but it does mean the feature silently degrades to the old broken behaviour on any setup where `GRAV_WEBROOT` + the asset path does not resolve to the real file (symlinked/aliased asset roots, CDN-rewritten paths). Worth a sentence in the docs/blueprint help so users can tell "not working" from "not applicable".

tests/unit/Grav/Common/AssetsTest.php
```diff
@@ -710,6 +710,73 @@ public function testTimestamps(): void
...
+    public function testPerFileTimestamps(): void
+    {
+        $fileA = GRAV_ROOT . '/tests/unit/data/assets/timestamp-a.css';
+        $fileB = GRAV_ROOT . '/tests/unit/data/assets/timestamp-b.css';
+
+        // Distinct, known mtimes for each fixture (not "now") so the assertions
+        // don't depend on filesystem timing/precision.
+        $mtimeA = time() - 3600;
+        $mtimeB = time() - 60;
+        touch($fileA, $mtimeA);
```
wakqasahmed: **The test is genuinely meaningful — I verified it is not vacuous.** Reverting just the three-line change in `BaseAsset::init()` and re-running gives `Tests: 25, Assertions: 140, Failures: 1` with `AssetsTest: Per file timestamps` failing on the exact-string assertion. It really does touch a real file's mtime and assert the rendered query string tracks it. 

Four gaps I would still like closed:

1. **`enable_asset_timestamp` itself is never exercised.** The test uses `setTimestamp("foo")` as a stand-in. Given the issue is specifically about the config flag's behaviour, and given the override currently keys off "token is truthy" rather than "token came from the flag" (see my comment on `BaseAsset.php:159`), a case that actually sets `system.assets.enable_asset_timestamp` would be the regression test that matters.
2. **No stream-based asset.** The common real-world case is `theme://js/script.js`, which takes the `$locator->findResource()` branch rather than `GRAV_WEBROOT . $asset`. That branch is untested here and is the one users in #4049 hit.
3. **No JS coverage.** `addJs()` goes through a different concrete asset class; a one-liner would cover it.
4. **`touch()` mutates repo-tracked fixture files and never restores them.** `timestamp-a.css` is left at `$mtimeA + 1800` after the run. Harmless today, but it makes the test order-sensitive against anything that hashes those files (e.g. the pipeline uid, which as noted elsewhere *does* incorporate mtime), and it dirties the working tree's stat cache for whoever runs the suite locally. Prefer creating the fixtures in a temp dir in `setUp()`/`_before()` and removing them after, or at least restoring the original mtimes in a `finally`.

Also: the pipeline assertion at the end is a regex that would still pass if the pipeline silently changed its uid derivation, which is fine as a smoke test but does not actually prove "pipelined assets keep the global cache key" as strongly as the comment above it claims.

wakqasahmed (commented):
Independent review (cold-start, no context from the implementation). I checked out the branch, read `BaseAsset.php`, `Assets.php`, `Pipeline.php` and `AssetUtilsTrait::renderQueryString()` myself, and ran `AssetsTest` under `php:8.3-cli` both with and without the patch rather than taking the PR body at its word.

**Verified as claimed**

- **Scoping is correct.** The rewrite sits inside `if (!Utils::startsWith($this->getType(), "inline"))` → `if (!$this->remote)`, after path resolution, and is guarded by `$this->timestamp && $this->modified`. Remote URLs never reach it and keep the global cache key; inline assets are untouched; a stream that fails `findResource()` returns `false` before it. Confirmed by reading the code, and the pre-existing `testTimestamps` (which asserts `http://somesite.com/test.css?foo`) still passes.
- **No performance concern at all** — and for a better reason than the PR gives. The PR body says it calls `filemtime()`; it does not. `$this->modified` is already populated one line above by `SplFileInfo::getMTime()`, which predates this PR. The added code reuses that value, so there is **zero** additional `stat()` per asset, even on pages with hundreds of them. The PR description should be corrected so it stops inviting a perf objection that does not apply.
- **The new test is not vacuous.** Reverting only the three added lines in `BaseAsset::init()` turns the suite red (`Tests: 25, Assertions: 140, Failures: 1`, failing on `Per file timestamps`). It touches real mtimes and asserts exact rendered query strings. Full `AssetsTest` on the branch: `OK (25 tests, 145 assertions)`.
- **Pipelined output is not switched to per-file mtimes.** The bundle `<link>`/`<script>` query string comes from `Pipeline::$timestamp`, seeded from the global key in `Assets::render()` — a separate property. That part of the claim holds.

**Findings**

1. *(highest value)* **`Assets::setTimestamp()` is silently overridden.** The rewrite triggers on `$this->timestamp` being truthy, not on `enable_asset_timestamp`. A theme or plugin that pins a manual release token via the public `setTimestamp("v1.2.3")` API now gets mtimes instead, with no opt-out. Details and a suggested fix inline on `BaseAsset.php:159`.
2. **The inline comment (and the PR body) is factually wrong about the pipeline.** "Pipelined/bundled assets never consult this per-asset timestamp" is not true: `Pipeline::renderCss()` hashes `json_encode($assets)`, and `BaseAsset::jsonSerialize()` → `getElements()` includes `timestamp`. I measured the bundle filename changing from `865167985240bb41f9533c00967fbd12` to `1a6ee842880d3bf127496d686d2c9a9b` for an identical file and mtime purely as a result of this patch. Harmless (one-time bundle regeneration on upgrade, orphaned files in `assets://`), but the comment must be corrected — it is the kind of comment a future maintainer will trust. Detail inline on `BaseAsset.php:154`.
3. **Design trade-off that the PR should acknowledge: mtime is not deploy-stable.** `git checkout`, `rsync` without `-t`, and container image rebuilds all reset mtimes on files whose *content* never changed. Combined with the `Cache-Control: immutable, max-age=31536000` setup the issue describes, this turns "assets never re-download" into "every asset re-downloads on every deploy". Worse, multi-node deployments will produce *different* mtimes per node, so the same asset gets different URLs depending on which server answered — cache fragmentation, and with hashed-bundle CDNs, potential misses on every request. The issue reporter explicitly preferred a content hash (`md5`/`xxh3`) for exactly this reason. mtime is a defensible pragmatic choice (it is free here, as noted), but the PR should say why it was chosen over a content hash, and the blueprint help text should warn that deploys must preserve mtimes for the caching benefit to hold.
4. **Silent no-op when the file is not resolvable.** Non-stream missing files keep the global key rather than erroring, which is why the old test still passes. Sensible, but it means the feature quietly degrades to the old #4049 behaviour on symlinked/aliased asset roots. Worth one line of documentation.
5. **`system/blueprints/config/system.yaml:1037`** — `assets.enable_asset_timestamp` help text is unchanged. Part of #4049 was that the option name/description misrepresents the behaviour; now that the behaviour is genuinely per-file for non-pipelined local assets, the description should be updated to match (and to state the pipelined and remote exceptions).
6. **Test gaps** — no coverage of the actual `enable_asset_timestamp` config path, no stream (`theme://`) asset, no JS asset, and `touch()` mutates repo-tracked fixtures without restoring them. Details inline on `AssetsTest.php:722`.

None of the above is a correctness bug in the rendered output for the case the issue describes — the core three-line change does what it says. Items 1 and 2 are the ones I would want addressed before merge; 3–6 are documentation/robustness.

system/src/Grav/Common/Assets/BaseAsset.php
wakqasahmed: Fixed — you're right, the comment was wrong. Rewrote it to state the actual behavior: the *rendered* pipelined tag's query string still uses the global cache key (`Pipeline::$timestamp`, unaffected), but the pipeline's internal bundle uid hash does incorporate this per-asset `$timestamp` via `jsonSerialize()`/`getElements()`. Harmless as you noted (one-time bundle regeneration + orphaned old file on upgrade), but the comment now says so instead of claiming isolation. Also corrected the PR description's equivalent claim. Pushed in bba3125.

system/src/Grav/Common/Assets/BaseAsset.php
wakqasahmed: Fixed — this was the real blocking issue. `BaseAsset::init()` now gates the per-file mtime rewrite on `Grav::instance()['config']->get('system.assets.enable_asset_timestamp')` directly, in addition to the existing `$this->timestamp && $this->modified` truthiness check, instead of firing on token truthiness alone. So a token set via the public `Assets::setTimestamp()` API is left untouched whenever the config flag is off (its default), and only gets rewritten when the flag is genuinely on. Added `AssetsTest::testManualTimestampPreservedWithoutConfigFlag()` to prove it — asserts `setTimestamp('v1.2.3')` survives verbatim with the flag off. Pushed in bba3125.

tests/unit/Grav/Common/AssetsTest.php
wakqasahmed: Addressed all four:

1. `testPerFileTimestamps()` now enables `system.assets.enable_asset_timestamp` for real (via `$grav['config']->set(...)` + `$assets->config(...)`) instead of standing it in with `setTimestamp('foo')`, and a separate new test (`testManualTimestampPreservedWithoutConfigFlag`) exercises the config-off / manual-token case.
2. Added a `tests://unit/data/assets/timestamp-a.css` case, going through the `$locator->findResource()` / `isStream()` branch.
3. Added a local JS fixture/case through `addJs()`.
4. Wrapped the `touch()` mutations in try/finally and restore each fixture's original mtime (`filemtime()` captured up front) after the test runs, regardless of pass/fail.

Also tightened the pipeline assertion's surrounding comment to say what it actually proves (rendered query string only), not more than that. Pushed in bba3125; full `AssetsTest` is 26/26 (148 assertions), broader unit suite unchanged from baseline (973/3060, 17 errors/1 failure, all pre-existing per `ZipArchive`/`ComposerTest`).

rhukster:
Merged, thanks — and real credit for going back and correcting your own description about the pipeline. That's noticed, and it made the review much faster.

You were being harder on yourself than necessary there. `modified` — the raw int mtime — was *already* in the serialized elements before this PR (the repo's own expectation at `tests/unit/Grav/Common/AssetsTest.php:68` shows `"modified":false` next to `timestamp`), so the bundle hash already moved whenever a source file's mtime moved. Rendering the pipeline twice around a `touch()` on both branches confirmed it: with the flag off the uids are byte-identical to develop, and with the flag on the churn rate is unchanged and only the absolute values shift once. And since `Install.php` runs `Cache::clearCache('all')` on every core upgrade with `asset://` in that list, the one orphaned bundle is swept by the very upgrade that creates it.

Worth stating for the record too, because it's the strongest point in the PR and the description hides it: this adds **zero** stat calls. `init()` already stats the file, and the patch reuses `$this->modified` rather than calling `filemtime()` again. The title and body both say "dechex(filemtime())", which sends a reviewer straight to a per-request-syscall objection that doesn't apply.

One real change on the way in: the `setTimestamp()` precedence was backwards. With the flag on *and* an explicit `Assets::setTimestamp()`, the caller's token was silently discarded — and the test locked that in. Pinning one release token across a fleet is exactly what that API is for, and there was no way to opt back out once the flag was on. The rewrite now only ever replaces the token the flag generated for itself, so an explicit call wins either way. I flipped that test and added one covering the flag-on case. Also shortened the `system.yaml` comment to fit the file's aligned comment column, and added the CHANGELOG entry.

Shipping in 2.0.20.

PR #4244, opened 2026-08-16
Fix markdown image embed failing on filenames with a colon (#3933)

### LINKED ISSUE #3933: Media embed in markdown fails when image filename has ":" character
I originally filed this on getgrav/grav-plugin-admin because the link in the admin interface, when it seems probably more likely to be an issue with grav.

https://github.com/getgrav/grav-plugin-admin/issues/2454

wakqasahmed:
## What changed

`Excerpts::parseUrl()` uses `Utils::multibyteParseUrl()` (a wrapper around `parse_url()`). Unlike RFC 3986, PHP's `parse_url()` allows a "scheme" to start with a digit, so a relative filename that merely contains a literal `:` — e.g. a timestamp like `2025-06-29T13:36:56.png` — gets misread as a `scheme:path` split (`scheme = "2025-06-29t13"`). Excerpts then treats the reference as an external/stream URL instead of resolving it against the page's own media, so the image fails to render.

The fix rebuilds the original path whenever the parsed `scheme`:
- has no `://` in the source string, **and**
- fails RFC 3986 scheme grammar (doesn't start with a letter), **and**
- isn't a name registered with Grav's `UniformResourceLocator` (`schemeExists()`)

Genuine schemes (`mailto:`, `tel:`, `xmpp:`, any other real protocol — these always start with a letter) and registered Grav streams (`image://`, `user://`, ...) are untouched, since they either contain `://` or start with a letter.

Closes #3933

## Test plan

Added to `tests/unit/Grav/Common/Helpers/ExcerptsTest.php` and ran via `codecept run unit` (PHP 8.3, gd extension):

- **Bug case**: `testImageFilenameWithColonIsResolvedAsLocalMedia` — an `` referencing the colon-bearing fixture `tests/fake/nested-site/user/pages/02.item2/02.item2-2/2025-06-29T13:36:56.png` now resolves to the page's own media and renders a proper cache-derivative URL, instead of failing to resolve.
- **Non-regression — real schemes without `://`**: `testMailtoAndTelLinksAreNotBrokenByColonFix` — `mailto:`, `tel:`, and `xmpp:` links still pass through untouched (this last case caught a bug in an earlier iteration of the fix that used a hardcoded external-scheme allowlist instead of RFC scheme grammar, which broke `xmpp:` — an existing `ParsedownTest::testSpecialProtocols` regression, now fixed and passing).
- **Non-regression — registered streams**: `testStreamImageIsNotBrokenByColonFix` — `image://watermark.png` still resolves through the locator exactly as before.

Full run: `php vendor/bin/codecept run unit` — 974 tests, 3108 assertions. All pass except 8 pre-existing errors (`ZipArchive` class not available in the bare `php:8.3-cli` test image — unrelated to this change) and 1 pre-existing failure (`ComposerTest::testGetComposerLocation`, an absolute-path assumption unrelated to this change) — both reproduce identically on `develop` without this patch.

### DIFF
--- a/system/src/Grav/Common/Page/Markdown/Excerpts.php
+++ b/system/src/Grav/Common/Page/Markdown/Excerpts.php
@@ -338,9 +338,24 @@ class Excerpts
         if (isset($url_parts['scheme'])) {
             /** @var UniformResourceLocator $locator */
             $locator = Grav::instance()['locator'];
+            $is_registered_stream = $locator->schemeExists($url_parts['scheme']);
 
-            // Special handling for the streams.
-            if ($locator->schemeExists($url_parts['scheme'])) {
+            if (
+                !$is_registered_stream
+                && strpos($url, '://') === false
+                && !preg_match('/^[a-zA-Z][a-zA-Z0-9+.-]*$/', $url_parts['scheme'])
+            ) {
+                // parse_url() misreads a relative filename that merely contains a
+                // literal ':' (e.g. "2025-06-29T13:36:56.png") as a scheme:path
+                // split, because unlike RFC 3986 it allows a "scheme" to start
+                // with a digit. A real scheme never does (mailto:, tel:, xmpp:,
+                // ...), so only a scheme failing that grammar - and matching no
+                // registered stream - is untrusted and rebuilt as a plain path
+                // (getgrav/grav#3933).
+                $url_parts['path'] = $url_parts['scheme'] . ':' . ($url_parts['host'] ?? '') . ($url_parts['path'] ?? '');
+                unset($url_parts['scheme'], $url_parts['host'], $url_parts['port'], $url_parts['user'], $url_parts['pass']);
+            } elseif ($is_registered_stream) {
+                // Special handling for the streams.
                 if (isset($url_parts['host'])) {
                     // Merge host and path into a path.
                     $url_parts['path'] = $url_parts['host'] . (isset($url_parts['path']) ? '/' . $url_parts['path'] : '');
--- a/tests/unit/Grav/Common/Helpers/ExcerptsTest.php
+++ b/tests/unit/Grav/Common/Helpers/ExcerptsTest.php
@@ -7,6 +7,7 @@ use Grav\Common\Page\Interfaces\PageInterface;
 use Grav\Common\Uri;
 use Grav\Common\Config\Config;
 use Grav\Common\Page\Pages;
+use Grav\Common\Page\Media;
 use Grav\Common\Language\Language;
 use RocketTheme\Toolbox\ResourceLocator\UniformResourceLocator;
 
@@ -116,6 +117,66 @@ class ExcerptsTest extends \PHPUnit\Framework\TestCase
         self::assertStringStartsWith(
             '<a href="https://meet.weikamp.biz/Support" rel="nofollow" target="_blank"',
             Excerpts::processLinkHtml('<a href="https://meet.weikamp.biz/Support?rel=nofollow&target=_blank">target and rel</a>')
+        );
+    }
+
+    public function testImageFilenameWithColonIsResolvedAsLocalMedia(): void
+    {
+        // getgrav/grav#3933: a bare relative filename containing a literal ':'
+        // (e.g. a timestamp) must still resolve to the page's own media, not be
+        // misread by parse_url() as a scheme:path split.
+        //
+        // The fixture is generated here at runtime rather than committed to
+        // git: ':' is a reserved character on NTFS, so a statically committed
+        // file with this name would make `git clone`/checkout of the whole
+        // repo fail on Windows. Page media loads lazily on first access, so
+        // it's enough for the file to exist before processImageHtml() runs.
+        $fixturePath = 'tests/fake/nested-site/user/pages/02.item2/02.item2-2/2025-06-29T13:36:56.png';
+        $image = imagecreatetruecolor(10, 10);
+        imagepng($image, $fixturePath);
+        imagedestroy($image);
+
+        // The page's media collection was already built (by Pages::init())
+        // before this fixture existed on disk, so force a fresh scan of the
+        // folder now that the file is there.
+        $this->page->media(new Media($this->page->getMediaFolder(), $this->page->getMediaOrder()));
+
+        try {
+            self::assertMatchesRegularExpression(
+                '|<img alt="Timestamped" src="\/images\/.*-2025-06-29t133656\.png" data-src="2025-06-29T13:36:56\.png\?cropZoom=300,300" \/>|',
+                Excerpts::processImageHtml('<img src="2025-06-29T13:36:56.png?cropZoom=300,300" alt="Timestamped" />', $this->page)
+            );
+        } finally {
+            @unlink($fixturePath);
+        }
+    }
+
+    public function testMailtoAndTelLinksAreNotBrokenByColonFix(): void
+    {
+        // Non-regression: any scheme without "://" that still has valid,
+        // letter-led URI scheme grammar (RFC 3986) must pass through
+        // untouched, whether or not Grav has special-case handling for it.
+        self::assertStringStartsWith(
+            '<a href="mailto:bob@example.com"',
+            Excerpts::processLinkHtml('<a href="mailto:bob@example.com">email</a>')
+        );
+        self::assertStringStartsWith(
+            '<a href="tel:+123456789"',
+            Excerpts::processLinkHtml('<a href="tel:+123456789">call</a>')
+        );
+        self::assertStringStartsWith(
+            '<a href="xmpp:xyx@domain.com"',
+            Excerpts::processLinkHtml('<a href="xmpp:xyx@domain.com">xmpp</a>')
+        );
+    }
+
+    public function testStreamImageIsNotBrokenByColonFix(): void
+    {
+        // Non-regression: registered Grav streams (image://, user://, ...) must
+        // keep resolving via the locator exactly as before.
+        self::assertStringStartsWith(
+            '<img alt="Stream" src="/system/images/watermark.png"',
+            Excerpts::processImageHtml('<img src="image://watermark.png" alt="Stream" />', $this->page)
         );
     }
 }
--- a/system/src/Grav/Common/Grav.php
+++ b/system/src/Grav/Common/Grav.php
@@ -859,6 +859,17 @@ class Grav extends Container
         $supported_types = $config->get('media.types');
 
         $parsed_url = parse_url(rawurldecode($uri->basename()));
+        if (isset($parsed_url['scheme']) && !preg_match('/^[a-zA-Z][a-zA-Z0-9+.-]*$/', $parsed_url['scheme'])) {
+            // getgrav/grav#3933: parse_url() misreads a basename that merely
+            // contains a literal ':' (e.g. a timestamp such as
+            // "2025-06-29T13:36:56.png") as a scheme:path split, because
+            // unlike RFC 3986 it allows a "scheme" to start with a digit.
+            // A real scheme never does, so rebuild the original basename as
+            // a plain path in that case (see Excerpts::parseUrl() for the
+            // same fix applied to markdown image/link resolution).
+            $parsed_url['path'] = $parsed_url['scheme'] . ':' . ($parsed_url['host'] ?? '') . ($parsed_url['path'] ?? '');
+            unset($parsed_url['scheme'], $parsed_url['host'], $parsed_url['port'], $parsed_url['user'], $parsed_url['pass']);
+        }
         $media_file = isset($parsed_url['path']) ? $parsed_url['path'] : '';
 
         $event = new Event([


system/src/Grav/Common/Page/Markdown/Excerpts.php
```diff
@@ -335,6 +335,23 @@ protected function parseUrl(string $url)
     {
         $url_parts = Utils::multibyteParseUrl($url);
 
+        if (isset($url_parts['scheme']) && strpos($url, '://') === false && !preg_match('/^[a-zA-Z][a-zA-Z0-9+.-]*$/', $url_parts['scheme'])) {
```
wakqasahmed: Independently verified the premise of this guard on `php:8.3-cli`, and it holds:

```
2025-06-29T13:36:56.png              {"scheme":"2025-06-29T13","path":"36:56.png"}
2025-06-29T13:36:56.png?cropZoom=..  {"scheme":"2025-06-29T13","path":"36:56.png","query":"cropZoom=300,300"}
mailto:bob@example.com               {"scheme":"mailto","path":"bob@example.com"}
tel:+123456789                       {"scheme":"tel","path":"+123456789"}
xmpp:xyx@domain.com                  {"scheme":"xmpp","path":"xyx@domain.com"}
image://watermark.png                {"scheme":"image","host":"watermark.png"}
```

So `parse_url()` really does accept digit-led "schemes" that RFC 3986 forbids, and the three real-scheme / stream cases are all left untouched by this condition (letter-led, or contain `://`). `data:image/png;base64,...` is also safe for the same reason. Since a digit-led scheme cannot be a valid URI scheme, the guard cannot swallow a legitimate absolute URL — the direction of the fix is sound.

One nice side effect worth noting in the comment block: the regex runs on the *urldecoded* scheme that `Utils::multibyteParseUrl()` returns, so a filename like `Screenshot at 13:36:56.png` (scheme decodes to `Screenshot at 13`, fails the grammar on the spaces) is also fixed, not just digit-led ones. I confirmed that case rendering correctly with the patch and failing without it.

Remaining (pre-existing, not a regression): a letter-led colon filename with no spaces, e.g. `photo:2025.png`, is still misread — verified identical output with and without the patch. That one is genuinely ambiguous with a real scheme, so leaving it is defensible; a sentence in the comment saying so would save the next reader from re-deriving it.

system/src/Grav/Common/Page/Markdown/Excerpts.php
```diff
@@ -335,6 +335,23 @@ protected function parseUrl(string $url)
...
+            if (!$locator->schemeExists($url_parts['scheme'])) {
+                // parse_url() misreads a relative filename that merely contains a
+                // literal ':' (e.g. "2025-06-29T13:36:56.png") as a scheme:path
+                // split, because unlike RFC 3986 it allows a "scheme" to start
+                // with a digit. A real scheme never does (mailto:, tel:, xmpp:,
+                // ...), so only a scheme failing that grammar - and matching no
+                // registered stream - is untrusted and rebuilt as a plain path
+                // (getgrav/grav#3933).
+                $url_parts['path'] = $url_parts['scheme'] . ':' . ($url_parts['host'] ?? '') . ($url_parts['path'] ?? '');
+                unset($url_parts['scheme'], $url_parts['host'], $url_parts['port'], $url_parts['user'], $url_parts['pass']);
```
wakqasahmed, line 356: Minor, non-blocking: the `host` / `port` / `user` / `pass` handling here is unreachable in practice.

I fuzzed 40k random short strings through `parse_url()` on PHP 8.3 looking for any input that (a) has no `://`, (b) yields a `scheme`, (c) whose scheme fails `^[a-zA-Z][a-zA-Z0-9+.-]*$`, and (d) also sets any of `host`/`port`/`user`/`pass`. Zero hits — PHP only populates the authority components when it saw `//`, which this branch has already excluded. (`a:8080/x.png` does produce `host`+`port`, but with **no** `scheme`, so `isset($url_parts[scheme])` keeps it out of the branch.)

So `($url_parts[host] ?? )` always concatenates the empty string and four of the five `unset()` targets are always absent. Harmless as defensive code — just flagging it so nobody later reads it as evidence that a `scheme` + `host` + no-`://` shape exists.

Also a small readability nit: this creates two adjacent `if (isset($url_parts[scheme]))` blocks that each fetch `Grav::instance()[locator]` and each call `schemeExists()` for the same scheme. Hoisting `$locator` above both (or folding into one `if (isset($scheme))` block with the stream handling in the `else`) would avoid the duplicate lookup.

tests/unit/Grav/Common/Helpers/ExcerptsTest.php
```diff
@@ -118,4 +118,44 @@ public function testTarget(): void
...
     }
+
+    public function testImageFilenameWithColonIsResolvedAsLocalMedia(): void
+    {
+        // getgrav/grav#3933: a bare relative filename containing a literal ':'
+        // (e.g. a timestamp) must still resolve to the page's own media, not be
+        // misread by parse_url() as a scheme:path split.
+        self::assertMatchesRegularExpression(
+            '|<img alt="Timestamped" src="\/images\/.*-2025-06-29t133656\.png" data-src="2025-06-29T13:36:56\.png\?cropZoom=300,300" \/>|',
+            Excerpts::processImageHtml('<img src="2025-06-29T13:36:56.png?cropZoom=300,300" alt="Timestamped" />', $this->page)
```
wakqasahmed: Confirmed this is a real red/green regression test, not a tautology. Reverting only `Excerpts.php` and re-running gives exactly the symptom from getgrav/grav#3933:

```
1) ExcerptsTest: Image filename with colon is resolved as local media
Failed asserting that
  <img src="2025-06-29T13:36:56.png?cropZoom=300,300" alt="Timestamped" data-src="..."></img>
matches PCRE pattern "|<img alt="Timestamped" src="/images/.*-2025-06-29t133656\.png" ...|"
```

i.e. the untouched `src` the issue reports. With the patch: `OK (6 tests, 12 assertions)`. Full `tests/unit/Grav/Common/` run on PHP 8.3 with gd: `OK (949 tests, 2960 assertions)` — no regressions in `UtilsTest`, `UriTest` or `ParsedownTest`.

Two notes on the fixture rather than the assertion, see my separate comment on the `.png` file — the binary itself is a genuine PNG (`PNG image data, 1024 x 768, 8-bit/color RGB`), and it has to be, since `cropZoom` here makes GD actually decode it. Good choice to exercise the derivative path rather than the raw-path branch.

Small nit: `testMailtoAndTelLinksAreNotBrokenByColonFix` partly duplicates the existing `ParsedownTest::testSpecialProtocols` coverage. Worth keeping anyway since it pins the behaviour at the `Excerpts` layer where the change lives, but a `@see` pointing at the Parsedown test would make the relationship explicit.

tests/fake/nested-site/user/pages/02.item2/02.item2-2/2025-06-29T13:36:56.png
wakqasahmed, line 1: **Main concern with this PR — the checked-in fixture filename, not the fix itself.**

`:` is a reserved character on NTFS/Windows. Git refuses to write such a path, so once this lands on `develop`, `git clone`/`git checkout` of grav on Windows fails with `error: invalid path tests/fake/nested-site/user/pages/02.item2/02.item2-2/2025-06-29T13:36:56.png` and leaves the working tree incomplete. Same applies to `git archive` extraction on Windows; there is no `.gitattributes` in the repo, so `tests/` is not `export-ignore`d either.

CI will not catch this — `.github/workflows/tests.yaml` runs `os: [ubuntu-latest]` only across the 8.3/8.4/8.5 matrix. It would land silently and only surface as a broken clone for Windows contributors.

Suggested fix that keeps the exact same test value: generate the fixture at runtime instead of committing it. In `setUp()`, copy the existing `sample-image.jpg`-style fixture (or `imagepng()` a tiny image) into the page folder under the colon name, delete it in `tearDown()`, and `markTestSkipped()` on `DIRECTORY_SEPARATOR === \\`. The bug is in `parse_url()` string handling, so the file only has to exist for `Media` to pick it up and for GD to decode under `cropZoom` — it does not have to be a tracked artifact.

(For the record, the committed bytes themselves are fine: `file` reports `PNG image data, 1024 x 768, 8-bit/color RGB, non-interlaced`, and I scanned every `*.png` under `tests/` — none are mislabelled JPEG. So whatever fixture-format problem existed mid-development is resolved at this head.)

wakqasahmed (commented):
Cold-start review — no prior context on this change, everything below re-derived and re-run locally (PHP 8.3 + gd, `vendor/bin/codecept`).

## Verdict

The fix is correct and well-targeted, and the test is a genuine red/green regression test. One blocking-ish portability problem with the **test fixture filename** (not the fix), plus two minor notes. Details are in the inline comments.

## What I verified independently

**1. The `parse_url()` premise is real.** Ran the cases directly rather than taking the PR description on trust:

| input | `parse_url()` |
|---|---|
| `2025-06-29T13:36:56.png` | `scheme=2025-06-29T13`, `path=36:56.png` |
| `mailto:bob@example.com` | `scheme=mailto`, `path=bob@example.com` |
| `tel:+123456789` | `scheme=tel` |
| `xmpp:xyx@domain.com` | `scheme=xmpp` |
| `image://watermark.png` | `scheme=image`, `host=watermark.png` |

PHP does accept digit-led schemes that RFC 3986 forbids. Since no valid absolute URL can have a digit-led scheme, the new guard cannot swallow a legitimate one — `mailto:`/`tel:`/`xmpp:`/`data:` are letter-led, streams carry `://` and are additionally protected by `schemeExists()`. Replacing the earlier hardcoded external-scheme allowlist with the RFC grammar was the right call; the allowlist approach was what broke `xmpp:`.

**2. The distinction holds where it matters.** Registered streams still resolve (`image://watermark.png` → `/system/images/watermark.png`). A pleasant extra: because the regex is applied to the *urldecoded* scheme from `Utils::multibyteParseUrl()`, filenames with spaces *and* colons (`Screenshot at 13:36:56.png`) are fixed too — I confirmed that case passing with the patch and failing without it. Still unfixed (pre-existing, unchanged by this PR): a letter-led no-space colon filename such as `photo:2025.png`, which is genuinely indistinguishable from a real scheme.

**3. The test exercises the real path and actually fails without the fix.** Reverting only `Excerpts.php` and re-running reproduces the issues exact symptom (untouched `src`, no `/images/` derivative). With the patch, `OK (6 tests, 12 assertions)`; full `tests/unit/Grav/Common/` → `OK (949 tests, 2960 assertions)`, no regressions in `UtilsTest`/`UriTest`/`ParsedownTest`.

**4. Fixture bytes are fine.** `PNG image data, 1024 x 768, 8-bit/color RGB` — a real PNG, which matters because `cropZoom` makes GD decode it. Scanned every `*.png` under `tests/`; none are mislabelled JPEG.

**5. Other call sites of the same scheme detection.** Checked all of them:
- `Utils::url()` (`system/src/Grav/Common/Utils.php:84`) — gates on `Utils::contains($input, ://)` before parsing, so it is structurally immune to this false positive.
- `Uri::isExternal()` (`system/src/Grav/Common/Uri.php:779`) — pure `str_starts_with` prefix matching, immune.
- `Uri::createFromString()` / `UriFactory::parseUrl()` — operate on real request URIs.
- **`Grav::fallbackUrl()` (`system/src/Grav/Common/Grav.php:861`) has the identical bug**: `parse_url(rawurldecode($uri->basename()))` on `2025-06-29T13:36:56.png` yields `path=36:56.png`, so `$media_file` is wrong and the `isset($media[$media_file])` media-object lookup at line 886 misses. Impact is limited — execution falls through to the raw `Utils::download($page->path() . / . $uri->basename())` branch, so the image is still served — but colon-named media silently skip the `Medium` path (URL media actions under `system.images.url_actions`, and the `unsupported_inline_types` inline/download decision). Same root cause, reasonably out of scope for this PR, but worth an issue or a follow-up so it does not get lost.

## Findings

- **Portability (please address):** committing a file whose name contains `:` breaks `git clone`/`git checkout` on Windows, and the Ubuntu-only CI matrix will not catch it. Suggested runtime-generated fixture in the inline comment.
- **Minor:** the `host`/`port`/`user`/`pass` handling in the reconstruction is unreachable — I fuzzed 40k inputs for a shape that reaches this branch with any authority component set and found none. Harmless, just noted.
- **Nit:** two adjacent `isset($url_parts[scheme])` blocks each fetch the locator and call `schemeExists()` for the same scheme.

No correctness or security issues found in the fix itself. Not approving — flagging the Windows-filename issue for the maintainers/author to decide on first.

tests/fake/nested-site/user/pages/02.item2/02.item2-2/2025-06-29T13:36:56.png
wakqasahmed: Fixed — the colon-named fixture is no longer committed to git at all. `tests/fake/nested-site/user/pages/02.item2/02.item2-2/2025-06-29T13:36:56.png` has been `git rm`'d and the commit amended (this branch had only the one commit, so the history no longer contains the colon path anywhere — verified with `git rev-list develop..HEAD` and by checking the pushed tree via the GitHub contents API). The test now generates the fixture at runtime with `imagepng(imagecreatetruecolor(...))` inside `testImageFilenameWithColonIsResolvedAsLocalMedia()`, uses it, and `unlink()`s it in a `finally` block, so nothing colon-named ever touches disk outside the test run. Went with generate-at-runtime over the copy-and-rename + Windows `markTestSkipped` approach since it avoids the problem entirely rather than working around it. Force-pushed the amended commit (5abcb51) to the branch.

system/src/Grav/Common/Page/Markdown/Excerpts.php
wakqasahmed: Merged the two adjacent scheme-handling branches into one `if (isset($url_parts['scheme']))` block that calls `schemeExists()` exactly once and reuses the boolean (`$is_registered_stream`) for both the rebuild-as-path branch and the stream-merge branch. Logic is unchanged — same four outcomes (rebuild path / stream merge+flag / stream flag only / leave untouched) as before, just without the duplicate locator fetch and `schemeExists()` call. Covered by the same 6 `ExcerptsTest` tests, all still green.

system/src/Grav/Common/Page/Markdown/Excerpts.php
wakqasahmed: Thanks for independently verifying this — confirms the direction. Left the guard's core logic as-is (still fixes both the digit-led and space-containing cases you traced), and added the letter-led ambiguous case (e.g. `photo:2025.png`) as an explicit comment note rather than attempting to fix it, per your suggestion, since it's genuinely indistinguishable from a real scheme without a locator-registered stream to disambiguate against. Also applied the same rebuild pattern to `Grav::fallbackUrl()` (system/src/Grav/Common/Grav.php), which had the identical `parse_url(rawurldecode($uri->basename()))` bug — a colon-named file there would miss the `$media[$media_file]` lookup and silently fall through to the generic download path instead of getting proper media-action handling. Same guard, no locator/schemeExists needed there since a URI basename never legitimately contains a stream scheme.

tests/unit/Grav/Common/Helpers/ExcerptsTest.php
wakqasahmed: Good catch on the letter-led overlap with ParsedownTest::testSpecialProtocols — left this test in place since it pins the invariant at the Excerpts layer specifically (this is where the fix lives), and it's cheap. Didn't add the @see cross-reference since it's a minor nit and the PR is otherwise already addressing the higher-priority Windows portability issue; happy to add it if you'd still like it. Fixture generation note: switched from a committed static file to a runtime-generated one (see my reply on the .png file comment) — the test now builds its own tiny PNG via imagecreatetruecolor()/imagepng() instead of relying on a checked-in binary, so the cropZoom/GD-decode path you confirmed is still exercised, just without a tracked colon-named file.

wakqasahmed:
Pushed a fixup addressing the review findings:

1. **Windows portability (main finding):** removed the colon-named test fixture (`2025-06-29T13:36:56.png`) from git entirely. The branch had a single commit, so it was amended to drop the file from history, not just delete it going forward — verified via `git rev-list develop..HEAD` and the pushed tree on GitHub. `ExcerptsTest::testImageFilenameWithColonIsResolvedAsLocalMedia` now generates the fixture at runtime (`imagecreatetruecolor()`/`imagepng()`), uses it, and cleans it up in a `finally` block.
2. **Same bug in `Grav::fallbackUrl()`:** applied the same RFC-3986-scheme-grammar rebuild to the `parse_url(rawurldecode($uri->basename()))` call in `system/src/Grav/Common/Grav.php`, so a colon-named media file resolves correctly there too instead of falling through to the generic download path.
3. **Minor dedup:** merged the two adjacent `isset($url_parts['scheme'])` blocks in `Excerpts::parseUrl()` into one, calling `schemeExists()` once instead of twice. No behavior change.

Verified via Docker (PHP 8.3 + ext-gd): `ExcerptsTest` — 6/6 passing; full unit suite — 974/974 passing, no regressions.

rhukster:
Merged, thanks — this is an unusually well-documented PR and the diagnosis is exactly right. `parse_url()` accepting a digit-leading "scheme" where RFC 3986 does not is the whole bug, and gating on the scheme grammar rather than a hardcoded allowlist was the right instinct; the `xmpp:` regression you caught in your own earlier iteration is a good illustration of why the allowlist approach was never going to hold. Verified before merging: 990 tests / 3154 assertions, all green.

Good call on generating the colon fixture at runtime rather than committing it — a `:` in a tracked filename would break `git clone` on Windows for the whole repo. I added a Windows skip guard and made the fixture path absolute so a hard-killed run can't leave a stray file behind.

The one substantive change on the way in: the fix was partial by construction. The grammar check only fires when the token before the colon starts with a digit or otherwise fails RFC scheme rules, so `note:2025.png`, `IMG:001.png` and `a:b:c.png` were still broken. Grammar alone can't separate those from an unknown protocol, so I added a second arm: a reference that carries a media extension (per `media.types`) is also rebuilt as a path. Grav's own external scheme list and `data:` are excluded outright, so no real protocol gets reinterpreted as a local file. There's a data provider covering the letter-led cases and a non-regression test for `mailto:`/`tel:`/`git:`.

For the record on the security question, since `parseUrl()` is shared by link and image handling: `javascript:` in all casings and `data:` skip the rebuild entirely, and the rebuild is monotonic in the safe direction — it only ever turns a scheme into a path, never the reverse.

The `Grav.php` hunk was the right call, incidentally, and reads as unexplained scope creep only because the description doesn't mention it — it's the same misparse in `fallbackUrl()`, the direct-media-request path. Worth a line in future PRs so a reviewer doesn't have to work it out.

Shipping in 2.0.20.

PR #4253, opened 2026-08-21
Fix invalid alt attributes on audio and video media

### LINKED ISSUE #3540: Non-standard `video` html generated by `media.html()`
## Grav twig template

```twig
{% if media.meta().type == 'video' %}
  {{ media.html() }}
{% endif %}
```

## Output

Current (after `1.7.31`):
```html
<video controls="controls" alt=""><source src="test.mp4">Your browser does not support the video tag.</video>
```

## Validator

Used: https://validator.w3.org/nu/

## Errors

* ~~Error: Bad value `1` for attribute `controls` on element [video](https://html.spec.whatwg.org/multipage/#the-video-element).~~
* Error: Attribute `alt` not allowed on element [video](https://html.spec.whatwg.org/multipage/#the-video-element) at this point.

## Reference

Specification `video`: https://html.spec.whatwg.org/multipage/media.html#the-video-element

## Possible solutions

* ~~` controls="1"` replace by `controls="controls"` or ` controls`~~
* `alt` replace by `title` or `aria-label` or `aria-describedby`

## History

* updated after partial fix in `1.7.31`

darkdi:
## What changed

- remove the invalid `alt` attribute from rendered `<video>` and `<audio>` elements
- preserve non-empty media alternative text as `aria-label`
- keep an explicitly configured `aria-label` unchanged
- add regression coverage for video, audio, empty alt text, and explicit aria labels

The generic media pipeline populates `alt` for every medium. That is valid for images but not for HTML audio/video elements, which is why `media.html()` can emit non-standard markup as reported in #3540.

This keeps the existing media metadata useful for assistive technology while making the generated player markup valid.

Fixes #3540

### DIFF
--- a/system/src/Grav/Common/Media/Traits/AudioMediaTrait.php
+++ b/system/src/Grav/Common/Media/Traits/AudioMediaTrait.php
@@ -43,6 +43,7 @@ trait AudioMediaTrait
     protected function sourceParsedownElement(array $attributes, $reset = true)
     {
         $location = $this->url($reset);
+        $attributes = $this->normalizePlayerAttributes($attributes);
 
         return [
             'name' => 'audio',
--- a/system/src/Grav/Common/Media/Traits/MediaPlayerTrait.php
+++ b/system/src/Grav/Common/Media/Traits/MediaPlayerTrait.php
@@ -104,6 +104,29 @@ trait MediaPlayerTrait
     }
 
     /**
+     * Normalize attributes for HTML audio and video elements.
+     *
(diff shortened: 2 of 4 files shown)


rhukster:
Thanks @darkdi — clean fix, and putting the normalization in the shared `MediaPlayerTrait` so both audio and video pick it up is exactly right. Preserving the alt text as `aria-label` keeps the accessibility metadata useful while making the markup valid, and the regression tests are a nice touch. Merged into `develop`; it'll ship in the next release. 🙏

PR #4257, opened 2026-08-22
[security] Restrict classes when unserializing lazy-index store rows

iliaal:
## Summary

The per-page index store (`PageIndexStore`, backing the experimental `system.pages.lazy_index` mode) reads its SQLite rows back with bare `unserialize()` — no `allowed_classes` restriction. Every other store Grav unserializes is gated: the blob pages cache verifies an HMAC before unserializing, and session flash objects carry a signed envelope. A tampered `cache://compiled/pages/*.sqlite` file can therefore instantiate arbitrary classes — a POP-chain entry point those other gates exist to deny.

This change:

- **`children` / `sorts` / `taxonomy` rows** are plain arrays by construction → unserialize with `['allowed_classes' => false]`, refusing object instantiation outright.
- **Page payloads** only ever contain `Grav\Common\Page\Page` objects plus plain `stdClass` data (verified by scanning every payload in a built store) → unserialize with those two classes named explicitly.

A corrupted or foreign-class row now falls through to the existing index-rebuild path instead of running code. Behavior for legitimate stores is unchanged; hydration was verified end-to-end (`Pages::get()` through `loadIndexedPage()`) with the allowlist active.

### DIFF
--- a/system/src/Grav/Common/Page/PageIndexStore.php
+++ b/system/src/Grav/Common/Page/PageIndexStore.php
@@ -289,7 +289,11 @@ final class PageIndexStore
                 return null;
             }
 
-            $value = @unserialize($payload);
+            // Rows in this table are plain arrays by construction. Refuse object
+            // instantiation outright so a tampered store file cannot be turned into
+            // a gadget-chain entry point; every other store Grav unserializes is
+            // gated the same way (FileCache HMAC, session flash v2 envelope).
+            $value = @unserialize($payload, ['allowed_classes' => false]);
 
             return is_array($value) ? $value : null;
         } catch (Throwable $e) {
@@ -313,7 +317,7 @@ final class PageIndexStore
 
             $result = [];
             foreach ($stmt->fetchAll(\PDO::FETCH_NUM) as [$path, $payload]) {
-                $info = @unserialize((string)$payload);
+                $info = @unserialize((string)$payload, ['allowed_classes' => false]);
                 $result[$path] = is_array($info) ? $info : [];
             }
 
@@ -335,7 +339,7 @@ final class PageIndexStore
 
             $map = [];
             foreach ($rows as [$type, $value, $path, $payload]) {
-                $info = @unserialize((string)$payload);
+                $info = @unserialize((string)$payload, ['allowed_classes' => false]);
                 $map[$type][$value][$path] = is_array($info) ? $info : [];
             }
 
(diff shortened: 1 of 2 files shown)


rhukster:
Thanks Ilia — you're right that this is the one store we left ungated, and the inconsistency is worth closing.

The three `PageIndexStore` hunks are correct and I'll take them: `children`, `sorts` and `taxonomy` rows are arrays by construction, the `is_array()` guards already downgrade anything else, and `allowed_classes => false` is a no-op on legitimate data.

The `Pages.php` half I need to push back on. The "`Page` + `stdClass`" survey holds for a stock site, but the payload graph isn't closed — `onPageProcessed` (`Pages.php:2243`) hands plugins the live object a few lines before `serializeIndex()` runs, and `Page` has no `__sleep()`. One listener calling `$page->media()`, which is what several image and media plugins do at build time, puts a `Grav\Common\Page\Media` into every payload. Under the allowlist that becomes:

```
hydrated as:              Grav\Common\Page\Page
->media property is:      __PHP_Incomplete_Class
->media() returned:       __PHP_Incomplete_Class
instanceof PageInterface: true
```

`MediaTrait::getMedia()` short-circuits on `null === $media`, so the incomplete object is handed straight back, and because the root is still a real `Page` the `loadIndexedPage()` rebuild path never fires. The site then fatals at render with "class definition ... was loaded _before_ unserialize()", which points at the autoloader rather than the store — and it re-poisons itself on every rebuild, so a cache clear doesn't fix it. (`allowed_classes` is exact-match too, so any `Page` subclass would break the same way.)

I think the answer is in the gate you cited. `FileCache` HMACs the blob and *then* does `allowed_classes => true`, and `Session::getFlashObject()` does the same. `JobQueue` is the only place we use a named allowlist, and it's explicitly defense-in-depth *behind* the HMAC on a single closed class. An HMAC over the page payload is strictly stronger here: it fails at the root, which drops straight into the existing `resetPages()` rebuild, and it catches mangled array rows as a bonus.

Would you be up for reworking it that way? Sketch: sign in `serializeIndex()` with `hash_hmac('sha256', $serialized, Security::getNonceKey())` in a `v1|<hmac>|<serialized>` shape, verify in a shared private helper used by both `hydrateIndexedPages()` and `loadIndexedPage()`, and bump `PageIndexStore::SCHEMA` 2 → 3 so existing stores invalidate cleanly instead of failing row by row. There's no test coverage for `lazy_index` at all right now, so a round-trip test — sign, tamper, confirm the rebuild fires — would be very welcome alongside it.

On classification: treating this as hardening rather than an advisory. `cache://compiled/pages/` is the sibling of `cache://compiled/config/*.php`, which we `include` directly (`CompiledBase.php:205`), so anyone who can plant a store file already has code execution without needing a POP chain — and `lazy_index` is off by default. It still deserves the gate for consistency with GHSA-gwfr-jfjf-92vv, and you'll have the changelog credit.

PR #4259, opened 2026-08-22
Preserve __Secure-/__Host- cookie name prefixes in session name

### LINKED ISSUE #3773: [BugReport] Session cookie naming, failure to add __Secure- prefix
Hi, I'm currently trying to set my session cookie to follow the MDN recommendation by adding the prefix `__Secure-` to my session cookie.

By putting `__Secure-session_cookie` in the session.name field of the `user/config/system.yaml` i get the following cookie name in my responses from the website `secure-session-cookie`, it seems that grav trims the leading underscores and puts everything in lowercase.

This issue appears with localhost and on a nginx hosted server.

yaml: 
```yaml
session:
  enabled: true
  initialize: true
  timeout: 1800
  name: __Secure-session-cookie
  uniqueness: path
  secure: true
  secure_https: true
  httponly: true
  samesite: Lax
  split: true
  domain: null
  path: null
```

result in website:
image

This appears using Microsoft Edge browser or by using wget.

Grav version: 1.7.43
Php version: 8.1.2

MDN link: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#cookie_prefixes

wakqasahmed:
Fixes #3773

The session cookie name (`system.session.name`) is always run through `Inflector::hyphenize()` in `SessionServiceProvider`. That helper strips leading underscores and lowercases the string, so a name like `__Secure-session-cookie` silently becomes `secure-session-cookie`.

Browsers only apply the `__Secure-` and `__Host-` cookie prefix protections when the cookie name is an exact, case-sensitive match, so this stripped the security benefit entirely without any error or warning to the site owner.

## Fix

`SessionServiceProvider` now only hyphenizes the configured session name when it isn't already a valid cookie name (RFC 6265 / RFC 2616 token). Names using a recognized cookie prefix (or any other already-valid cookie name) are preserved as-is; anything with characters that wouldn't be valid in a cookie name is still hyphenized exactly as before, so existing sites keep the same session cookie name after upgrading.

## Test plan

- Added `tests/unit/Grav/Common/Service/SessionServiceProviderTest.php` covering:
  - `__Secure-` prefixed names are preserved
  - `__Host-` prefixed names are preserved
  - the default `grav-site` name is unchanged
  - names with invalid characters are still hyphenized (existing behavior)
- `php -l` on both changed/added files
- Ran the full `tests/unit/Grav/Common` Codeception suite in a `php:8.3-cli` container after `composer install`; new tests pass, and the pre-existing `InflectorTest` suite is unaffected. The only failures/errors in that run are pre-existing environment gaps unrelated to this change (missing `gd`/`ZipArchive` extensions in the bare PHP image, and an unrelated `ComposerTest` path assertion).

### DIFF
--- a/system/src/Grav/Common/Service/SessionServiceProvider.php
+++ b/system/src/Grav/Common/Service/SessionServiceProvider.php
@@ -11,6 +11,7 @@ namespace Grav\Common\Service;
 
 use Grav\Common\Config\Config;
 use Grav\Common\Debugger;
+use Grav\Common\Inflector;
 use Grav\Common\Security;
 use Grav\Common\Session;
 use Grav\Common\Uri;
@@ -86,7 +87,7 @@ class SessionServiceProvider implements ServiceProviderInterface
                 $cookie_lifetime = 9999999999;
             }
 
-            $session_prefix = $c['inflector']->hyphenize($config->get('system.session.name', 'grav-site'));
+            $session_prefix = static::resolveSessionPrefix($c['inflector'], (string) $config->get('system.session.name', 'grav-site'));
             $session_uniqueness = $config->get('system.session.uniqueness', 'path') === 'path' ?  substr(md5(GRAV_ROOT), 0, 7) :  md5(Security::getNonceKey());
 
             $session_name = $session_prefix . '-' . $session_uniqueness;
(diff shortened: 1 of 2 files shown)


rhukster:
Merged, thanks — and thanks for tracking it back to the root cause. You're right that Grav was quietly rewriting the session cookie name and destroying the `__Secure-` / `__Host-` protection in the process. It's been open since 2023.

Worth being explicit about one thing, since the description says existing cookie names are preserved and that turns out not to be quite true. Keeping any name that's already legal for a cookie is a wider net than just the two prefixes: a site configured with `MySite`, `my_site` or a leading dash now gets a different cookie name than before, which signs its users out once on upgrade. I ran both versions side by side to check. We're taking that — very few sites change the session name at all, and the prefixes can't work without it — but it's called out in the changelog so nobody is surprised.

I added a few things on top:

**The prefixes come with strings attached.** A browser only accepts a `__Secure-` cookie over HTTPS, and only accepts a `__Host-` cookie when there's no domain set and the path is the site root. Grav's shipped config leaves `domain:` empty, which resolves to the request host a few lines further up — exactly what `__Host-` forbids — so the cookie would have been discarded by the browser with no error anywhere and nobody able to log in. The attributes now get lined up with whichever prefix is in use, so the name sticking and the cookie being valid go together.

**Dots are excluded** from the "already valid" test. PHP mangles dots in request variable names, so a cookie named `my.site` is stored by the browser but appears to PHP as `my_site`, and `Session::start()` can never resume it — a brand new session on every request rather than a one-time logout.

**The prefix is resolved separately from the rest of the name.** Treating the whole string as one token meant `__Secure-My Site` came out as `secure-my-site`, dropping the prefix in exactly the case this is meant to protect.

Tests went from four to twelve so each of those is pinned down — the original four passed under either approach, so they couldn't tell them apart.

Good find, and a genuinely useful fix.

PR #4261, opened 2026-08-23
Fix redirect_trailing_slash dropping custom_base_url path (#3822)

### LINKED ISSUE #3822: redirect_trailing_slash doesn't work with custom_base_url
`redirect_trailing_slash` funcionality doesn't respect `custom_base_url` path. For example `https://my.example.com/custombaseurl/page1/` will be redirected to `https://my.example.com/page1`

It looks like $url in Location: redirect here doesn't include `custom_base_url` at all: https://github.com/getgrav/grav/blob/ee8d783d050588aaf35261b7d1832e2bdc433fe6/system/src/Grav/Common/Grav.php#L500
Sadly, I'm not sure what is the best way to fix this.

wakqasahmed:
## Summary

`InitializeProcessor::handleRedirectRequest()` builds the trailing-slash
redirect target from the raw PSR-7 request URI path. When
`system.custom_base_url` is configured and the physical request reaching
PHP does not include that custom path segment (e.g. behind a reverse
proxy/router that strips the external prefix before forwarding to Grav),
the redirect dropped the custom base path entirely — reproducing exactly
the symptom in #3822 (`.../custombaseurl/page1/` redirecting to
`.../page1` instead of `.../custombaseurl/page1`).

The fix restores the custom base path onto the redirect target when the
raw request path doesn't already start with it, using the same
`Uri::rootUrl()` that already correctly resolves `custom_base_url`
elsewhere in Grav. When there's no custom base (or the physical path
already includes it, e.g. a front-controller setup where `REQUEST_URI`
carries the full external path), behavior is unchanged.

## Test plan

- Added `tests/unit/Grav/Common/Processors/InitializeProcessorTest.php`
  covering both the "physical path already includes the custom base"
  and the "physical path has the custom base stripped" scenarios.
- Verified the new stripped-path test fails against the pre-fix code
  (`Location: http://localhost:8080/page1` instead of
  `.../custombaseurl/page1`) and passes after the fix.
- `vendor/bin/codecept run unit --filter '/Uri|InitializeProcessor/'` —
  31 tests, 860 assertions, all passing (no regressions to existing
  `UriTest` custom-base-url coverage).
- Manually reproduced end-to-end with a live PHP built-in server
  (`system/router.php`) and `system.custom_base_url` configured,
  confirming the redirect `Location` header now preserves the custom
  base path in both scenarios.

Fixes #3822

### DIFF
--- a/system/src/Grav/Common/Processors/InitializeProcessor.php
+++ b/system/src/Grav/Common/Processors/InitializeProcessor.php
@@ -424,6 +424,15 @@ class InitializeProcessor extends ProcessorBase
         $path = $uri->getPath() ?: '/';
         $root = $this->container['uri']->rootUrl();
 
+        // When system.custom_base_url is set behind a proxy that strips the custom
+        // path segment before it reaches PHP, the raw request path won't include it,
+        // even though external URLs (and the redirect target) must. Restore it so the
+        // redirect doesn't drop the custom base path.
+        if ($root !== '' && $root !== '/' && !Utils::startsWith($path, $root)) {
+            $path = $root . $path;
+            $uri = $uri->withPath($path);
+        }
+
         if ($path !== $root && $path !== $root . '/' && Utils::endsWith($path, '/')) {
             // Use permanent redirect for SEO reasons.
             return $this->container->getRedirectResponse((string)$uri->withPath(rtrim($path, '/')), $code);
--- a/tests/unit/Grav/Common/Processors/InitializeProcessorTest.php
+++ b/tests/unit/Grav/Common/Processors/InitializeProcessorTest.php
@@ -1,0 +1,82 @@
+<?php
+
+use Codeception\Util\Fixtures;
+use Grav\Common\Grav;
+use Grav\Common\Processors\InitializeProcessor;
+use Nyholm\Psr7\Factory\Psr17Factory;
+
+/**
+ * Class InitializeProcessorTest
+ */
+class InitializeProcessorTest extends \PHPUnit\Framework\TestCase
+{
+    /** @var Grav $grav */
+    protected $grav;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $grav = Fixtures::get('grav');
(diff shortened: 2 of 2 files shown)


rhukster:
Landed, thanks — you found the right cause. The trailing-slash redirect was building its target from the raw request path and never re-applying the site's base path, and pulling that from `Uri::rootUrl()` was the right instinct.

I reworked two parts before merging it:

**Where the path comes from.** Some sites use `custom_base_url` to publish at a different path than the folder Grav actually lives in — installed at `/grav`, served at `/blog`. Prepending the base onto the raw path produces `/blog/grav/page1` for those, a URL that exists nowhere, which is worse than the bug being fixed. `Uri::init()` has already mapped the physical base onto the public one, so `Uri::uri(true)` is the answer; reading it from there handles both the proxy-stripped case and the remapped case.

**How "is the base already there" is decided.** `Utils::startsWith()` is a raw text prefix test, so `/subscribe` counted as being under a base of `/sub` and the fix was skipped — leaving the original bug live for every route whose first segment merely starts with the same letters. It now compares whole path segments.

There's also a guard against building a redirect from a malformed `rootUrl()`, which can happen under `reverse_proxy_setup` when a port is involved.

On tests: 33 now, and the thing I most wanted to be sure of is that ordinary installs are untouched. Running the suite against unfixed develop, 22 fail — and every one that passes there is a no-custom-base or already-correct case, so the behaviour on a normal site is provably identical before and after. There's coverage for docroot and subfolder installs, prefix-lookalike routes, full-URL and host-only bases, percent-encoding, query strings, non-GET methods and the segment comparison itself.

One nice side effect worth knowing: the homepage of a proxy-stripped custom-base site previously redirected to the bare domain with nothing after it. That's fixed too.

Appreciate the clear write-up and the tests — that made this quick to verify.

rhukster:
Closing this out — the fix is on `develop` as efeb262f7, with the two reworks described above and a test suite around it. You're credited in the commit and the changelog. Thanks again for finding the root cause.

PR #4262, opened 2026-08-23
Add Collection::notOfType() to exclude page templates

### LINKED ISSUE #3910: [FeatureRequest] Twig: Filter a Collection by notOfType
Currently Grav supports filtering a Page Collection by including specific templates types via [Collection::ofType](https://github.com/getgrav/grav/blob/master/system/src/Grav/Common/Page/Collection.php#L606).

I would also like to **exclude** specific template types from a Page Collection, so I can display the results differently.

This could be achieved via `Collection::notOfType`. 
Preferably this new function accepts an Array or List for to remove multiple templates at once.

**Acceptance Criteria**
* In Twig It is possible to exclude a single/multiple template types from a Page Collection

**Solution**
* `page.collection()|filter(module => module.template != 'modular/links')` where `modular/links` is the template to exclude

wakqasahmed:
Fixes #3910

## Summary

Adds `Collection::notOfType($type)`, the inverse of the existing `Collection::ofType()` / `Collection::ofOneOfTheseTypes()`, so a page collection can exclude one or more template types instead of only including them.

`$type` accepts either a single template name (string) or an array of template names, matching the acceptance criteria in the issue ("exclude a single/multiple template types").

Exposed the same way as the existing type filters so it is directly callable from Twig:

```twig
{% for p in page.collection().notOfType('modular/links') %}
    ...
{% endfor %}
```

As @rhukster invited on the issue:

> How about a PR? Seems you have a need, have a solution, just create a PR for this feature, and it could get merged.

## Changes

- `system/src/Grav/Common/Page/Collection.php` — new `notOfType()` method, mirrors `ofType()`/`ofOneOfTheseTypes()` with the match inverted.
- `system/src/Grav/Common/Page/Interfaces/PageCollectionInterface.php` — interface declaration.
- `system/src/Grav/Common/Flex/Types/Pages/PageCollection.php` — Flex collection implementation + registered in `getCachedMethods()` alongside `ofType`/`ofOneOfTheseTypes`.
- `system/src/Grav/Common/Flex/Types/Pages/PageIndex.php` — proxy method delegating to the Flex collection via `__call()`, same pattern as `ofType`/`ofOneOfTheseTypes`.
- `tests/unit/Grav/Common/Page/PagesTest.php` — new `testOfTypeAndNotOfType()` covering single-type and multi-type exclusion, and that `ofType` + `notOfType` partition the full collection.

## Test plan

- [x] `php -l` on all changed files
- [x] Full `composer install` + Codeception unit suite for `tests/unit/Grav/Common/Page` (73 tests, 225 assertions) — all green, including the new test
- [x] Manually verified `notOfType` and `ofType` results are complementary/non-overlapping on the fixture page set

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 1. [](#new)
     * The `url()` Twig function now takes a language, so a link to a route that isn't a page - a search page, a form action - can carry the site's language prefix: `{{ url('/search', lang=true) }}`
+    * Page collections can now exclude one or more template types with `notOfType()`, the counterpart to the existing `ofType()` [#3910](https://github.com/getgrav/grav/issues/3910)
 1. [](#improved)
     * Building a URL is now faster, which adds up over the hundreds of asset and link URLs a single page render produces
 1. [](#bugfix)
--- a/system/src/Grav/Common/Flex/Types/Pages/PageCollection.php
+++ b/system/src/Grav/Common/Flex/Types/Pages/PageCollection.php
@@ -93,6 +93,7 @@ class PageCollection extends FlexPageCollection implements PageCollectionInterfa
                 'nonRoutable' => true,
                 'ofType' => true,
                 'ofOneOfTheseTypes' => true,
+                'notOfType' => true,
                 'ofOneOfTheseAccessLevels' => true,
                 'withOrdered' => true,
                 'withModules' => true,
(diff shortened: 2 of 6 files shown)


rhukster:
Thanks for picking this up — this is a clean piece of work, and you covered all four places the feature needed to touch, which is more than most first passes get right. In particular you registered the new method with the Flex page cache and added the matching method on the page index; those two are easy to miss and leave you with a feature that quietly works but never caches, or one that breaks on Flex sites. You also matched the two different styles the old and new page collections use internally rather than copy-pasting one into the other.

Accepting either a single template name or a list was the right reading of the request, and `notOfType('modular/links')` does exactly what #3910 asked for.

Two small things before this goes in:

1. A CHANGELOG entry under the `[](#new)` heading in the top block — something like *"Page collections can now exclude one or more template types with `notOfType`, the counterpart to the existing `ofType`"* with a link back to #3910.
2. In the new test, the multi-type check would still pass if the filter returned nothing at all. Adding `self::assertGreaterThan(0, count($withoutBlogOrDefault->toArray()));` after that loop closes the gap.

One thing on my side rather than yours: this adds a method to a public interface, which is a bigger deal in a patch release than a feature one, so I may hold the merge for the next feature version. That's a scheduling decision and doesn't need anything from you.

Nice job.

rhukster:
Merged into `develop` as f58b3ea8d. Both follow-ups were exactly what was asked for, and I re-ran the full unit suite on your head commit before merging: 1059 tests, 3273 assertions, all green across PHP 8.3/8.4/8.5 in CI too.

Thanks for the clean work on this one, and for spotting that the branch had moved past 2.0.21 and putting the changelog entry in the right block.

One note for anyone reading this later rather than anything you need to do: since this adds a method to a public interface, the entry may move out of the 2.0.22 block when the release it ships in is decided. That is on us at release time, not on the PR.

PR #4265, opened 2026-08-26
Apply force_ssl redirect on 404 fallback page too (#3703)

### LINKED ISSUE #3703: force_ssl doesn't work on 404 notfound page
When querying http://gravsite.example.com/pagedoesnexist with force_ssl option enabled user is not redirected to HTTPS. That's because https://github.com/getgrav/grav/blob/c261d0d3f749ecf2dc2c9f88cb3aec0bd4bfa785/system/src/Grav/Common/Service/PagesServiceProvider.php#L72 is enforced only on existing pages.
Thought to do a patch myself but was not sure about the best way to fix this. From what I understand by looking at the code the best place is to move this check to Page class itself, but I don't know if that would not be a problem for file downloads or other requests.

wakqasahmed:
Closes #3703.

When `system.force_ssl` is enabled and the requested path doesn't match a real page, the redirect check in `PagesServiceProvider` never runs. It's only inside the `if ($page)` block, and `$page` is null right up until the notfound fallback gets built further down, so a 404 request over plain HTTP just serves the 404 page instead of bouncing to HTTPS first.

I moved the force_ssl check up so it runs right after `$pages->dispatch($path)`, before the found/not-found branch, instead of duplicating it on both paths. That way it applies to every request regardless of whether a page was found.

Added a test in `PagesServiceProviderTest` that drives the actual `page` container service with force_ssl on: one case hitting an existing page (already worked, checked it doesn't regress) and one hitting a path with no matching page. Confirmed the new test fails against the old code (no redirect thrown) and passes with the fix.

### DIFF
--- a/system/src/Grav/Common/Service/PagesServiceProvider.php
+++ b/system/src/Grav/Common/Service/PagesServiceProvider.php
@@ -60,19 +60,19 @@ class PagesServiceProvider implements ServiceProviderInterface
             $path = $uri->path() ? urldecode($uri->path()) : '/'; // Don't trim to support trailing slash default routes
             $page = $pages->dispatch($path);
 
+            if ($config->get('system.force_ssl')) {
+                $scheme = $uri->scheme(true);
+                if ($scheme !== 'https') {
+                    $url = 'https://' . $uri->host() . $uri->uri();
+                    $grav->redirect($url);
+                }
+            }
+
             // Redirection tests
             if ($page) {
                 // some debugger override logic
                 if ($page->debugger() === false) {
                     $grav['debugger']->enabled(false);
-                }
-
-                if ($config->get('system.force_ssl')) {
-                    $scheme = $uri->scheme(true);
-                    if ($scheme !== 'https') {
-                        $url = 'https://' . $uri->host() . $uri->uri();
-                        $grav->redirect($url);
-                    }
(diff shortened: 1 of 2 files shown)


rhukster:
Merged, thanks — this one had been sitting open since 2023 and your diagnosis was exactly right. The SSL check really was stuck inside the branch that only runs for pages that exist, so anything falling through to the 404 handler never got the chance to bounce to HTTPS.

I checked the obvious risk before taking it, which was whether anything else in the request lifecycle already enforces HTTPS and would now double-redirect. Nothing does — `force_ssl` has exactly one call site in the whole codebase, so hoisting it is safe. CLI and the `/api` routes are unaffected, since neither resolves a page through that closure.

I pushed one small thing on top. Because the check now runs for requests that previously fell through, it also covers media and asset fallback requests, which makes the empty-hostname case from #3702 reachable in more places — it would have built `https:///path`. Added a host guard so it skips the redirect instead.

Worth noting for anyone who finds this later: `force_ssl` behind a reverse proxy still needs `http_x_forwarded.protocol` turned on, or the scheme is misread and the site redirect-loops. That's pre-existing, but this change makes it reachable on more URLs.

Your test is good and I kept it — it fails without the fix and passes with it, which is exactly what I want.

PR #4266, opened 2026-08-26
Skip image-only defaults for non-image media (#4264)

### LINKED ISSUE #4264: [BUG] Image configuration is being applied to non-image media
`system.yaml`
```
images:
  cls:
    auto_sizes: true
    retina_scale: '1'
  defaults:
    loading: lazy
    link: true
```

When `link: true`, non-image files such as `.mp3` get turned into links and `loading=lazy` is also added.

wakqasahmed:
`system.images.defaults` gets applied to every medium that comes through `Excerpts::processMediaActions()`, not just images. If you set `loading: lazy` (or `decoding`/`fetchpriority`) there, it also gets called on audio/video/document media embedded via markdown image syntax, e.g. ``.

`loading()`/`decoding()`/`fetchpriority()` only exist as real methods on `ImageMedium`/`StaticImageMedium` (via their traits). For anything else, the call falls through to the generic `__call()` on `MediaObjectTrait`, which just appends `loading=lazy` etc. to the medium's querystring — so it leaks into the rendered output for media types where it means nothing.

Fixed by skipping those three default actions in the defaults loop when the medium isn't `ImageMediaInterface`. Left `link` and the other defaults alone since those are legitimate for any media type.

Added a test confirming a non-image medium (mp3) no longer picks up `loading`/`decoding`/`fetchpriority` from `system.images.defaults`, plus a non-regression test that real images still get them.

Closes #4264

### DIFF
--- a/tests/unit/Grav/Common/Helpers/ExcerptsTest.php
+++ b/tests/unit/Grav/Common/Helpers/ExcerptsTest.php
@@ -211,6 +211,50 @@ class ExcerptsTest extends \PHPUnit\Framework\TestCase
         ];
     }
 
+    public function testImageDefaultsAreNotAppliedToNonImageMedia(): void
+    {
+        // getgrav/grav#4264: system.images.defaults (loading/decoding/fetchpriority)
+        // must only apply to actual image media, not every medium type embedded
+        // via markdown image syntax (audio, video, documents, ...).
+        $this->config->set('system.images.defaults', [
+            'loading' => 'lazy',
+            'decoding' => 'async',
+            'fetchpriority' => 'high',
+            'link' => true,
+        ]);
+
+        $fixturePath = GRAV_ROOT . '/tests/fake/nested-site/user/pages/02.item2/02.item2-2/sample-audio.mp3';
+        file_put_contents($fixturePath, '');
+
+        $this->page->media(new Media($this->page->getMediaFolder(), $this->page->getMediaOrder()));
+
+        try {
+            $result = Excerpts::processImageHtml('<img src="sample-audio.mp3" alt="Sample Audio" />', $this->page);
+
+            self::assertStringNotContainsString('loading=', $result);
+            self::assertStringNotContainsString('decoding=', $result);
+            self::assertStringNotContainsString('fetchpriority=', $result);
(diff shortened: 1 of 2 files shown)


tests/unit/Grav/Common/Helpers/ExcerptsTest.php
```diff
@@ -211,6 +211,50 @@ public function letterLedColonFilenameProvider(): array
...
+
+        $this->page->media(new Media($this->page->getMediaFolder(), $this->page->getMediaOrder()));
+
+        try {
+            $result = Excerpts::processImageHtml('<img src="sample-audio.mp3" alt="Sample Audio" />', $this->page);
+
+            self::assertStringNotContainsString('loading=', $result);
+            self::assertStringNotContainsString('decoding=', $result);
+            self::assertStringNotContainsString('fetchpriority=', $result);
+            // 'link' is a universal action and should still be honoured.
```
3e33, line 237: The issue is image configuration being applied to non-images, when link is under the image configuration, it should not be applied to non-images.

rhukster:
Merged, thank you — and this turned out to be a more valuable find than the issue suggested, because the leak happens on every Grav site out of the box, not just ones that changed these settings. The shipped defaults are `auto` for all three, which is a non-empty array, so every audio, video and SVG URL has been picking them up.

I pushed two adjustments on top:

The interface check doesn't quite hold on its own. `ImageMediaInterface` is an empty marker interface, and an SVG or animated GIF is a `StaticImageMedium`, which carries only the loading trait — no `decoding()` or `fetchpriority()`. So those two still ended up on an SVG's URL after the patch. I swapped the name list for a check that the medium actually has the method, which also means the list can't go stale against what is a free-form config map, and it matches the allowlist a few lines above.

I also gated `link`. I know you left it deliberately, but linking an audio or video file replaces the whole player with a static placeholder image, which is the reporter's actual complaint — the attribute leak was the part they noticed second. Anyone who wants a linked file can still ask for it per item with `Doc`.

Your test needed one assertion inverted as a result (it asserted the anchor as correct), and I added a case for the SVG leak. Good catch, and thanks for the test coming with it.

PR #4278, opened 2026-09-02
Convert true/false in GRAV_CONFIG__ env overrides to real booleans

### LINKED ISSUE #4277: Bug: .env values true/false misinterpreted
Environment variables only ever carry strings. Left as-is, an override such as GRAV_CONFIG__system__cache__enabled=false lands in config as the *string* "false" - and PHP's (bool) cast (used throughout core, e.g. Cache::init()) treats any non-empty, non-"0" string as true, silently doing the opposite of what was asked. Coerce the two boolean literals; every other value (numbers, arbitrary strings) is left untouched, matching prior behavior.

InitializeProcessor.php. Line 235
```
                if (is_string($value)) {
                    if (strcasecmp($value, 'true') === 0) {
                        $value = true;
                    } elseif (strcasecmp($value, 'false') === 0) {
                        $value = false;
                    }
                }
```

AdilAzhariOmsan:
Fixes #4277.

Env vars are always strings, so setting `GRAV_CONFIG__system__cache__enabled=false` puts the string `"false"` into the config. Core reads that flag (and others like it) with a `(bool)` cast, and `(bool)"false"` is `true`, so the override does the opposite of what you'd expect. Nothing in the logs points at it either.

The env override loop in `initializeConfig()` now passes each value through a small `castEnvironmentValue()` helper before `$config->set()`. It converts the literal strings `true`/`false` (case-insensitive) to booleans and leaves everything else alone - numbers, `"0"`/`"1"`, arbitrary strings all behave exactly as before, so existing setups aren't affected.

Tests are in `InitializeProcessorTest` using the same reflection + data provider approach the file already has for `pathHasBase`. `vendor/bin/codecept run unit` is green.

Left `CHANGELOG.md` alone since there's no unreleased section right now - happy to add a line wherever you want it.

### DIFF
--- a/system/src/Grav/Common/Processors/InitializeProcessor.php
+++ b/system/src/Grav/Common/Processors/InitializeProcessor.php
@@ -232,6 +232,7 @@ class InitializeProcessor extends ProcessorBase
                 foreach ($aliases as $alias => $real) {
                     $key = str_replace($alias, $real, $key);
                 }
+                $value = static::castEnvironmentValue($value);
                 $list[$key] = $value;
                 $config->set($key, $value);
             }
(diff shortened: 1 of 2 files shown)


rhukster:
Thanks, this is a clean fix and I'm taking it.

One thing worth recording, because it explains why this was never caught: Grav 1.7 users mostly hit `.env` through the DotEnv plugin, and that plugin did coerce types — it ran `json_decode` over `true`, `false`, `null` and anything numeric. But it did that on its own `.gravenv` file with dot-notation keys, which is a completely separate path from `GRAV_CONFIG__`. That prefix has never coerced anything, in 1.7 or 2.0. So when 2.0 absorbed `.env` reading natively, the type handling didn't come with it, and anyone moving off the plugin quietly lost it.

I'm glad you scoped this to the two boolean literals and left numbers alone. The old plugin coerced numerics too, and `is_numeric('1.10')` is true, so `json_decode` would turn a version string into `1.1`. Leading zeros have the same problem. Booleans are the only case where the intent is unambiguous.

Two things I handled on merge:

Your note that there's no unreleased CHANGELOG section isn't quite right — there's a `v2.0.23` block at the top, above tag `2.0.22`. I added the bullet there and credited you both.

The docs needed updating too, and they're the reason this bug survived so long: the example in the configuration docs is `GRAV_CONFIG__system__cache__enabled=true`, which appears to work only because `(bool)"true"` happens to be true. Anyone who copied it and flipped it to `false` got the opposite of what they asked for, with nothing in the logs to say so. I've switched the example to the case that used to be broken and documented the conversion.

Nice tests too — `'falsey'`, the padded `' true '` and the already-boolean cases are exactly the edges I'd have asked for.

AdilAzhariOmsan:
@rhukster Thanks for taking it, good call on the CHANGELOG section I misread the v2.0.23 block as released. and thanks for handling the docs example that's a better fix than anything I'd have sent for it.

Issue #4260, opened 2026-08-23
Unwritable cache/compiled/config is an uncaught fatal — the Problems plugin can never report it


sandymac:
### Environment

- **Grav:** 2.0.21
- **PHP:** 8.4
- **OS:** FreeBSD, on a shared host where the web server and the CLI run as different users
- Problems plugin installed and enabled

### What happens

If the web-server user cannot write `cache/compiled/config`, `ConfigServiceProvider::saveCachedFileList()` fails and `AbstractFile::save()` throws:

```php
if ($tmp === false) {
    throw new RuntimeException('Failed to save file ' . $filename);
}
```

Nothing catches it. Every route returns 500 — the front end *and* `/admin` — so there is no in-browser path to recovery. Only SSH or FTP gets you out.

```
PHP Fatal error: Uncaught RuntimeException: Failed to save file
  <site>/cache/compiled/config/filelist-config-<env>.php
  in <site>/vendor/rockettheme/toolbox/File/src/AbstractFile.php:374
#1 ConfigServiceProvider.php(430): PhpFile->save(Array)
#2 ConfigServiceProvider.php(158): ConfigServiceProvider::saveCachedFileList(...)
#3 ConfigServiceProvider.php(55): ConfigServiceProvider::load(...)
#4 Pimple/Container.php(147)
#5 InitializeProcessor.php(174): Container->offsetGet('config')
#6 InitializeProcessor.php(73): InitializeProcessor->initializeConfig()
```

### Why the Problems plugin cannot help

This is the case the Problems plugin exists for, but it is structurally unable to catch it. `problems.php` subscribes to `onPluginsInitialized`, and registers its `onFatalException` handler *inside* that callback:

```php
public function onPluginsInitialized(): void
{
    // ...
    'onFatalException' => ['onFatalException', 0],
}
```

The fatal occurs in `InitializeProcessor::initializeConfig()` (line 73), well before the handler chain at line 129 that runs `PluginsProcessor`. Plugins are never initialized, so the fatal handler is never registered, so the diagnostic page never renders.

### Why this is worth fixing

The compiled config file is **cache** — by definition reconstructible. Failing to write it should not be fatal. Grav can serve the request by parsing the YAML directly, at a performance cost.

The message is also not actionable. "Failed to save file ..." gives no hint that this is a directory permission problem, which is what it almost always is.

### Steps to reproduce

1. Install Grav and load any page once, so `cache/compiled/config` exists.
[...]

PR #4285, opened 2026-09-04
Fix GRAV_CONFIG env override gate to also check $_SERVER/$_ENV

### LINKED ISSUE #4279: getenv()-only reads in the bootstrap: the GRAV_CONFIG gate and the GRAV_ENVIRONMENT family
Follow-up to #4275, which fixed `Uri::ip()` reading request variables through `getenv()` alone. Two more spots in the bootstrap share the same bug class. Neither is security-relevant and neither affects a per-request path, but both fail silently, which is what makes them worth fixing.

### 1. The `GRAV_CONFIG` gate disagrees with its own body

`system/src/Grav/Common/Processors/InitializeProcessor.php:209`

```php
$prefix = 'GRAV_CONFIG';
$env = getenv($prefix);          // gate: process environment only
if ($env) {
    ...
    $env = $_ENV + $_SERVER;     // body, eleven lines later: reads $_SERVER
```

The gate and the loop it guards read different sources. Set `GRAV_CONFIG` with Apache `SetEnv` or nginx `fastcgi_param` on a host whose SAPI does not answer `getenv()`, and the gate is false while every `GRAV_CONFIG__*` key sits unread in `$_SERVER`. The entire environment-override feature is skipped with nothing logged.

Confirmed that `$_SERVER` alone does not satisfy the gate:

```php
$_SERVER['GRAV_CONFIG'] = 'true';
var_dump(getenv('GRAV_CONFIG'));   // bool(false)
```

### 2. `Setup.php` reads the `GRAV_ENVIRONMENT` family through `getenv()` only

`system/src/Grav/Common/Config/Setup.php:185, 207, 240, 243` — `GRAV_ENVIRONMENT`, `GRAV_SETUP_PATH`, `GRAV_ENVIRONMENT_PATH`, `GRAV_ENVIRONMENTS_PATH`.

Meanwhile `system/src/Grav/Common/Config/Env.php:74, 93, 129` already reads the *same* `GRAV_ENVIRONMENT` key as `$_SERVER[...] ?? $_ENV[...] ?? (getenv(...) ?: null)`, which is exactly the pattern #4275 introduced. So core already contains the fix, applied inconsistently.

This one is partly mitigated: `Env.php` deliberately constructs Dotenv with `usePutenv(true)` *because* `Setup.php` reads `getenv()`, so `.env`-supplied values work. The gap is only for server-supplied values on a non-answering SAPI.

### Suggested fix

Use the `Env.php` pattern in both places — `$_SERVER` first, then `$_ENV`, then `getenv()` — so a variable set by the server config is honoured however the SAPI exposes it. For the `GRAV_CONFIG` gate specifically, reading the same `$_ENV + $_SERVER` array the body already builds would be the smallest change and removes the disagreement entirely.

wakqasahmed:
The `GRAV_CONFIG` override gate in `InitializeProcessor::initializeConfig()` checks `getenv($prefix)`, but the body one line below already reads `$_ENV + $_SERVER`. On Apache with `SetEnv` or nginx with `fastcgi_param`, the variable lands in `$_SERVER` but never in the process environment `getenv()` reads, so the gate is false and the whole `GRAV_CONFIG__*` override feature does nothing — same failure mode as #4260/#4275.

`Setup.php` has the identical pattern for `GRAV_ENVIRONMENT`, `GRAV_SETUP_PATH`, `GRAV_ENVIRONMENT_PATH` and `GRAV_ENVIRONMENTS_PATH`.

`Env.php` already solves this correctly (`$_SERVER[$key] ?? $_ENV[$key] ?? (getenv($key) ?: null)`) and has its own test coverage proving the precedence. This PR applies the same fallback at all five call sites instead of inventing a new pattern.

Added a test that sets the override only in `$_SERVER` and drives the real `initializeConfig()` — confirmed it fails against the old code and passes against the fix. Ran the full `InitializeProcessorTest` and `EnvTest` suites after, both green (59 tests).

Fixes #4279

### DIFF
--- a/system/src/Grav/Common/Config/Setup.php
+++ b/system/src/Grav/Common/Config/Setup.php
@@ -182,7 +182,7 @@ class Setup extends Data
 
         // If environment is not set, look for the environment variable and then the constant.
         $environment = static::$environment ??
-            (defined('GRAV_ENVIRONMENT') ? GRAV_ENVIRONMENT : (getenv('GRAV_ENVIRONMENT') ?: null));
+            (defined('GRAV_ENVIRONMENT') ? GRAV_ENVIRONMENT : ($_SERVER['GRAV_ENVIRONMENT'] ?? $_ENV['GRAV_ENVIRONMENT'] ?? (getenv('GRAV_ENVIRONMENT') ?: null)));
 
         // If no environment is set, make sure we get one (CLI or hostname).
         if (null === $environment) {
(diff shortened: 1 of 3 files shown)


wakqasahmed:
Closing in favor of #4286

Issue #3643, opened 2022-10-26
1.7.32: Why impose max sizes for field types? Admin fails on large content sizes.


pamtbaau:
In Grav 1.7.32, in commit [3e7f67f](https://github.com/getgrav/grav/commit/3e7f67f589267e61f823d19824f3ee1b9a8a38ff), max sizes for field types have been imposed.

According the CHANGELOG:
- Set default maximum length for text fields
  - password: 256
  - email: 320
  - text, url, hidden, commalist: 2048
  - text (multiline), textarea: 65536

Why have these limits been imposed? Grav seems to be working fine when pages have content larger then 65536. A page with size 86691 (500 [lorum ipsum](https://www.lipsum.com/) paragraphs) is working just fine. But when saving a large page in Admin, an error is thrown after Grav core validates the page's content.

This has lead to issues when users create pages with long content using Admin. 
- Github Admin: [Limitation of large .md failing to save in admin (unless in expert mode) #2308](https://github.com/getgrav/grav-plugin-admin/issues/2308)
- Discourse: [Invalid input in “Content” Failed to save entry: Form validation failed, please check your input](https://discourse.getgrav.org/t/invalid-input-in-content-failed-to-save-entry-form-validation-failed-please-check-your-input/21505)]

Nerothos:
Can we get an update on this?
I'm working on a markdown table listing, it has "only" 280 rows (forced by updating the file directly) and 11 columns and it already broke at the 172 rows. I need to have  almost 10 times that number.
This is really too limiting to need to directly change files when the admin should suffice

mke21:
This is still a problem. Some informative sites have seriously long textual content per page. Please consider to make the limit much larger in order to accommodate these sites.

rhukster:
Fixed for the next release of both lines — `1.7.53.3` and `2.0.22`.

The multiline default has gone from 65,536 characters to 2,000,000. It was only ever meant as a guard against a runaway payload, and 65,536 is a MySQL `TEXT` column size, which is a strange ceiling for a CMS that writes Markdown to a flat file. Nothing downstream needed it: the request body is already bounded by `post_max_size`, and the person hitting the limit is an authenticated editor saving their own page. Single-line fields keep their 2,048 default. You can also set `max: 0` on any field to remove the check entirely.

A field rejected purely for its length now says so, and gives both numbers:

```
Invalid input in "Content" (150 characters, maximum is 100)
```

[...]

PR #4289, opened 2026-09-06
Stop the page Content tab from shadowing the content field (#4271)

### LINKED ISSUE #4271: Page blueprint's `content` field is shadowed by the `content` tab, so page content is never validated
In `system/blueprints/pages/default.yaml` the Content **tab** and the Content **field** are both keyed `content`:

```yaml
fields:
  tabs:
    type: tabs
    fields:
      content:            # <- the tab
        type: tab
        title: PLUGIN_ADMIN.CONTENT
        fields:
          content:        # <- the markdown field
            type: markdown
            validate:
              type: textarea
```

`BlueprintSchema` flattens every field into a single map keyed by name, containers included, so the two collide and the tab wins:

```
items['content']['type']     = 'tab'
items['content']['validate'] = null
nested['content']            = 'content'
```

`BlueprintSchema::validateArray()` resolves the rule for the posted `content` value through that map, gets the tab, and looks for a `typeTab()` validator. There isn't one, so validation silently succeeds no matter what the value is.

### Effect

Page content on the default page type is **not validated at all**. Whatever the blueprint says about it — `validate.type`, any length or pattern rule — is inert, and `checkSafety()` runs against the tab rather than the markdown field.

This is how [#3643](https://github.com/getgrav/grav/issues/3643) went so long without a clean reproduction: the 65,536-character multiline cap could not fire on a stock install, only on theme and plugin page blueprints that declare `content` outside a same-named tab.

### Reproduction

```php
$bp = new Blueprint('default');
$bp->setContext('/path/to/grav/system/blueprints/pages');
$bp->load()->init();

var_dump($bp->schema()->getState()['items']['content']['type']);  // 'tab'
```

Confirmed on stock 1.7.49.5 and 2.0.7.

### Why this is not a quick fix

Repairing the collision turns on type validation *and* `checkSafety()` XSS checking for default-page content for the first time. Sites that have been saving content those checks would now reject would start failing on save, so this needs its own look at what should actually be enforced on a page body, not just a rename.

Filed to track; not scheduled.

wakqasahmed:
Fixes #4271

In `system/blueprints/pages/default.yaml`, the Content tab and the markdown content field are both keyed `content`. BlueprintSchema flattens every field into one map keyed by name, containers included, and it does this by writing the field's own entry after recursing into its children, so the tab's entry gets written last and wins. The markdown field's `validate` config sits in that map under `content` too, but it's never the one that gets read back, so it never actually runs.

I renamed the tab's key to `content_tab` and left the field's key alone, since `content` is the actual page body key used throughout Grav and themes and isn't safe to touch. `modular.yaml` and `external.yaml` both extend `default.yaml` and override that same tab by key, so I updated them to `content_tab` too. Otherwise their overrides would stop merging into the tab and would themselves collide with the content field the same way.

I checked how much this actually changes at runtime before assuming a rename alone was safe. `checkSafety()` (the XSS scan) already runs on this field today regardless of which rule wins, since it only skips fields typed `unset`, so no new XSS behavior gets switched on here. Type validation is the part that starts working, and the field already has `max: 0` set from #3643 to opt out of the length cap, so in practice this now only requires content to be a string, which is what a textarea always posts anyway.

Added a test that loads the real `pages/default.yaml` blueprint and asserts the flattened `content` item resolves to the markdown field's type and validate config rather than the tab's.

Ran the existing blueprint/data tests plus the full page test suite locally, all green, and confirmed the new test fails against the old blueprint (resolves to type `tab`) and passes with the fix.

### DIFF
--- a/system/blueprints/pages/default.yaml
+++ b/system/blueprints/pages/default.yaml
@@ -16,7 +16,7 @@ form:
       active: 1
 
       fields:
-        content:
+        content_tab:
           type: tab
           title: PLUGIN_ADMIN.CONTENT
 
--- a/system/blueprints/pages/external.yaml
+++ b/system/blueprints/pages/external.yaml
@@ -13,7 +13,7 @@ form:
 
       fields:
 
-        content:
+        content_tab:
           fields:
 
             header.title:
--- a/system/blueprints/pages/modular.yaml
+++ b/system/blueprints/pages/modular.yaml
@@ -8,7 +8,7 @@ form:
       active: 1
 
       fields:
-        content:
+        content_tab:
           fields:
 
             modular_title:
--- a/tests/unit/Grav/Common/Data/PageBlueprintContentTest.php
+++ b/tests/unit/Grav/Common/Data/PageBlueprintContentTest.php
@@ -1,0 +1,37 @@
+<?php
+
+use Codeception\Util\Fixtures;
+use Grav\Common\Data\Blueprint;
+
+/**
+ * Regression test for #4271: the Content tab and the markdown `content` field
+ * in the default page blueprint were both keyed `content`, so BlueprintSchema
+ * flattened them into the same slot and the tab (added after its own fields
+ * during parsing) always won, leaving the field's `validate` config
+ * unreachable.
+ */
+class PageBlueprintContentTest extends \PHPUnit\Framework\TestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $grav = Fixtures::get('grav');
+        $grav();
+    }
+
+    public function testContentFieldIsNotShadowedByTab(): void
+    {
+        $blueprint = new Blueprint('default');
+        $blueprint->setContext('blueprints://pages');
+        $blueprint->load()->init();
+
+        $items = $blueprint->schema()->getState()['items'];
+
+        self::assertArrayHasKey('content', $items);
+        self::assertSame('markdown', $items['content']['type']);
+        self::assertSame('textarea', $items['content']['validate']['type'] ?? null);
+
+        self::assertArrayHasKey('content_tab', $items);
+        self::assertSame('tab', $items['content_tab']['type']);
+    }
+}


rhukster:
Thanks Waqas, and I want to be clear up front that the diagnosis is right. I filed #4271 myself, you found the same thing I did, and you did the homework I asked for in the issue. I verified all three of your claims and they hold: `checkSafety()` does already run today (the early return only fires on type `unset`), `max: 0` does keep the length cap off, and the filter path is unchanged, since `tab` and `textarea` both fall through to `filterText`. The only new enforcement really is "content must be a string". That analysis is the hard part and you did it properly.

The remedy is where I have to stop, and it isn't a small objection.

The `content` tab key isn't private to those three files. It's the merge point that every page blueprint in the wild extends. Across my own workspace, 317 blueprint files declare a `content` container at the tabs level, 261 of them in first-party Grav repos, including Quark's own `blueprints/modular/text.yaml`.

I applied your patch to a sandbox, pointed it at stock Quark 2, and loaded a `modular/text` page:

```
items[content].type = ''
tabs children = content_tab(tab) | options(tab) | advanced(tab) | content(type=-,title=-)
```

The theme's `content:` override no longer merges into the renamed tab, so it becomes an untyped sibling, and it takes the same slot back. On a default install the bug isn't fixed, it moves, and the theme's own fields end up outside every tab. Admin Next drops untitled containers outright, and the classic tabs template renders them as an unreachable pane with the nav indices out of step.

So renaming is off the table, here and in general. That key is long-standing, deeply ingrained structure, and anything extending the base blueprint depends on it. Any fix for this has to be invisible to existing blueprints.

Where I think the real answer is: one layer down. Containers and their children share a single flat namespace in `items`, because `getFieldKey()` only extends the prefix for dotted keys, and `parseFormField()` writes the container's own entry *after* recursing into its children. So a container always wins a name collision with one of its own leaves. Writing that entry *before* the recursion makes the nested leaf win instead, needs no blueprint edits at all, and fixes third-party blueprints too, which a rename never could. In a sandbox that gives `items['content']` resolving to the markdown field with its validate config on default, modular and Quark's `modular/text`, every tab key unchanged, and the core unit suite still green.

That's a change to `rockettheme/toolbox` rather than to Grav, and it changes how every blueprint in every Grav site is parsed, so I want to evaluate it properly rather than take it on the strength of one sandbox run. I'll pick that up separately and report back here.

Leaving this open in the meantime, but I'm not going to merge the rename. Thanks for the depth on this one. Your test is well placed and I expect to want it once the safe fix lands.

rhukster:
Fix is in. `rockettheme/toolbox@43b6421` on `v2`, which is the approach I described above: the container's `items` entry gets written before the recursion instead of after, so the nested leaf wins the key. Grav picks it up through the composer pin at `d0033a941` on develop and it ships in 2.0.25. No blueprint changes, so nothing that extends the base page blueprint moves.

I checked what it actually changes before taking it. Patched a scratch install, built the merged schema for every blueprint I could find across 1,377 files, and diffed the result against stock.

`nested` comes out byte identical every time, so the form still renders the same, field ordering is the same, defaults are the same. Only the flat `items` map changes, and that's what validation and filtering read.

46 keys collide in total, but most of those are `unset@` removals that never reach the parser. The ones that really change: `content` on core `default` and `modular`, and on the helios and typhoon page types, which all inherit the core field. Then `options`, `sync`, `view`, `speaker` and `print` in grav-plugin-presentation, which now resolve to the real fields instead of the sections and dropdowns sitting on top of them. Presentation's `options` was filtering down to `""` and throwing the saved array away, so that one was a bug in its own right.

Two things I wanted to rule out first, and you had already worked through both of them.

The 2048 character `typeText` cap can't fire on a leaf with no type. `parseFormField()` always sets `type`, to `''` if the field doesn't declare one, and `Validation::validate()` only falls back to `text` when `type` is missing entirely. `''` goes looking for a `type()` method that isn't there and validates nothing. The bare `content: {}` leaves I found in helios and typhoon aren't bare at runtime anyway, both extend the core page blueprint.

And you were right about `checkSafety()`. `validateArray()` sets `$checkXss` regardless of which rule it resolved, so page content is already XSS checked. I ran `<script>`, `onerror=` and a `javascript:` URI through the default blueprint on stock and patched builds and got the same rejection both times, with 3 KB and 100 KB bodies passing either way.

Your test is in at `tests/unit/Grav/Common/Data/PageBlueprintContentTest.php`, adapted for the no-rename version and credited to you. Worth having, because there's nothing on Grav's side stopping those two keys colliding again. 3 of its 4 cases fail against the old toolbox; the fourth checks the tab keeps its name, so it passes either way on purpose.

Closing this one in favour of the upstream fix. Your write-up is what pointed at the right layer, appreciate it.

PR #4292, opened 2026-09-09
Fix date2timestamp() misparsing non-string dates (#3812)

### LINKED ISSUE #3812: YYYY-MM-DD formatted date in header doesn't work but "YYYY-MM-DD" and DD-MM-YYYY do
I've read quite a bunch of issues related to date format and some of them relate to the issue I'm having, but I've seen no proper answer to the problem. Please feel welcome to correct me if I'm wrong.

More details on the issue are visible [here](https://discourse.getgrav.org/t/date-header-does-not-seem-to-be-taken-into-consideration/25123).

After more investigations, here is the situation:
with the following in `config/system.yaml`:
```
dateformat:
    default: Y-m-d
    short: Y-m-d
    long: Y-m-d_H:i
```

The following date headers will make the post display the appropriate date:
```
date: 06-01-2022
```
or
```
date: "2022-01-06"
```
but
```
date: 2022-01-06
```
won't, it will display today's date instead (or maybe the date the markdown file has been saved).

Seems like a bug to me. But maybe there is an explanation.
It's not a big deal, but not so straightforward to understand why something is not working for users, so may be worth addressing.

Apart from that, thanks a lot for grav, it's amazing :) :) :)

wakqasahmed:
Fixes #3812.

Turns out the reported behavior is worse than "falls back to today's date". An unquoted YAML date header like `date: 2022-01-06` gets parsed by the YAML component as an int timestamp, not a string, since it looks like an ISO date. `Utils::date2timestamp()` only handled string input: `DateTime::createFromFormat()` fails silently on the int (PHP coerces it to the string `"1641427200"`, which doesn't match the `Y-m-d` format), and the `strtotime()` fallback then misparses that same numeric string into a nonsense timestamp - I checked, it lands on the year 7200, not "now".

Quoting the date in frontmatter (`date: "2022-01-06"`) sidesteps it because then it really is a string, which matches what the two "working" formats in the issue have in common.

Fix is to short-circuit `date2timestamp()` for int/float and `DateTimeInterface` input before it reaches the string-based parsing branches. Added a test alongside the existing `testDate2timestamp()` covering an int timestamp, a `DateTime`, and a `DateTimeImmutable`. Ran the full `UtilsTest` suite locally, all 36 tests green.

### DIFF
--- a/system/src/Grav/Common/Utils.php
+++ b/system/src/Grav/Common/Utils.php
@@ -10,6 +10,7 @@
 namespace Grav\Common;
 
 use DateTime;
+use DateTimeInterface;
 use DateTimeZone;
 use Exception;
 use Grav\Common\Flex\Types\Pages\PageObject;
@@ -1465,13 +1466,24 @@ abstract class Utils
     /**
      * Get the timestamp of a date
      *
-     * @param string $date a String expressed in the system.pages.dateformat.default format, with fallback to a
-     *                     strtotime argument
+     * @param string|int|DateTimeInterface $date a String expressed in the system.pages.dateformat.default format,
+     *                     with fallback to a strtotime argument. An unquoted YAML date header (e.g. `date: 2022-01-06`)
+     *                     is parsed by the YAML component into an int timestamp or a DateTimeInterface rather than a
+     *                     string, so those are accepted directly instead of being coerced into strtotime(), which
+     *                     misparses a bare numeric string into a bogus date.
      * @param string|null $format a date format to use if possible
      * @return int the timestamp
      */
     public static function date2timestamp($date, $format = null)
     {
+        if ($date instanceof DateTimeInterface) {
+            return $date->getTimestamp();
+        }
+
+        if (is_int($date) || is_float($date)) {
+            return (int) $date;
+        }
+
         $config = Grav::instance()['config'];
         $dateformat = $format ?: $config->get('system.pages.dateformat.default');
 
--- a/system/src/Grav/Common/Utils.php
+++ b/system/src/Grav/Common/Utils.php
@@ -1466,11 +1466,10 @@ abstract class Utils
     /**
      * Get the timestamp of a date
      *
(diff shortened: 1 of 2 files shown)


rhukster:
Good catch, and the diagnosis is spot on. I reproduced it end to end before touching anything: an unquoted `date: 2022-01-06` comes out of the YAML parser as the int `1641427200`, `createFromFormat('Y-m-d')` fails on it silently, and `strtotime("1641427200")` lands the page in the year 7200. Not "falls back to today" at all, exactly as you said.

I pushed one commit on top rather than send it back to you, since it's a small thing.

The int branch caught a case that was already working. `date: 20220106` is also an int by the time it reaches `date2timestamp()`, but it's the number the author typed rather than a timestamp, and `strtotime()` has always read it correctly as a date. Returning every int unchanged moved those pages to August 1970:

| header | develop | your patch | now |
| --- | --- | --- | --- |
| `date: 2022-01-06` | 7200-09-09 | 2022-01-06 | 2022-01-06 |
| `date: 20220106` | 2022-01-06 | 1970-08-23 | 2022-01-06 |
| `date: "2022-01-06"` | 2022-01-06 | 2022-01-06 | 2022-01-06 |
| `date: 2022-01-06 10:30` | 2022-01-06 10:30 | 2022-01-06 10:30 | 2022-01-06 10:30 |

So the int short circuit now only fires for a number that can't be read as `Ymd`; anything that can falls through to the parsing that already handled it. Your `DateTimeInterface` branch is untouched, it was right.

Added `testDate2timestampKeepsReadingBareNumericDates` to hold that down, and a float case to the test you wrote. It fails against the first version of the patch and passes now, which is the point of having it. Full suite is green, and CI passed on 8.3, 8.4 and 8.5.

Worth saying that `date: 2022` is still nonsense either way, today at 20:22 before and 1970 now. It's a malformed header rather than something to guess at, so I left it alone.

Merging. Thanks for chasing this one down to the actual cause instead of stopping at the symptom in the issue.

PR #4293, opened 2026-09-10
Allow overriding html/json/xml/atom/rss content types via media.types (#3735)

### LINKED ISSUE #3735: Allow for overriding content-type of html, rss, atom, etc
In the code https://github.com/getgrav/grav/blob/490bdd6ce7aaae315c033ca5af112a778a42455a/system/src/Grav/Common/Utils.php#L800, Grav hardcodes the content type for some extensions. I want to have a different content type returned for atom/rss in order to style it with xslt.

wakqasahmed:
Closes #3735.

`Utils::getMimeByExtension()` hardcoded the mime type for `html`, `json`, `xml`, `atom` and `rss` in a switch statement, checked before the `media.types` config was ever consulted. Every other extension already falls through to `media.types`, so a site could override, say, `doc`'s mime type, but not `html`'s or `xml`'s — even though both already exist as regular entries in `media.yaml` with the exact same value the switch hardcoded.

Flipped the order: `media.types` is checked first now, and the switch is just the fallback for `atom`/`rss`, which weren't registered as media types at all before this. Added them to `media.yaml` following the existing `xml`/`json`/`html` pattern (type: file, a thumb icon, the same mime values as before), so they're overridable too.

No behavior change for anyone not overriding these five in `media.types` - the switch's values and the config's values are identical today. Added a test covering the override case (`media.types.rss.mime` in config actually takes effect), alongside the existing `testGetMimeByExtension` which still passes unchanged.

### DIFF
--- a/system/src/Grav/Common/Utils.php
+++ b/system/src/Grav/Common/Utils.php
@@ -932,10 +932,19 @@ abstract class Utils
     {
         $extension = strtolower($extension);
 
-        // look for some standard types
+        if ($extension === '') {
+            return $default;
+        }
+
+        // media.types (which a site can override) takes precedence over the built-in defaults
+        $media_types = Grav::instance()['config']->get('media.types');
+        $mime = $media_types[$extension]['mime'] ?? null;
+        if ($mime) {
+            return $mime;
+        }
+
+        // fallback for a few standard types not present in media.types
         switch ($extension) {
-            case null:
-                return $default;
             case 'json':
                 return 'application/json';
             case 'html':
@@ -948,9 +957,7 @@ abstract class Utils
                 return 'application/xml';
         }
 
-        $media_types = Grav::instance()['config']->get('media.types');
-
-        return $media_types[$extension]['mime'] ?? $default;
+        return $default;
     }
 
     /**
--- a/system/src/Grav/Common/Utils.php
+++ b/system/src/Grav/Common/Utils.php
@@ -943,7 +943,8 @@ abstract class Utils
             return $mime;
         }
 
-        // fallback for a few standard types not present in media.types
+        // fallback if media.types is missing one of these standard entries entirely
+        // (e.g. a site config that replaces media.yaml wholesale instead of merging)
         switch ($extension) {
             case 'json':
                 return 'application/json';
--- a/tests/unit/Grav/Common/UtilsTest.php
+++ b/tests/unit/Grav/Common/UtilsTest.php
@@ -221,6 +221,15 @@ class UtilsTest extends \PHPUnit\Framework\TestCase
         self::assertEquals('text/html', Utils::getMimeByExtension('foo', 'text/html'));
     }
 
+    public function testGetMimeByExtensionHonorsMediaTypesOverride(): void
+    {
+        $this->grav['config']->set('media.types.rss.mime', 'application/xml; charset=utf-8');
+
+        self::assertEquals('application/xml; charset=utf-8', Utils::getMimeByExtension('rss'));
+        // untouched types are unaffected
+        self::assertEquals('application/atom+xml', Utils::getMimeByExtension('atom'));
+    }
+
     public function testGetExtensionByMime(): void
     {
         self::assertEquals('html', Utils::getExtensionByMime('*/*'));
--- a/system/config/media.yaml
+++ b/system/config/media.yaml
@@ -104,6 +104,14 @@ types:
     type: file
     thumb: media/thumb-xml.png
     mime: application/xml
+  atom:
+    type: file
+    thumb: media/thumb-xml.png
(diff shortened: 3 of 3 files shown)


system/src/Grav/Common/Utils.php
```diff
@@ -932,10 +932,19 @@ public static function getMimeByExtension($extension, $default = 'application/oc
...
+        }
+
+        // media.types (which a site can override) takes precedence over the built-in defaults
+        $media_types = Grav::instance()['config']->get('media.types');
+        $mime = $media_types[$extension]['mime'] ?? null;
+        if ($mime) {
+            return $mime;
+        }
+
+        // fallback for a few standard types not present in media.types
```
wakqasahmed: Minor: this comment is now inaccurate for all five cases, not just atom/rss. `html`/`json`/`xml` were already in `media.yaml` before this PR, and `atom`/`rss` are added to `media.yaml` by this same PR. So on any default/unmodified install, `media.types` will always have an entry for all five, and this switch never actually executes — it only fires if a site explicitly strips one of these keys from `media.types`. Not wrong, just worth a comment update (e.g. "fallback if a site removes these from media.types") so the next person reading this doesn't assume atom/rss are still missing from the yaml, which was true before this diff but isn't anymore.

tests/unit/Grav/Common/UtilsTest.php
```diff
@@ -221,6 +221,15 @@ public function testGetMimeByExtension(): void
         self::assertEquals('text/html', Utils::getMimeByExtension('foo', 'text/html'));
     }
 
+    public function testGetMimeByExtensionHonorsMediaTypesOverride(): void
+    {
+        $this->grav['config']->set('media.types.rss.mime', 'application/xml; charset=utf-8');
```
wakqasahmed, line 226: This mutates `Grav::instance()['config']` (a singleton) via `$this->grav['config']->set(...)` and there's no `tearDown` to reset `media.types.rss.mime` afterward. It happens to be safe today because this test runs after `testGetMimeByExtension` in declaration order and nothing else in the suite depends on the default rss mime later, but it's a latent test-isolation issue — if execution order ever changes (e.g. `--order-by=random`) or another test starts asserting the default rss mime, this leaks state across tests silently. Might be worth resetting the value (or using a scoped config clone) in case order-independence is ever enforced.

wakqasahmed (commented):
Went through this cold against the linked issue (#3735) and the base commit. Confirmed the reordering is behavior-preserving for a default config: `html`/`json`/`xml` in `media.yaml` already resolve to `text/html`/`application/json`/`application/xml`, exactly matching what the old switch hardcoded, so nobody who hasn't touched `media.types` sees a change. The `$extension === ''` check is also a faithful replacement for the old `case null:` — after `strtolower()` the value can never actually be `null`, and `null == $str` was only ever true for `$str === ''`, so the trigger set is identical, not just similar. Config access at the existing call sites (Page::header, PageLegacyTrait, ControllerResponseTrait, etc.) isn't a new risk either — the old code already called `Grav::instance()['config']` unconditionally for every extension outside the hardcoded five, so this doesn't introduce a new ordering dependency, just makes it uniform. New test directly exercises the override path and leaves the existing `testGetMimeByExtension` untouched. Two minor things left as inline comments, neither blocking.

system/src/Grav/Common/Utils.php
wakqasahmed: Fixed — reworded to describe it as a fallback for these entries being missing from media.types entirely (e.g. a site that replaces media.yaml wholesale), rather than implying they're never present.

tests/unit/Grav/Common/UtilsTest.php
wakqasahmed: Checked this — no change needed. tests/_bootstrap.php's grav fixture calls Grav::resetInstance() inside the closure Fixtures::get('grav') returns, and UtilsTest::setUp() calls that closure fresh before every single test method (unconditional on order, since setUp runs before each test, not once per class). So there's no cross-test leakage risk tied to ordering here — every test already gets a brand-new Grav instance regardless of what ran before it.

rhukster:
Thanks for this, and for the clear write-up and the test.

It turns out we crossed paths: the same change landed on develop a few hours after you opened this, in 0b22b23ed, as part of a batch of 2.1 fixes. It works the way yours does. `media.types.<ext>.mime` is checked first and the built-in switch is only the fallback, with a test for the RSS override. Because the fallback still covers `atom` and `rss`, it didn't need the extra `media.yaml` entries, so I'm closing this one as superseded rather than merging a second copy of the same fix.

Your diagnosis was spot on, and I've added a credit to the 2.1.0 changelog. Thanks for picking up such an old issue.

PR #4296, opened 2026-09-11
Fix custom_base_url matching pages by literal prefix instead of path segment

### LINKED ISSUE #3057: Pages whose name starts with `custom_base_url` are not found.
I've a setup relying on `custom_base_url`:
- https://mysite.com is WordPress
- https://mysite.com/act/* is Grav
- Apache side: `RewriteRule ^/act/(.+)$ http://127.0.0.4:8000/$1 [P]` (Reverse-proxy to that that local IP vhost Grav runs on)
- On Grav side, `config/system.yaml` contains `custom_base_url: /act` and `absolute_urls: false`

This works well for any page like https://mysite.com/act/foobar.

But I discovered it failed for **act**/**act**ion-bar

In `Uri::init`, the `$root_path == /act` is extracted from `custom_base_url` and `$this->root` end-up being `https://mysite.com/act` so far so good.
In the meantime, `$this->url` is built from `$this->base` (`http://mysite.com`) and `$this->uri` (`/action-bar`)

And when `replaceFirstOccurrence` (`mb_strpos` intend to replace `http://mysite.com/act` by "" and subject (`this->url`) is `http://mysite.com/action-bar`.

The page looked for is then... `/ion-bar` what results in a 404.

wakqasahmed:
`Uri::init()` strips the custom base from the request URL with a raw substring search (`Utils::replaceFirstOccurrence`), no path-segment boundary check. So a `custom_base_url` of `/act` matches the start of `/action-bar` as plain text, and the strip turns it into `ion-bar` — any page whose name happens to start with the same letters as the base 404s.

`InitializeProcessor::pathHasBase()` already does the correct, boundary-aware comparison for a related case (added for #3822), with a docblock calling out this exact failure mode by name — this is the surviving instance of the same bug in `Uri::init()`.

Fix checks that the root is followed by `/` (or matches exactly) before stripping it. When a request doesn't genuinely fall under the custom base, it falls back to stripping just the site's own base instead — the same result as if no custom_base_url were configured for that request, which is what makes the page findable again rather than leaving the host/port in the path.

Added a regression test alongside the existing `testCustomBase()` in `UriTest.php`.

Fixes #3057

### DIFF
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -193,7 +193,13 @@ class Uri implements \Stringable
 
         $this->url = $this->base . $this->uri;
 
-        $uri = Utils::replaceFirstOccurrence(static::filterPath($this->root), '', $this->url);
+        $root = static::filterPath($this->root);
+        $root_prefix = Utils::endsWith($root, '/') ? $root : $root . '/';
+        if ($this->url === $root || Utils::startsWith($this->url, $root_prefix)) {
+            $uri = Utils::replaceFirstOccurrence($root, '', $this->url);
+        } else {
+            $uri = Utils::replaceFirstOccurrence($this->base, '', $this->url);
+        }
 
         // remove the setup.php based base if set:
         $setup_base = $grav['pages']->base();
--- a/system/src/Grav/Common/Uri.php
+++ b/system/src/Grav/Common/Uri.php
@@ -193,10 +193,14 @@ class Uri implements \Stringable
 
         $this->url = $this->base . $this->uri;
 
-        $root = static::filterPath($this->root);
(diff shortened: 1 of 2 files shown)


rhukster:
Thanks @wakqasahmed, good catch, and the diagnosis was spot on.

There was one problem with the boundary check. `$this->url` includes the query string, so a root followed by `?` didn't count as a boundary. On a subfolder install, `https://example.com/sub?x=1` would have gone to a page called `/sub` instead of the home page, even with no custom base set. And with a custom base of `/act`, `/act?page=2` did the same.

I pushed a follow-up commit to your branch that treats `?` and `#` as boundaries too, plus a set of tests covering the home page with a query string, pages under the base, the prefix collision itself and a request where a proxy has already removed the base.

Merged for 2.1.0. Thanks again!

PR #4302, opened 2026-09-13
Bypass js_pipeline for the Clockwork debugger script (#3871)

### LINKED ISSUE #3871: clockwork-badge does not show up, if `assets.js_pipeline: true`
If `assets.js_pipeline: true` in `user/config/system.yaml`, the „Clockwork badge“ (bottom left corner) does not show up.

https://github.com/getgrav/grav/blob/a1e583f657f28305e5cbaa4c6b73aa70e1706d62/system/assets/debugger/clockwork.js#L4-L9

checks for `#clockwork-script`, which does not exist, when all scripts are pipelined.

Error in JavaScript console:

```
Clockwork Debugger: Script tag with id 'clockwork-script' not found.
```

wakqasahmed:
When `assets.js_pipeline` is enabled, `clockwork.js` gets swept into the combined pipelined `<script>` tag along with every other JS asset. The pipeline merge only carries over the `loading` attribute across assets, so the `id="clockwork-script"` and `data-route` attributes this particular script depends on get dropped once it's merged. The client script does `document.getElementById('clockwork-script')` and just logs a console error and bails when that id isn't there, so the Clockwork badge silently never shows up with `js_pipeline: true` - reproduces exactly as described in the issue.

The fix is small: pass `'pipeline' => false` alongside the existing `id`/`data-route` options when `Debugger::addAssets()` registers `clockwork.js`, so this one script always renders as its own standalone tag no matter the pipeline setting. Nothing else about pipelining changes.

Added a regression test in `AssetsTest` that turns `js_pipeline` on, adds another JS asset so there's something to actually pipeline, then runs the real `Debugger::addAssets()` + `Assets::js()` path and asserts the clockwork script still comes out as its own tag with `id="clockwork-script"` and the right `data-route`, while the other asset gets pipelined normally. Confirmed it fails against the current code (asset gets merged, no id) and passes after the fix.

Closes #3871

### DIFF
--- a/system/src/Grav/Common/Debugger.php
+++ b/system/src/Grav/Common/Debugger.php
@@ -667,7 +667,8 @@ class Debugger
                 $assets->addCss('/system/assets/debugger/clockwork.css');
                 $assets->addJs('/system/assets/debugger/clockwork.js', [
                     'id' => 'clockwork-script',
-                    'data-route' => $route
+                    'data-route' => $route,
+                    'pipeline' => false
                 ]);
             }
 
--- a/tests/unit/Grav/Common/AssetsTest.php
+++ b/tests/unit/Grav/Common/AssetsTest.php
@@ -3,6 +3,7 @@
 use Codeception\Util\Fixtures;
 use Grav\Common\Grav;
 use Grav\Common\Assets;
+use Grav\Common\Page\Page;
 
 /**
  * Class AssetsTest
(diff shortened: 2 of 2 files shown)


rhukster:
Thanks — this is exactly the right fix. The pipeline merge only carries `loading` across, so `pipeline: false` is the correct opt-out for a script that depends on its own attributes, and the regression test is a nice touch.

PR #4306, opened 2026-09-14
Register theme blueprints regardless of blueprints/pages (#4303)

### LINKED ISSUE #4303: Theme blueprints are only registered when the theme also ships blueprints/pages
### Summary

`Themes::initTheme()` registers a theme's whole `blueprints://` path only when the theme happens to have a `blueprints/pages` folder:

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Themes.php#L96-L101

```php
// Register blueprints.
if (is_dir('theme://blueprints/pages')) {
    $locator = $this->grav['locator'];
    $locator->addPath('blueprints', '', ['theme://blueprints'], ['user', 'blueprints']);
}
```

The guard names `blueprints/pages`, but what it controls is the registration of `theme://blueprints` as a whole. So a theme that ships blueprints of any other kind — most obviously `blueprints/flex-objects/<type>.yaml`, but equally `blueprints/user/`, `blueprints/config/` or a custom namespace — never gets them on the stream at all, unless it also happens to ship page blueprints it may have no use for.

### Impact

Nothing is logged and nothing errors; the blueprint is simply never found. For Flex this reads as a directory that silently does not exist: the frontend renders an empty collection and Admin Next omits the type entirely. Both symptoms misdirect toward data or permissions.

Surfaced while fixing trilbymedia/grav-plugin-flex-objects#240, which is a *different* bug in the same area (that one was a plugin-side re-registration gated on `isAdmin()`, fixed in flex-objects 1.4.15). This core gap is the remaining half, and no plugin-side workaround can reach it — the path is never added, so there is nothing to re-scan.

### Steps to reproduce

1. Take any theme with no `blueprints/pages` directory.
2. Add `blueprints/flex-objects/widget.yaml` to it, and enable `widget` under the Flex Objects plugin's Directories setting.
3. The type never registers. Adding an empty `blueprints/pages/` folder to the same theme makes it work.

### Suggested fix

Gate on the directory actually being registered:

```php
// Register blueprints.
if (is_dir('theme://blueprints')) {
    $locator = $this->grav['locator'];
    $locator->addPath('blueprints', '', ['theme://blueprints'], ['user', 'blueprints']);
}
```

Themes that ship `blueprints/pages` are unaffected — they already satisfy the wider condition. The change only adds the path for themes that have blueprints but no page blueprints, which today get nothing.

Worth a check before merging: the path is added with `['user', 'blueprints']` as the base, so a theme could now shadow core blueprints it previously could not reach. That was already true for any theme shipping `blueprints/pages`, so this widens an existing capability rather than introducing one, but it is the thing to look at.

wakqasahmed:
Fixes #4303.

**Edit:** my original approach here (widen the `is_dir()` check to `theme://blueprints`, mounting the whole folder at the `blueprints://` root) was wrong for the same reason @rhukster's own fix attempt at 8488c81d7 got reverted — it lets a theme's unnamespaced, root-level content (Quark's `blueprints/default.yaml`, which self-extends via `extends@: default`) shadow the stream root and recurse into itself. Replaced with the approach from his revert comment: register each immediate subfolder of `blueprints/` individually instead of mounting the parent tree.

---

`initTheme()` only registered the theme's `blueprints://` stream when the theme had a `blueprints/pages` folder. A theme shipping only `blueprints/flex-objects/*.yaml` (or `blueprints/user/`, etc.) never got anything registered — nothing errors, the blueprint just silently never resolves.

Now each immediate subfolder of the theme's `blueprints/` (`pages`, `flex-objects`, `user`, `config`, ...) gets mounted individually at its own `blueprints://<namespace>` prefix, rather than mounting the whole `blueprints/` folder at the stream root. A theme's page blueprints resolve at the exact same `blueprints://pages/...` URIs as before. Anything sitting directly under `blueprints/` with no subfolder (Quark's case) is never mounted anywhere, so it can't collide with the stream root — same as before this fix, no regression.

Tests: the original two (non-pages namespace resolves, pages namespace still works) plus a new one for the Quark-shaped case — a root-level self-extending blueprint must not resolve through `blueprints://` at all.


sridharkalaibala:
This guard change matches the one reverted in [e3d5d491d](https://github.com/getgrav/grav/commit/e3d5d491d7aad3b48d3669335505a6d0ece0236b). The maintainer explains the regression in [#4303](https://github.com/getgrav/grav/issues/4303#issuecomment-5658420087): mounting the whole blueprint tree lets a theme's root `default.yaml` shadow the system default, so Quark's `extends@: default` resolves back to itself.

The two locator tests cover successful lookup, but neither includes that root-level page blueprint or exercises inheritance. Please preserve the system default lookup and test that case when revising this toward individual namespace registration. This is based on the current diff and the maintainer's verified revert; I have not independently rerun the out-of-memory failure.

wakqasahmed:
Good catch — and it turned out to be the exact same bug the maintainer's own attempt at this hit and reverted (8488c81d7 → e3d5d491d), which I found independently around the same time you commented.

Switched to registering each `blueprints/` subfolder individually at its own `blueprints://<namespace>` prefix instead of mounting the whole tree, so unnamespaced root-level content (Quark's case) never gets mounted anywhere and can't collide. Added the regression test for that, plus one for the case you flagged: a theme's own `pages/` blueprint doing `extends@: default` now goes through `Blueprint::load()->init()` and asserts it actually inherits the system default's fields (`form/fields/tabs/fields/content`) rather than losing them or recursing — `blueprints://pages/default.yaml` is a different, already-supported override slot than the root-level Quark case, so that inheritance chain was never broken by the mount-mode itself, but good to have it pinned down explicitly now.

rhukster:
This is the right approach now. I ran the Quark case against all three versions: on `8488c81d7` `blueprints://default.yaml` resolves to the theme's self-extending file and building a default blueprint dies at 256MB, and on this branch it's back to `false`, with the rendered page blueprints byte-identical to develop (243053 bytes across all seven Quark page types). Your regression test for the root-level case is the only one of the five that fails on `8488c81d7`, so it genuinely pins the thing that got reverted.

One thing to fix first. Registering the namespace as its own prefix puts the theme *ahead* of `environment://blueprints`, which used to win — a namespace prefix is a fresh bucket, so it sorts ahead of everything in the `''` prefix where these lookups used to land. The `['user', 'blueprints']` override argument you carried over is dead code, since `addPath()` only consults it when the prefix already exists.

Pass `environment://` first in the path list, and put a trailing slash on the prefix so `pages` can't also match `pages-legacy`:

```php
$locator->addPath('blueprints', "{$namespace}/", [
    "environment://blueprints/{$namespace}",
    "theme://blueprints/{$namespace}",
]);
```

That restores the original deduped order exactly. Worth knowing while you're in there: none of the 48 first-party themes ship a `blueprints/pages` folder, so develop's guard is dead for every theme we ship — which is why this went unnoticed for so long.

Add a bugfix bullet under the 2.1.6 block too and I'll get this in.

rhukster:
Merging this, with the ordering fix applied on top so you don't have to do another round trip.

Two corrections to my own review while I'm here.

**The `environment://` point was right, and I've now proved it rather than asserted it.** A small locator harness, same registration shape as core, one file in each of `environment://blueprints/pages` and `theme://blueprints/pages`:

```
PR as submitted     blueprints://pages/default.yaml resolves to: theme
with the fix        blueprints://pages/default.yaml resolves to: env
```

A namespace prefix is its own bucket, `krsort` puts it ahead of the `''` prefix these lookups used to land in, and `addPath()`'s `$override` argument is only consulted inside `if (isset($this->schemes[$scheme][$prefix]))` — so on a brand-new prefix the `['user', 'blueprints']` you carried over was genuinely dead. Listing `environment://` first in the path array restores the old precedence.

**The trailing-slash point was weaker than I made it sound.** I said `pages` could match `pages-legacy`; the prefix test at `UniformResourceLocator::find()` really is a bare `strpos($file, $prefix) !== 0`, so it does match — but the lookup then falls through to the `''` prefix and resolves correctly anyway, and I could not construct a case where it actually mis-resolves. I've kept the trailing slash because it says what we mean, not because it fixes a live bug. Sorry for overstating that one.

Verified before merge, on a real Quark2 site rather than only in the harness:

- `blueprints://default.yaml` stays `false`, so the root-shadowing case that got `8488c81d7` reverted does not come back.
- Built page blueprints are byte-identical to develop across `default`, `blog`, `item`, `modular` and `modular/features`, stable over repeated runs with the cache cleared between them. (I briefly thought `item` had changed; that was a stale cache between my own checkouts, not your change.)
- The whole unit suite passes, 1301 tests.
- Your five tests pass with the ordering fix in place.

One thing worth recording for anyone who finds this later: of the first-party themes I have checked out, **not one** with a `blueprints/` folder has a `blueprints/pages` subfolder. The namespaces they actually ship are `modular`, `partials` and `asset`, none of which collide with core's `config`, `flex`, `pages` or `user`. So develop's guard registered nothing for any theme we ship, which is why this sat unnoticed.

I'll write the changelog entry, since 2.1.6 went out a few hours ago and this now lands in 2.1.7. Thanks for turning this around so quickly after the revert, and for the regression test — it's the only one of the five that fails on the reverted commit, which is exactly what it needed to do.

PR #4307, opened 2026-09-14
Isolate per-asset CSS minification failures (#4305)

### LINKED ISSUE #4305: CSS minifier breaks on modern CSS: `calc()` spacing and `hsl(h s% l%)`
Hi Andy,

Grav 2.1.2's CSS pipeline minifies through `tubalmartin/cssmin` 4.1.1 (`Assets/Pipeline.php`), whose last commit is from January 2018. Two things it does to current CSS:

- It strips the spaces around `+` and `-` inside `calc()`, which browsers then discard as invalid, so the rule silently disappears.
- It throws on space-separated colour syntax such as `hsl(210 40% 50%)`.

The pipeline has no per-file isolation, so one throw 500s every page. Same family as #3095 (2020), whose answer was to keep the file out of the pipeline.

Ask: a maintained minifier, or failing that, catch a minifier throw and pass that one asset through unminified. Until then `assets.css_minify: false` costs about 2 KB a page.

Cheers,
Trev

wakqasahmed:
Fixes #4305.

`renderCss()` concatenated every pipelined CSS asset into one buffer and minified the whole thing in a single `cssmin` call. `tubalmartin/cssmin` hasn't been touched since 2018 and doesn't handle space-separated `hsl(210 40% 50%)` color syntax — it splits the color's `explode(',', ...)` on a string that has no commas, ends up reading undefined array offsets, and under Grav's own Whoops-based error handler that turns into a thrown `ErrorException`. With no per-file isolation, one asset like that takes the whole page down.

The JS side of this pipeline already solved the same problem (`gatherAndMinifyJs`/`renderJs`): minify each file individually, exclude anything that throws from the combined buffer and the cache key, and let the caller render the failed one on its own. `Assets::render()` already handles that `{output, failed}` return shape generically — it's not JS-specific, it was just never fed by the CSS side. This mirrors that pattern for CSS (new `gatherAndMinifyCss()`), catching `\Throwable` rather than `\Exception` since the actual failure here arrives via PHP's error handler, not a library-thrown exception. Left the non-minify path and the JS pipeline untouched.

This addresses the crash. It doesn't fix the separate `calc()` spacing issue the same library has (stripping the spaces silently rather than throwing) — that's not something a try/catch can help with, it needs a different minifier or a preserve-whitespace pass, and felt out of scope for this PR.

Added two fixtures under `tests/unit/data/assets/` (one with the broken `hsl()`/`calc()` syntax, one plain) and a test asserting the pipeline no longer throws and the broken asset still renders (unminified, on its own) instead of disappearing.


sridharkalaibala:
The fallback changes stylesheet order, which can change the cascade even though all assets remain present. I reproduced this using this PR's exact Pipeline.php at a6453f0 in a Grav 2.1.2 runtime with warnings converted to ErrorException, as in the reported failure:

1. First asset: `.example { color: hsl(210 40% 50%); }`
2. Second asset: `.example { color: blue; }`

The output is the successful blue stylesheet in the combined bundle, followed by the failed first asset as a separate link. Feeding those actual emitted stylesheet contents to Chromium changes the computed color from the intended `rgb(0, 0, 255)` to `rgb(77, 128, 179)`.

`Assets::render()` appends failed assets after the entire successful bundle. The existing regression checks that the assets are present, but cannot catch this ordering change. Please preserve the original order when a file fails—such as by emitting ordered bundle segments around failed assets, or falling back for the whole affected group—and add an order/cascade regression. No live site or external asset requests were used in the reproduction.

wakqasahmed:
Good catch, that's a real gap. Bundling the successful assets and rendering the failed one separately afterward would put the failed asset's rule after everything in the bundle regardless of where it actually sits in the pipeline group, changing which rule wins the cascade.

Went with your second option: when any asset in the group fails to minify, the whole group falls back to individual, unminified `<link>` tags in original order rather than a partial bundle plus a reordered failure. Updated the test to assert the exact expected output (both links, in original order) instead of just checking presence.

rhukster:
Thanks for turning this around so fast, and @sridharkalaibala, thanks for catching the cascade issue. I pulled the branch and ran it.

Verified: the order regression is gone. With the broken asset first and a valid one second, the output is both links in the original order, so the reproduction no longer changes the computed colour. Full unit suite green on the branch.

Two things before this goes in.

**The fallback is more expensive than it needs to be.** Because the buffer is empty on the failure path, nothing gets written to the assets cache — so every page view on an affected site re-reads all the CSS, re-runs cssmin over all of it, throws the result away, and writes another line to `grav.log`. That's worse than just setting `css_minify: false`, which at least caches the concatenated bundle.

The cheaper way to preserve order is to keep the bundle and skip the minify:

```php
$buffer = $this->gatherLinks($assets, self::CSS_ASSET);
```

`gatherLinks()` already concatenates in original order and already hoists `@import`, so that's exactly the output `css_minify: false` produces for those files. You keep one `<link>` instead of N, the bundle gets cached, the failing minify runs once per asset change instead of once per request, and the group attributes passed to `css()` survive — right now they're dropped, since `Css::render()` only knows the asset's own attributes.

**That also lets `renderCss()` keep returning `string|false`.** As written the PR changes a public method's return type to `array|string|false`, which hands "Array to string conversion" to any plugin calling it directly. `renderJs()` already does that, but there's no reason to spread it. With the `gatherLinks()` fallback there are never any failed assets to hand back, so `$failedAssets`, the empty-output branch and the array return can all come out — a smaller diff than what's there now.

Worth noting for anyone who objects to per-asset minification on performance grounds: it isn't slower. Over six real theme stylesheets, single-pass was 174884 bytes in 26.1ms and per-asset was 174886 bytes in 13.7ms. Two bytes bigger, twice as fast, since cssmin's regex passes scale badly with input size.

Everything else I'd keep as-is. `catch (\Throwable)` is the right call over `\Exception`, it wraps only the `run()` call, OOM is a fatal and still gets through, and logging to both the debugger and `grav.log` means a site owner can find out why their CSS stopped minifying. The JS path needs that same `\Throwable` and has the same reordering problem — I've filed that separately so it doesn't hold this up.

On the `calc()` half you left out: good call, and it's narrower than the issue suggests. I tested 36 variants and cssmin doesn't touch `calc()` spacing inside rule bodies at all — only in at-rule preludes, and only for `+`. I'm replacing the minifier rather than patching around it, which is a 2.2 change. This PR's isolation is worth having either way.

Make those two changes and I'll merge it.

wakqasahmed:
Both changes made in 3aadfa4.

**Fallback now bundles instead of re-minifying every request.** `renderCss()`'s failure branch is now `$buffer = $this->gatherLinks($assets, self::CSS_ASSET);` (system/src/Grav/Common/Assets/Pipeline.php:158) instead of leaving the buffer empty and pushing the whole group into `$failedAssets`. That's the same concatenation `css_minify: false` produces, gets written to the assets cache like any other bundle, and renders as the one `<link>` built further down — no separate per-asset links or repeated minify-and-discard on every request.

**`renderCss()` is back to `string|false`.** Removed `$failedAssets`, the `empty($failedAssets)` output branch, and the `array{output, failed}` return — it just returns `$output` now (Pipeline.php:177). `renderJs()` is untouched.

Updated `testCssMinificationFailureFallsBackForWholeGroup` to match: it now asserts a single cached `<link href="/assets/...">`, reads that file back, and checks both the broken and valid rule bodies are present with the broken one still ahead of the valid one in the bundle (order preserved, nothing lost).

Ran the full AssetsTest suite plus the full unit suite via docker (php:8.3, matching composer.json's `^8.3`) — AssetsTest: 30/30 green; full unit suite: 1298 tests, 9 pre-existing failures unrelated to this change (image/GD fallback rendering in ParsedownTest/ExcerptsTest, not touched by this PR).

sridharkalaibala:
Thanks — I checked `3aadfa4` against the ordering case I originally reported. The fallback now writes one cached, unminified bundle with the broken asset before the valid asset, so the cascade order is preserved without changing `renderCss()`'s return type. The hosted PHP 8.3, 8.4, and 8.5 unit jobs are also green.

rhukster:
Thanks for the quick turnaround on both changes. I checked 3aadfa4: a failed minify now falls back to one cached, unminified bundle in the original order, the group attributes come through, it only retries when the assets change, and `renderCss()` is back to returning a string. Tests pass here too. Merged, thanks again, and thanks @sridharkalaibala for the careful review.

PR #4308, opened 2026-09-14
Fix division by zero for number fields with step any

### LINKED ISSUE #3585: Data validation fails for some numbers
This is related to issue #3433.

To replicate the error: Create a number field with validate.step of .0000001 and validate.min -90. Try inputting value 81.96 on the admin panel blueprint.

Expectation: The blueprint saves.

Result: The blueprint fails to save, indicating invalid input in that field.

A couple of notes:

- Changing the step to .000001, removing the min, or changing the input value to 81.9 or 81.9692 (as a few examples) all provide the expected result. Changing the step .00001, however, still fails validation. This is consistent with issues with the PHP fmod function.
- The fmod function used to be used in Grav/Common/Data/Validation.php in the typeNumber function (and did, in fact, cause the exact problems expected). It is no longer used in that function, however, so I think the error must be introduced elsewhere.

sridharkalaibala:
A number field with `validate.step: any` accepts arbitrary decimal steps in HTML, but server-side `Validation::typeNumber()` casts `any` to zero and divides by it. For example, value `0.12345` with min 0/max 1 raises `DivisionByZeroError` instead of validating.

Skip the step-grid calculation for the explicit `any` keyword. Numeric validation and min/max checks run first; numeric step values retain their existing validation behavior.

Related to #3585's request for a way to opt out of step restrictions. This does not fix the separate floating-point alignment issue for numeric steps and should not close #3585.

Validation on PHP 8.3.6 using PHPUnit 9.6.36: the new tests produce two division-by-zero errors before the change, then pass all 3 tests/10 assertions after it. They cover arbitrary decimals, numeric/range rejection, boundary values, unchanged numeric step-grid behavior and an omitted step. PHP syntax and `git diff --check` pass. Tests use the actual current core Validation class; the full Codeception suite and Admin/browser save flow were not run. Based on develop d581c327c. Prepared with AI assistance.

Follow-up verification: the [HTML standard](https://html.spec.whatwg.org/multipage/input.html#the-step-attribute) defines `any` as ASCII case-insensitive. Updated the comparison accordingly and added an uppercase/mixed-case regression (it raised DivisionByZeroError before the follow-up). Final focused suite: 4 tests, 12 assertions pass. Chromium confirms arbitrary decimal acceptance and both range rejections for `any`, `ANY` and `Any` (9 cases). An actual Grav Blueprint/Data validate/filter/save probe preserves numeric 0.12345 in synthetic YAML for all three spellings. These are separate browser-control and backend persistence checks, not an authenticated Admin HTTP round trip.


rhukster:
Confirmed the crash on develop: your three new tests raise `DivisionByZeroError` at `Validation.php:647` before the change and pass after. The `strcasecmp` comparison is right and matches how the HTML spec defines `any`, min/max/numeric all still run underneath it, and the test file matches the house style.

I've taken this into 2.1.6, with two extensions on top rather than sending it back.

`any` turned out to be one of several ways that line divides by zero — `step: 0`, `'0'`, `''`, `false` and any non-numeric step all still 500 a blueprint save, and so does `typeRange`, which delegates here. The spec's own rule is that anything which doesn't parse as a positive number falls back to the default step instead of erroring, so guarding on `$step > 0` after the cast covers the whole set. There's also a sibling in `typeArray()` at line 917, which `typeSelect()` runs through: a bare `% $params['step']` with no cast, so `step: any` is a `TypeError` there and `step: 0` a `Modulo by zero`.

And while I was in that function I fixed #3585 properly, which your work led me to — the step-grid check now runs on the decimal text rather than on binary floats, so `81.96` with min `-90` and step `0.0000001` validates. Your test file is the base for the extended one and you're credited in the changelog for both.

Thanks for this, and for the follow-up on ASCII case-insensitivity — that was a good catch I'd have missed.

rhukster:
Released in **2.1.6**, so this can close.

Your `strcasecmp` guard is in there at `Validation.php:657`, but it ended up inside a larger rewrite of `typeNumber()` rather than as a standalone patch, which is why the PR now shows as conflicting. Merging it would fight the rewrite for no gain, so I'm closing it as landed rather than as rejected. You're credited in the changelog for it:

> A number, range or select field no longer takes the site down when its step is `any`, zero, or not a number. [...] Thanks @sridharkalaibala

What went in on top of your change: `any` turned out to be one of several ways that line divided by zero, so `step: 0`, `'0'`, `''`, `false` and any non-numeric step are all handled now, along with `typeRange`, which delegates here, and the sibling in `typeArray()` that `typeSelect()` runs through. And your work led me to #3585, which is fixed in the same release — the step-grid check now runs on the decimal text rather than on binary floats, so a latitude of `81.96` on a field stepping by `0.0000001` validates instead of being refused.

Your test file is the base for the extended one in the release. Thanks also for the follow-up on ASCII case-insensitivity in the HTML spec, which I'd have missed.

PR #4310, opened 2026-09-14
Rewrite user/.htaccess to avoid AllowOverride AuthConfig (#4309)

### LINKED ISSUE #4309: Admin going blank upon admin2 plugin update
PHP Version: 8.3.19
Server Software: Apache
Host OS: Linux Debian 6.1.129-1 (2025-03-06) x86_64

On updating the admin2 plugin through the backend, my admin page went completely blank. Checking the inspector/console on firefox, it looked like every single .js file was being loaded as html and blocked (MIME error). At the bottom of the huge list of errors i got `Security Error: Content at [site]/admin/plugins may not load or link to file:///`. Meanwhile the frontend was also broken, with both js and css files failing to load for the same reason.

Error log entry: `[2026-09-14T12:56:55.373293+00:00] grav.WARNING: Grav\Common\File\CompiledFile::logCorruptCache(): Corrupt compiled cache /var/www/html/rsirius/paginas/cache/compiled/files/e22e3a27ac81130855f41891509b4cb7.yaml.php for /var/www/html/rsirius/paginas/system/config/media.yaml (Unclosed '[' on line 124); regenerating from source. [] []`

Upon renaming the `.htaccess` file on the /user folder to `.htaccess.bak` the site returned to normal. Contents in the file were as such:

```# Deny direct web access to code and data files anywhere under user/.
# This is a defense-in-depth backup for the rules in the site root .htaccess.
#
# Why it exists: mod_rewrite rules are not inherited by subdirectories. Any
# .htaccess in a theme or plugin folder that turns on RewriteEngine replaces the
# root ruleset for that folder and everything beneath it, so a package shipping
# its own .htaccess (to block a Makefile, say) can silently expose its .yaml,
# .md, .twig and .php files (getgrav/grav#4236). Access control written with
# Require / FilesMatch is merged into subdirectories rather than replaced, so
# these rules keep applying no matter what a subfolder does with rewriting.
#
# Keep the extension list in sync with the `^(user)/(.*)\.(...)` rule in the
# site root .htaccess.
<FilesMatch "(?i)\.(txt|md|json|yaml|yml|php|php2|php3|php4|php5|phar|phtml|pl|py|cgi|twig|sh|bat)$">
    <IfModule mod_authz_core.c>
        Require all denied
    </IfModule>
    <IfModule !mod_authz_core.c>
        Order allow,deny
        Deny from all
    </IfModule>
</FilesMatch>

# Dot-files: .env, git and CI metadata, editor state. Files *inside* a
# dot-folder cannot be matched here, because <DirectoryMatch> is not allowed in
# .htaccess; those still rely on the root rewrite rules.
<FilesMatch "^\.">
    <IfModule mod_authz_core.c>
        Require all denied
    </IfModule>
    <IfModule !mod_authz_core.c>
        Order allow,deny
        Deny from all
    </IfModule>
</FilesMatch>
```

config info yaml file:

[rsirius-grav-system-info-2026-09-14.yaml](https://github.com/user-attachments/files/32201034/rsirius-grav-system-info-2026-09-14.yaml)

wakqasahmed:
Fixes #4309.

`user/.htaccess` uses `Require`/`FilesMatch` to deny access to config/data files, since that syntax merges into subdirectories instead of getting replaced by a subfolder's own `.htaccess` (unlike `mod_rewrite` rules). That's the right property, but `Require`/`FilesMatch` needs `AllowOverride AuthConfig` (or `Limit`), and a lot of hosts don't grant that. When a host doesn't, Apache doesn't skip the directive and move on — it 500s every request under `user/`, which took out the admin panel and the front end together for the reporter, triggered by nothing more than an admin plugin update.

`RewriteOptions InheritDownBefore` gets the same "don't get replaced by a subfolder's rewrite rules" property using only `mod_rewrite`, which every host running Grav already has to allow since the site root `.htaccess` depends on it too. Rewrote the file to use that instead, matching what @rhukster posted on the issue.

No test included — this is a plain Apache config file with no PHP entry point, so there's nothing in the existing Codeception suite that exercises it.


rhukster:
Thanks — this is the right instinct, and it's the same file I posted on #4309 as a stopgap, so no argument with the direction. develop went further in two steps though, and I'd rather explain than just close it on you.

**2.1.5** added the two exceptions the site root already makes: avatar images under `user/accounts` (both the `avatars/<file>` and the `<username>/<file>` layout Flex folder storage uses) and public asset uploads under `user/data`. Restating the folder denials without those is #4311 — every avatar and every Flex Object image comes back 403. That only stays hidden on a site that has the per-folder `.htaccess` files under `user/accounts` and `user/data`; plenty of upgraded installs don't, and there it's immediate. 2.1.5 also rebuilt those per-folder files on mod_rewrite for the same AllowOverride reason, and added a migration — because `user/` is in the installer's `$ignores` list, changing the shipped file alone never reaches an existing site, which is exactly why rolling back didn't help the reporter.

**2.1.6** added the part the question on the issue was right to push on. `InheritDownBefore` stops the accidental case: a package that ships `RewriteEngine On` with no `RewriteOptions` gets our rules prepended and can't switch them off. But a folder that writes its own `RewriteOptions Inherit` and ends with an unconditional `[L]` flips the order and stops before the inherited rules run. I tested it — under a rewrite-only `user/.htaccess` that folder served its own `secret.yaml`, `.md`, `.twig` and `.env`, and executed its `.php`, all of which the pre-2.1.4 `Require` version denied. So moving off `Require` genuinely narrowed this, and a second set of rules on mod_alias now backs it up: `RedirectMatch` is FileInfo-class too, so it costs nothing on a restricted host, but it sits outside the rewrite pipeline and merges into subdirectories instead of being replaced, and no `[L]` reaches past it.

Net: the AllowOverride fix you're describing shipped in 2.1.5, and the rest lands in 2.1.6. Closing this as already fixed — but the diagnosis in your description is spot on, and it's the reason #4309 got turned around in a day. Appreciated.

PR #4321, opened 2026-09-19
Isolate JS pipeline minification failures, same as CSS (#4313)

### LINKED ISSUE #4313: JS asset pipeline reorders failed assets, swallows Errors, and never reads its cache
### Summary

The JS pipeline has the same three problems the CSS pipeline is being fixed for in #4307, and one of them is worse on the JS side.

`Pipeline::gatherAndMinifyJs()` minifies each file individually, excludes any that throw from the combined buffer, and returns them in `failed`. `Assets::render()` then appends every failed asset *after* the whole bundle:

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Assets.php#L481-L488

For CSS that changes which rule wins. For JS it changes **execution order**. If the file the minifier chokes on is a dependency — jQuery, a polyfill, anything that defines a global — it now loads after everything that expects it, and the page dies with a `ReferenceError` it did not have before. The failure is silent at build time and only shows up in the browser.

### Three separate problems

**1. Failed assets execute out of order.** As above. The fix is the same one #4307 lands for CSS: when any asset in the group fails to minify, bundle the whole group unminified in its original order rather than emitting a partial bundle plus a reordered tail.

**2. `catch (\Exception)` lets `\Error` through.**

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Assets/Pipeline.php#L415

A `TypeError` out of JShrink is an `\Error`, not an `\Exception`, so it escapes the catch and 500s the page — which is the exact crash the per-asset isolation was added to prevent. #4307 correctly uses `catch (\Throwable)` for CSS; JS should match.

**3. A group with one bad file re-minifies on every request.**

https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Assets/Pipeline.php#L243-L248

`renderJs()` only reads the cache when there are no failures (`if (empty($failedAssets) && file_exists($filepath))`), but writes the partial bundle every time. So a site with a single unminifiable JS file runs JShrink over its entire JS group on every page view, forever, and never reads the file it just wrote.

### Suggested fix

Mirror whatever lands in #4307 for CSS, plus widen the catch:

- On any failure in the group, fall back to `gatherLinks()` for the whole group so order is preserved and the bundle still caches.
- `catch (\Exception $e)` → `catch (\Throwable $e)`.

Found while reviewing #4307. Filing separately so the CSS fix isn't held up.

wakqasahmed:
Fixes #4313.

The JS pipeline had the same three problems the CSS pipeline was just fixed for in #4307, and one of them is worse here.

`Assets::render()` appends every failed asset after the entire successful bundle, since `renderJs()` returns `{output, failed}` and the failed ones get rendered on their own afterward. For CSS that can change which rule wins the cascade; for JS it changes execution order — if the file that fails to minify is a dependency like jQuery, it now loads after code that expects it, and the page throws a `ReferenceError` it didn't have before. `gatherAndMinifyJs()` also caught `\Exception`, so a `\TypeError` (or any other `\Error`) out of JShrink escapes uncaught and 500s the page — the exact crash per-asset isolation exists to prevent, and CSS already moved to `\Throwable` for the same reason. And since `renderJs()` only reads its cache when there are no failed assets but writes the partial bundle every time regardless, a group with one permanently-failing file re-minifies the whole JS group from scratch on every single request, forever, without ever reading the file it just wrote.

Mirrors #4307 exactly: when any asset in the group fails to minify, fall back to the whole group concatenated unminified in its original order, rather than a partial bundle plus a reordered tail — the same output `js_minify: false` produces for these files — so the bundle still caches as one file, in order, this time actually getting read back on the next request. Since a minify failure no longer produces a separate list of failed assets to hand back, `renderJs()` goes back to always returning `string|false` like `renderCss()` already does, and the now-unreachable array-result branch in `Assets::render()` comes out with it.

Added a broken/valid JS fixture pair (an unterminated string, which JShrink throws a `RuntimeException` on) and a test asserting the fallback bundle contains both, in original order, as a single cached `<script>` tag — plus one confirming the successful-minify path is unaffected. Ran the full `AssetsTest` suite (30/30) and the full unit suite; the only failures are pre-existing, unrelated image/GD-fallback and Markdown rendering tests already present on `develop`.


wakqasahmed (commented):
Reviewed cold, cross-checked against #4313, #4307 (current head `3aadfa4`, not the stale PR description — its body still describes an earlier per-asset-render design that #4307 itself moved away from after review, converging on the same whole-group-fallback pattern this PR uses), and a full-codebase search for other consumers.

**Mirroring claim (#4307): confirmed accurate.** `renderJs()`/`gatherAndMinifyJs()` structurally match `renderCss()`/`gatherAndMinifyCss()` at #4307's current head line-for-line in approach: cache-check-before-compute, lazy `gatherAndMinifyJs()` only on cache miss, fallback to `gatherLinks()` on any failure, comment style reused almost verbatim. The one divergence — the extra `$type`/`JS_MODULE_ASSET` parameter and `renderJs_Module()` delegation — is justified; CSS has no module-asset concept.

**`renderJs()` return-type narrowing: safe.** Searched the entire repo (`system/`, `tests/`, `user/plugins` — empty, only `.gitkeep`) for `renderJs`, `renderJs_Module`, `is_array($result)`, and `['failed']` reads. The only caller is `Assets::render()`'s dynamic `$render_pipeline` dispatch, already updated in this diff. `renderJs_Module()` only delegates to `renderJs()` and has no other callers either. No dead code or breakage left behind.

**Dynamic dispatch (`Assets::render()`):** confirmed the simplified `if ($result === false) { continue; } $pipeline_output .= $result;` is correct for all three pipeline types post-fix — `renderCss()`, `renderJs()`, and `renderJs_Module()` (which just calls `renderJs()`) now all return `string|false` uniformly. No remaining non-string/non-false case.

**Test gap (real, but shared with #4307's own precedent):** verified directly against JShrink's source (`tedious/JShrink` `Minifier.php`) — unclosed strings throw `\RuntimeException`, which `extends \Exception`. The new `broken-unclosed-string.js` fixture therefore does *not* exercise problem #2 (`\Exception` → `\Throwable`); the old `catch (\Exception)` would have caught this too. I checked #4307's own current test fixture the same way: its `broken-modern-syntax.css` case goes through cssmin's warning → Whoops `ErrorException`, which also `extends \Exception`. So the claim of parity with #4307 holds — neither PR's tests actually prove the `\Throwable` widening prevents a real `\Error`/`\TypeError` from escaping. Left as an inline note below rather than a blocker since it mirrors accepted precedent, but a fixture that trips an actual `\Error` (e.g. a minifier call with a bad type) would close this gap for good on both sides.

Everything else checked out: no other correctness issues found, style matches the codebase, CI is green (8.3/8.4/8.5 unit-tests all pass). This is a solid, well-scoped fix — approving.

system/src/Grav/Common/Assets/Pipeline.php
```diff
@@ -416,7 +420,7 @@ private function gatherAndMinifyJs(array $assets, int $type): array
                 $file = JSMinifier::minify($file);
                 $file = rtrim($file) . PHP_EOL;
                 $buffer .= $file;
-            } catch (\Exception $e) {
+            } catch (\Throwable $e) {
                 // Track failed asset for individual rendering
```
wakqasahmed: Minor/pre-existing: this comment ("Track failed asset for individual rendering") is now stale given this PR's whole point — failed assets are no longer rendered individually, they trigger the whole-group unminified fallback instead. Not touched by this diff (only the catch type changed on this block), but worth updating while you're here since it's directly adjacent to what changed.

tests/unit/Grav/Common/AssetsTest.php
```diff
@@ -613,6 +613,48 @@ public function testClockworkScriptBypassesPipeline(): void
         self::assertMatchesRegularExpression('#<script src="/assets/[a-f0-9]+\.js"></script>#', $js);
     }
 
+    public function testJsMinificationFailureFallsBackForWholeGroup(): void
```
wakqasahmed, line 616: Verified against JShrink's source: an unclosed string throws `\RuntimeException`, which `extends \Exception`. So this fixture (and `testJsMinificationSucceedsWhenNoAssetFails` below it) proves the whole-group-fallback/ordering fix (problem #1) and the cache-write-on-fallback fix (problem #3), but does not exercise the `catch (\Exception)` → `catch (\Throwable)` widening (problem #2) — the old catch clause would have caught a `RuntimeException` too. A fixture that trips a real `\Error` (e.g. `\TypeError`) would be needed to prove #2. Noting this is not unique to this PR — I checked, and #4307's own CSS-side fixture has the identical gap (its failure also arrives as an `ErrorException`, which also extends `\Exception`), so this is parity with existing precedent rather than a regression. Non-blocking, but worth a follow-up fixture on both sides if you want the widening actually proven.

rhukster:
Thanks, this is a nice mirror of #4307. I ran it on its own and merged it together with #4307: both land cleanly and the Assets tests pass. Nothing else in core or the first-party plugins used the old array return from `renderJs()`, so switching back to a plain string is safe. Checking the cache before minifying is a nice extra win too, since the old code re-minified every JS file on every request.

PR #4324, opened 2026-09-20
Include the requested max width in image derivatives (#2429)

### LINKED ISSUE #2429: [derivatives] mathematic mistake
derivatives does not include the right boundary,
```
{{ image.derivatives(320,960,320).sizes('(max-width: 20.063em) 20rem, (max-width: 40.063em) 40rem, 60rem').html() }}
```
will generate:
```
<img alt="" src="/images/0/5/9/4/c/0594ce39a0931d4441dfe06177b634266e4079f9-01.jpeg" 
srcset="/images/9/a/2/6/f/9a26f38ba8fda920fb2e8a2a711468a99897a489-01320w.jpeg 320w,
 /images/e/0/c/8/4/e0c84893b1949338da6a8e942c95c98e0e614c63-01640w.jpeg 640w,
 /images/0/5/9/4/c/0594ce39a0931d4441dfe06177b634266e4079f9-01.jpeg 1600w" sizes="(max-width: 20.063em) 20rem, (max-width:
                 40.063em) 40rem, 60rem" class="lslide" style="width: 960px; margin-right: 0px;">
```
not include 960 as expected, so should set derivatives(320,961,320) to fit the usage
maybe it's not a problem, but isn't very intuitive.

I think need be change a "<" to "<=" somewhere  😆

Grav v1.5.10

wakqasahmed:
Fixes #2429.

derivatives()'s min/max/step branch builds widths with a strict `<` upper bound, so an explicitly requested `$max_width` never actually lands in the generated srcset, even when it's smaller than the real image. `derivatives(320, 960, 320)` on an image wider than 960px produces 320w and 640w and silently drops 960w, which is exactly what the reporter ran into.

Changed the loop to `<=` so the requested max width is included, with a guard that skips any width at or past the base image's actual width. That guard is needed because `ImageMedium::srcset()` always appends the base image at its own full width as the last entry, so without it a max_width that lands on the base width would show up twice. The single-argument branch of derivatives() a few lines up already guards the same way (`< $base->get('width')` before adding a width), so this just keeps both branches consistent.

Added a test alongside the existing `ImageMediumAlternativesFormatTest` (same fixture, same setup style): one case reproducing the reported bug (200/1000/200 on the 1024px fixture, which now includes 1000w), and another checking the boundary itself doesn't produce a duplicate entry when max_width lands on or past the base width.

Ran the Medium test suite in php:8.3 (matching composer.json) — 40/41 pass, one pre-existing failure in `testQualityReachesEveryAlternative` unrelated to this change and present identically on unmodified develop. Also ran phpstan level 2 against the changed file with no errors.


rhukster:
Thanks @wakqasahmed, this is a clean fix and the test is exactly right. I checked the boundary cases (max past the image width, max landing on it, steps that don't divide evenly, and the single-argument form) and they all behave, and it merges cleanly onto current develop where the Medium tests all pass. Merged, thank you!

PR #4326, opened 2026-09-22
Exempt .well-known/ from the built-in-server dot-path block (#4016)

### LINKED ISSUE #4016: Ignore hidden does nothing
Setting ignoring hidden does nothing.

## Steps to reproduce

1. Set ignore hidden = False. 

Image

2. Create a file like `.well-known/test`
3. Access `example.org/.well-known/test` in the browser

## Expected result

Raw content of `.well-known/test` is shown

## Actual result

404 error page is rendered

## Possible reason

In `system/router.php`, line 21, the access to files and folders is blocked, regardless of the setting

```
// Block all direct access to files and folders beginning with a dot
strpos($path, '/.') !== false
// Block all direct access for these folders
|| preg_match('`^/(\.git|cache|bin|logs|backup|webserver-configs|tests)/`ui', $path)
```

Uncommenting the line solves the issue

## Reasoning

Why I need this: Grav should be not only be serving a blog but also some static files that are required to be served on the root domain by the matrix protocol: https://spec.matrix.org/latest/client-server-api/.

wakqasahmed:
Fixes #4016.

The reporter's diagnosis pointed at `system.pages.ignore_hidden`, but that setting only controls whether hidden page folders are skipped when Grav scans `user/pages/` — it has nothing to do with serving static files, so it wasn't actually the cause here.

The real issue is in `system/router.php`, which is only used when running Grav through PHP's built-in dev server (`bin/grav server` / `php -S`). It blocks every path containing `/.` unconditionally, which also catches `.well-known/...`. Both of the shipped production webserver configs already carve out an exception for this:

- `.htaccess`: `RewriteRule (^|/)\.(?!well-known) - [F]`
- `nginx.conf`: `location ~ (^|/)\.(?!well-known(?:/|$)) { return 403; }`

So this looks like drift rather than a deliberate restriction — the dev-server router just never got the same exception. This PR mirrors the existing negative-lookahead intent in `system/router.php` so `.well-known/` paths are served directly, while every other block (`.git`, `cache`/`bin`/`logs`/`backup`, restricted file types, other hidden files, etc.) stays exactly as it was — including hidden files nested inside `.well-known` itself (e.g. `.well-known/.git/config` is still blocked).

Added `tests/unit/Grav/RouterTest.php`, which boots the actual `system/router.php` under PHP's built-in server and makes real HTTP requests against it, covering both the newly-allowed `.well-known` paths and everything that should still be blocked (including a `.well-known-private` decoy folder and nested hidden paths inside `.well-known`).

**Testing**
- `vendor/bin/codecept run unit Grav/RouterTest.php` — 2 tests, 32 assertions, all passing
- Confirmed the new test fails against `develop` before this change (`.well-known/anything` served the Grav index instead of the static file) and passes after


wakqasahmed:
Note on the CI failure: \`AssetsTest::testCssMinificationKeepsSpaceSeparatedColors\` fails on unmodified \`develop\` too (verified in an isolated PHP 8.3 container with only this PR's two files reverted — the rest of the suite is untouched). Looks like a pre-existing CSS-minifier color-rounding mismatch (\`#4d80b3\` expected vs \`#4d7fb3\` actual), unrelated to this change. \`RouterTest\` (the new test in this PR) passes on all three PHP versions.

Issue #4317, opened 2026-09-16
Let format() and quality() reach srcset alternatives


onetrev:
Hi Andy,

Grav 2.1.6. `images.defaults` accepts `format: webp`, but only the base medium is converted. `ImageMediaTrait::derivatives()` builds each alternative fresh from the file (and `reset()` puts its format back to `guess`), and `format()` and `quality()` are declared methods on the trait, so `ImageMedium::__call()` never fans them out to the alternatives the way it does for `resize` and the other magic actions. A Markdown image ends up with a WebP `src` and largest candidate, and JPEG/PNG for every smaller one in the srcset.

Suggestion: have `format()` and `quality()` loop over `$this->alternatives` like `__call()` does. Then `images.defaults: { format: webp }` serves every body image as WebP with no theme code.

Happy to help test, just say the word!

Cheers,
Trev

rhukster:
if you want to create a PR, i will review it.. Cheers!

PR #4329, opened 2026-09-23
Propagate quality() and format() to derivatives() alternatives

### LINKED ISSUE #3146: Image option 'quality' and 'format' do not work with derivatives() and responsive images
```
{{ image.derivatives([480, 600, 1000]).quality(1)|html|raw }}
```

This does not work with `derivatives`. It only changes the quality of the main image to 1, but not the others in srcset. Originally reported here: https://github.com/getgrav/grav/pull/3144#issuecomment-758840822

wakqasahmed:
Closes #3146.

`derivatives()` generates each srcset alternative as its own fresh `Medium` instance up front, and only after that do `quality()`/`format()` get called in the fluent chain (`.derivatives([...]).quality(1)`, per the docs). Since those setters only touched `$this`, everything after `derivatives()` -- the alternates already created -- kept its default quality/format, and only the base image changed.

I made `quality()` and `format()` also apply to whatever's already in `$this->alternatives` when they're called, since that's the only way the documented chained usage can behave the way it reads. Added a regression test covering both setters against the actual generated alternatives.


rhukster:
Thanks for digging into this one, and for the regression test. It turns out this was already fixed in 2.1.7 (e0a56e0cb, while working on #4317): `quality()` and `format()` now reach every alternative, whether `derivatives()` ran before or after them, and retina alternatives too. `ImageMediumAlternativesFormatTest` covers the same cases as your test, so I'm closing this as already fixed, along with #3146. Appreciate the contribution!

